PR #4, opened 2020-04-09
Don't run CI pipeline on pull requests that change documentation only

just95:
Since the master branch is now protected, updates of the README and documentation have to be performed via pull requests, too. In order to avoid wasting resources, I suggest that the CI pipeline no longer runs on pull requests that modify only the README or other Markdown files only. I've modified the GitHub actions workflow configuration file such that the same rules that applied to pushes before now apply to pull requests as well.

### DIFF
--- a/.github/workflows/ci-pipeline.yml
+++ b/.github/workflows/ci-pipeline.yml
@@ -1,10 +1,9 @@
 name: CI Pipeline
 
-# The CI pipeline runs under the following criteria.
-#   - On every push that modifies Cabal configuration, source or example files.
-#     If just the README or documentation changes, the pipeline does not have
-#     to run. It also runs when the Workflow configuration changed itself.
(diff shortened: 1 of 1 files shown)


just95:
Another interesting observation: Even if the CI pipeline ran for a commit on a push event already, it apparently runs on a pull request again (see above :sweat_smile:). @ichistmeinname do you think that I should disable the CI pipeline on push events entirely?

ichistmeinname:
Since all checks (i.e. the pipeline) were running for this pull request already, I guess that we can then disable to rerun them when the request merged. @nbun Do you agree?

ichistmeinname (changes requested):
Remove GitHub action for push events.

nbun (approved):

ichistmeinname (approved):

PR #10, opened 2020-04-17
Fix #9

just95:
The problem described in #9 was caused by the changes in b20216bb44a5990f0eff14d0b3fd135473ac2986. Before that commit, the IR pretty-printer supported some syntactic sugar form Haskell. Since the same syntactic sugar is intentionally left out from the IR parser but the pretty-printer should only print valid strings of the intermediate language, the syntactic sugar was removed. The syntax used by IR for Haskell's build-in types is not compatible with the Haskell parser. Since the module interface file decoder was not updated to use the new IR parser instead of the Haskell parser and the incompatible Prelude definitions are always part of module interface files, decoding module interface files failed.

Issue #8, opened 2020-04-15
Document Windows-specific information for installation and usage guide


ichistmeinname:
Several Windows users mentioned that the installation process as currently documented needs to be adapted to work on Windows.
I suggest to rerun the whole installation Windows step-by-step in order to work out and document the Windows-specific changes necessary to have a running system. Furthermore, if the usage differs from a unix-based system as well, the corresponding documentation should have a Windows-specific section as well.

ichistmeinname:
We don't care enough for now. Windows users should just use `cabal v2-build` and `cabal v2-exec` instead of `cabal v2-install`.

PR #12, opened 2020-04-21
Fix #11

### LINKED ISSUE #11: Git hash is missing from `--version` output if the compiler is installed
### Description

When the compiler is installed as described in the README, the `--version` command line option of `freec` does not include the compile-time output of `git describe --always --dirty`.

### Steps to Reproduce

1. Install the compiler as described in the README using the following command.

    ```bash
   cabal new-install freec
    ```

2. Run the following command.

    ```bash
   freec --version
    ```

#### Expected behavior

The output should be the same as the output of the following command.

```bash
./tool/run.sh --version
```

For example,

```
The Free Compiler, version 0.1.0.0 (v0.1.0.0-N-gXXXXXXX, ghc 8.6, linux, x86_64)
```

where `N` is the number of commits since `v0.1.0.0` has been tagged and `XXXXXXX` is the current short Git commit hash.
This information is intended for bug reports to help maintainers know which exact version of the compiler is affected.

#### Actual behavior

Because Cabal builds local packages in a temporary directory when using the `new-install` command, [`githash`](https://hackage.haskell.org/package/githash) does not have access to the `.git` directory. Thus, the output of `git describe --always --dirty` is not included.

```
The Free Compiler, version 0.1.0.0 (ghc 8.6, linux, x86_64)
```

### Versions

 - `freec --version`: `The Free Compiler, version 0.1.0.0 (ghc 8.6, linux, x86_64)`  
    but should be `The Free Compiler, version 0.1.0.0 (v0.1.0.0-525-gf48fc57, ghc 8.6, linux, x86_64)`
 - `ghc --version`: 8.6.5
 - `cabal --version`: 2.4.1.0
 - OS: Ubuntu 19.10

### Known workarounds

I have not found a way to configure Cabal to run the installation in the package's root directory instead of a temporary directory. I think the only way to work around this issue is by writing the output of `git describe --always --dirty` to a `VERSION` file before `cabal new-install` is invoked and including the contents of this file into the `--version` command at compile time. I'll post a PR for this approach soon. Suggestions for alternative approaches are welcome.

just95:
## Description

The issue described in #11 occurs because Cabal builds the executable installed via `new-install` in a temporary directory. Inside the temporary directory only files from the sdist tarball are accessible. Thus, [`githash`](https://hackage.haskell.org/package/githash) cannot access the `.git` directory to find the exact version at compile-time. To fix this issue, we have to save the version before Cabal creates the sdist archive. This pull request accomplishes this goal as follows.

 1. There is now a `./tool/install.sh` script which basically just runs the `cabal new-install` command. However, before the command is executed, the output of `git describe --always --dirty` is written to a `VERSION` file and the `VERSION` file is deleted once the compiler has been installed (or the installation has been canceled by the user).

 2. The `VERSION` file is listed as a `data-file` in the Cabal configuration of the compiler. This makes it part of the sdist archive and thus available in the temporary build directory.

 3. The contents of the `VERSION` file are included into the output of the `--version` command line option using Template Haskell. If the file is not present (e.g. when the compiler is executed using `./tool/run.sh` ) we can fall back to `githash`.

## Drawbacks

This solution is far from perfect, though. 

 - Because `VERSION` is a `data-file` but (intentionally) exist only for a brief period during the execution of the installation script, running `cabal new-install freec` manually does not work anymore. The upside is that it is not possible to build the compiler without exact version information (which fixes #11). The downside is that installation is highly dependent on the new Bash script. This could lead to problems on Windows for example (see also #8).

 - Users have described the current installation process as "frustrating" already. There is the risk that these changes adds to the frustration.

 - Special care has been taken to ensure that the `VERSION` file is cleaned up properly. However, if this process fails for some reason, the `cabal new-install freec` command can be executed again. Thus it is possible to build an executable which only pretends to provide exact version information (which could be very confusing).

## Alternatives

The following are some alternatives that I have considered.

 - We could provide a default `VERSION` file (e.g., whose contents are `unknown`) which is temporarily overwritten by the installation script. This file can then be used as a fallback on systems where the installation script does not work. However, there is the risk that users always use `cabal new-install freec` rather than the installation script.

 - We could release new tagged versions more frequently. The demand for the exact version information arises because the information that the compiler is installed in version `0.1.0.0` is not helpful at all since it always has been in version `0.1.0.0`. I suggest releasing `0.2.0.0` once #10 (plus potentially this PR) has been merged. Additionally, we should establish a proper release cycle.

 - In an ideal world, we could tell Cabal to apply Template Haskell to the `FreeC.Application.Option.Version` module before creating the source distribution archive. However, I don't think that such an option exists.

@ichistmeinname, @nbun what are your thoughts on this?


ichistmeinname:
Can you check if the following library might help?

https://github.com/valderman/data-embed

I found it via [a discussion on reddit](https://www.reddit.com/r/haskell/comments/erhwe8/what_is_the_best_way_to_inject_information_into_a/).

If not I'd rather use a better versioning scheme for the project (as you suggested as an alternative) and demand bug reports include the git hash the compiled project was built with.

nbun:
I agree that it should not be a problem for users to include the commit hash in bug reports since installation is supported only by means of cloning the repository for now. When we get to the point where other ways of installing the compiler are supported, a version number should be sufficient I hope.

just95:
Thanks for your feedback : )

@ichistmeinname, I had a look at `Data.Embed`, but I don't think that it solves our problem since the `embedtool` would have to be executed after the compiler has been built. Thus, the current approach using Template Haskell is more suitable. I guess the key takeaway from the reddit post is that there is currently no good way to solve the problem (other than "hacks" like the one I've implemented before).

>  it should not be a problem for users to include the commit hash in bug reports.

@nbun I agree and in case of the other two repositories the bug report template explicitly asks for the Git commit hash only. The reason I would still prefer `freec --version` to contain the commit hash is that it is easy for the installation and Git repository to get out of sync. Consider the following scenario for example.

 1. A user clones the Git repository and installs the compiler.
 2. They encounter a bug.
 3. The user tries to reproduce the bug in the latest version of the compiler.
     - They pull the repository.
     - They run `cabal new-install freec`.
     - They do not notice the easy to miss warning of Cabal:
        ```
        cabal: Path '~/.cabal/bin/freec' already exists. 
        Use --overwrite-policy=always to overwrite.
        ```
     - The compiler has not actually been updated, thus the bug seems to persist in the latest version even if it was fixed in the meantime.
 4. If the user reports the bug, the output of `freec --version` provides no information that they have been using an outdated version and the inclusion of the Git hash (which will be HEAD of master due to the pull) is now just confusing everybody who is involved.

---

I've come up with the following compromise solution.

 - We use the same installation procedure as before. The `./tool/install.sh` script has been removed.
 - The output of `freec --version` will not contain the compile-time output of `git describe --always --dirty`. The `VERSION` file has been removed.
 - However, the `./tool/run.sh --version` command still provides exact version information using `git describe --always --dirty` via [`githash`](https://hackage.haskell.org/package/githash). This is possible since `new-build` uses in-tree builds.
 - There are going to be version suffixes `-stable` and `-dev`. The `-stable` suffix is added only for the single tagged commit that corresponds to the release of a version. All other commits have a version with `-dev` suffix. Since version tags like these are deprecated in Cabal, I'm using the CPP language extension to include this information into the `--version` command.

A new version of the compiler would now be released as follows.

 1. The `version` field of the Cabal file is incremented. This step is only necessary for the first release that uses the new release cycle (e.g., `0.1.0.0 → 0.2.0.0`).
 2. The `FREEC_VERSION_SUFFIX` pre-processor constant is set to `-stable` in the Cabal file.
 3. The "unreleased" section of the `CHANGELOG.md` is changed to a "X.X.X.X / YYYY-MM-DD" section.
 4. The changes above are committed as "Release version X.X.X.X".
 5. The commit above is tagged as `vX.X.X.X`.
 6. Immediately afterwards (even before pushing the release commit and tag) the `version` is incremented again (e.g., `0.2.0.0 → 0.3.0.0`).
 7. The `FREEC_VERSION_SUFFIX` pre-processor constant is set back to `-dev` in the Cabal file.
 8. A new "unreleased" section is added at the top of the `CHANGELOG.md`.
 9. The changes above are committed as "Start version Y.Y.Y.Y release cycle".
10. Now both commits and the tag can be pushed.

This approach allows us to distinguish the following cases based on the `--version` output when a bug report is opened. The main distinction we can make is whether they used a tagged version (and do not have to provide additional information) or they are using a development version (and may have to provide more information).

##### Pre `0.1.0.0` versions

###### `The Free Compiler, version 0.1.0.0 (ghc 8.6, linux, x86_64)`

The user must have installed the compiler between ddd7914 and the merge of this PR. The user should include the Git hash in their bug report and whether or not they have made local modifications. Ideally, they run `./tool/run.sh --version` instead.

###### `The Free Compiler, version 0.1.0.0 (v0.1.0.0-N-gXXXXXXX, ghc 8.6, linux, x86_64)`

The user runs the compiler without installation (e.g., using `./tool/run.sh`)  in a clean working directory from commit `XXXXXXX` in a version that pre-dates the new release cycle. There is no need for them to include the Git hash in their bug report. If they include the hash nonetheless, it should match `XXXXXXX`.

###### `The Free Compiler, version 0.1.0.0 (v0.1.0.0-N-gXXXXXXX-dirty, ghc 8.6, linux, x86_64)`

The user runs the compiler without installation (e.g., using `./tool/run.sh`)  from commit `XXXXXXX` in a version that pre-dates the new release cycle but there are uncommitted changes. If the bug cannot be reproduced, we should ask for which changes they have made.

##### Stable post `0.1.0.0` versions

###### `The Free Compiler, version Y.Y.Y.Y-stable (ghc 8.6, linux, x86_64)`

The user has either installed the compiler from a tag `vY.Y.Y.Y` or is running it using `./tool/run.sh` but Git is not installed (e.g. when they downloaded an archive for `vY.Y.Y.Y` from GitHub). Unless they made local modifications to the source code, the bug should be reproducible in version `Y.Y.Y.Y`.

###### `The Free Compiler, version Y.Y.Y.Y-stable (vY.Y.Y.Y, ghc 8.6, linux, x86_64)`

The user is running the compiler without installation (e.g., using `./tool/run.sh`) from tag `vY.Y.Y.Y` in a clean working directory. The bug should definitely be reproducible in version `Y.Y.Y.Y`.

###### `The Free Compiler, version Y.Y.Y.Y-stable (vY.Y.Y.Y-dirty, ghc 8.6, linux, x86_64)`

The user is running the compiler without installation (e.g., using `./tool/run.sh`) from tag `vY.Y.Y.Y` but they have uncommitted changes. If the bug cannot be reproduced, we should ask for which changes they have made.

Due to the rules for the release cycle, a `git describe` output of the form `vY.Y.Y.Y-N-gXXXXXXX` is impossible.

##### Post `0.1.0.0` development versions

###### `The Free Compiler, version Y.Y.Y.Y-dev (ghc 8.6, linux, x86_64)`

The user has installed the compiler from some unknown commit after the release of `vX.X.X.X` but before `vY.Y.Y.Y` has been released. The user should include the exact Git hash in their bug report and whether or not they have made local modifications. Ideally, they run `./tool/run.sh --version` instead.

###### `The Free Compiler, version Y.Y.Y.Y-dev (vX.X.X.X-N-gXXXXXXX, ghc 8.6, linux, x86_64)`

The user is running the compiler without installation (e.g., using `./tool/run.sh`)  in a clean working directory from commit `XXXXXXX`. There is no need for them to include the Git hash in their bug report. If they include the hash nonetheless, it should match `XXXXXXX`.

###### `The Free Compiler, version Y.Y.Y.Y-dev (vX.X.X.X-N-gXXXXXXX-dirty, ghc 8.6, linux, x86_64)`

The user is running the compiler without installation (e.g., using `./tool/run.sh`)  from commit `XXXXXXX` but there are uncommitted changes. If the bug cannot be reproduced, we should ask for which changes they have made.

ichistmeinname (changes requested):
As @just95 has a clear plan how to manage versions and it clearly seems important to him to incorporate this feature, I'll agree with the merge.
Please document the procedure you illustrated in the pull request (as a paragraph in CONTRIBUTING.md?).

nbun:
I am not sure whether solving an organizational problem with more organization will work out as expected, but as @just95 said, it is a compromise.

Issue #2, opened 2020-04-09
Compiler installation fails with cabal 2.4.0.0


MajaRet:
The installation of the compiler via `cabal new-install freec` fails with the following message:
`Wrote tarball sdist to 
~/free-compiler/dist-newstyle/sdist/free-compiler-0.1.0.0.tar.gz`
`cabal: The impossible happened: a local package isn't local`
when I use cabal version 2.4.0.0.
Upgrading to cabal 3.2.0.0 fixes the problem.

The GHC version is 8.6.5.

ichistmeinname:
Might be related to https://github.com/haskell/cabal/issues/5643.

just95:
I've just installed Cabal 2.4.0.0 locally and can confirm, that the `new-install` command is in fact broken in that version. Until now I have been using version 2.4.1.0 on my machine and in the CI pipeline. I've updated the documentation to clarify that the compiler has been tested with version 2.4.1.0 only. Of course you can continue using 3.2.0.0 if that version works. Unfortunately, I'll not be able to test 3.2.0.0 automatically as the latest version currently available in GitHub actions is 3.0.0.0.

ichistmeinname:
@MajaRet Can you modify the cabal-version specified in free-compiler.cabal to 2.4.1.0? Please check Justin's contributing guide ([CONTRIBUTING.md](https://github.com/FreeProving/free-compiler/blob/master/CONTRIBUTING.md#how-can-i-contribute)) beforehand.

ichistmeinname:
As far as we checked, we cannot annotate the exact cabal-version.

PR #14, opened 2020-04-28
Adjust to #2

MajaRet:
Since the `new-install` command does not work correctly in Cabal version 2.4.0.0, the compiler cannot be installed when that version is used. 

Therefore, the cabal version in free-compiler.cabal was modified to 2.4.1 to avoid this problem.


MajaRet:
The CI pipeline failed with the message 
Run cabal new-update
Config file path source is default config file.
Config file /home/runner/.cabal/config not found.
Writing default configuration to /home/runner/.cabal/config
CabalFileParseError "./free-compiler.cabal" [PError (Position 0 0) "Unsupported cabal-version. See https://github.com/haskell/cabal/issues/4899."] (Just (mkVersion [2,4,1])) []
##[error]Process completed with exit code 1.

The exact line in free-proving.cabal right now is "cabal-version: 2.4.1". 
Using "2.4.1.0" causes an error when running the unit test script locally instead:
"cabal-version should be at the beginning of the file starting with spec version".
What am I doing wrong here...?

just95:
I did some research and it seems that the [`cabal-version` field](https://www.haskell.org/cabal/users-guide/developing-packages.html#pkg-field-cabal-version) does not actually specify the minimum version of `cabal` that has to be installed but the version of the [Cabal Package Description Format](https://www.haskell.org/cabal/users-guide/file-format-changelog.html#spec-history). However, the file format did not change from 2.4.0.0 to 2.4.1.0, thus 2.4 is still the right value for `cabal-version`. There is probably no way to specify in the `.cabal` file that the compiler is not compatible with Cabal 2.4.0.0.

ichistmeinname:
Sorry, my mistake. Going to abort this request.

PR #32, opened 2020-05-25
Add some documentation to the coq backend

JonasHoefer:
While reading the backend code i added and edited some comments (but mostly added notes for myself). If some of them are useful we could add the to the main code base.


just95 (approved):

src/lib/FreeC/Backend/Coq/Converter/Expr.hs
```diff
@@ -40,6 +40,15 @@ convertExpr expr = convertExpr' expr [] []
 convertExpr' :: IR.Expr -> [IR.Type] -> [IR.Expr] -> Converter Coq.Term
 
 -- Constructors.
+-- 
+-- Partial applied constructors are not evaluated in Haskell and therefor
```
ichistmeinname, line 44: Partially

ichistmeinname:
```suggestion
-- Partially applied constructors are not evaluated in Haskell and therefor
```

src/lib/FreeC/Backend/Coq/Converter/Expr.hs
```diff
@@ -40,6 +40,15 @@ convertExpr expr = convertExpr' expr [] []
...
+-- cannot be @⊥@. The translated type of a constructor @C : τ₀ -> … -> τₙ@ is
+-- @c : τ₀' -> … -> τₙ*@ instead of @m(τ₀' -> m(τ₁' -> m(… -> τₙ')))@.
```
ichistmeinname, line 46: Unfortunately, I see here some undefined (unicode?) characters : /

src/lib/FreeC/Backend/Coq/Converter/Expr.hs
```diff
@@ -40,6 +40,15 @@ convertExpr expr = convertExpr' expr [] []
...
+--
+-- Note that the return type is translated using * not ', because a constructor
+-- in Coq cannot return a wrapped value. A smart constructor @C@ is generated,
+-- which wrapps the value of @c@. Is therefor sufficient to just convert and
```
ichistmeinname:
```suggestion
-- which wrapps the value of @c@. It is therefor sufficient to just convert and
```

ichistmeinname (changes requested):

src/lib/FreeC/Backend/Coq/Converter/Expr.hs
```diff
@@ -134,14 +143,28 @@ convertExpr' (IR.Var srcSpan name _) typeArgs args = do
...
+-- > convertExpr'  / \   tArgs args = convertExpr' e (τ : tArgs) args
+-- >              e   τ
 convertExpr' (IR.TypeAppExpr _ e t _) typeArgs args =
   convertExpr' e (t : typeArgs) args
 
 -- @if@-expressions.
+--
+-- > ⎡Γ ⊢ p:Bool  Γ ⊢ t:τ  Γ ⊢ f:τ⎤'     Γ' ⊢ p':Bool'  Γ' ⊢ t':τ'  Γ' ⊢ f':τ'
+-- > ⎢――――――――――――――――――――――――――――⎥ = ―――――――――――――――――――――――――――――――――――――――――――
+-- > ⎣ Γ ⊢ if p then t else f : τ ⎦   Γ' ⊢ p' >>= λx:𝔹'.if x then t' else f' : τ'
```
ichistmeinname, line 164: Very good idea!

src/lib/FreeC/Backend/Coq/Converter/Expr.hs
just95:
```suggestion
-- which wrapps the value of @c@. It is therefore sufficient to just convert and
```

src/lib/FreeC/Backend/Coq/Converter/Expr.hs
JonasHoefer: The [styleguide](https://github.com/FreeProving/guidelines/blob/master/CONTRIBUTING.md#haddock-styleguide) suggested to use Unicode symbols over inline LaTeX. I thought the comment would be easier to read with constructor parameters that use actual subscript indices (similar to the algorithm description in [DecreasingArguments.hs](https://github.com/FreeProving/free-compiler/blob/master/src/lib/FreeC/Backend/Coq/Analysis/DecreasingArguments.hs)). The comment contains a Greek `tau`, subscript `1`, `2` and `n` and the `bottom` symbol.

ichistmeinname: It's the _subscript n_ that my browser refuses to render : / Weird, but it's finde then.

ichistmeinname (approved):

PR #37, opened 2020-05-28
 Add wrapper module for Agda AST

### LINKED ISSUE #27: Add wrapper module for Agda AST
### Description

Add a module similar to [`FreeC.Backend.Coq.Syntax`](https://github.com/FreeProving/free-compiler/blob/master/src/lib/FreeC/Backend/Coq/Syntax.hs) that reexports the Agda AST. The goal is to reduce coupling with the Agda libraries. If a module uses the Agda AST, just the following import should be needed.

```haskell
import qualified FreeC.Backend.Agda.Syntax as Agda
```

Add an entry to [`.hlint.yaml`](https://github.com/FreeProving/free-compiler/blob/master/.hlint.yaml) to enforce this rule.

Smart constructors  for Agda AST nodes will be added to this module later.

### Dependencies

 - #30

JonasHoefer:
Adds the module [`FreeC.Backend.Agda.Syntax`](https://github.com/FreeProving/free-compiler/blob/issue27/src/lib/FreeC/Backend/Agda/Syntax.hs), which reexports the Agda AST to reduce coupling with the Agda libraries. Closes #27


just95 (approved):
Looks good to me

Issue #31, opened 2020-05-25
Add interface for Agda base library


just95:
### Description

In order to make the compiler more maintainable and less dependent on the base library, the Coq back end defines constants for identifiers of the free monad, its constructors and arguments in the [`FreeC.Backend.Coq.Base`](https://github.com/FreeProving/free-compiler/blob/master/src/lib/FreeC/Backend/Coq/Base.hs) module. We need a similar module for the Agda base library.

The interface module for the Agda base library is supposed to be imported as follows.

```haskell
import qualified FreeC.Backend.Agda.Base as Agda.Base
```

Add an entry to [`.hlint.yaml`](https://github.com/FreeProving/free-compiler/blob/master/.hlint.yaml) to enforce this rule.

### Dependencies

 - #30 
 - #28

PR #38, opened 2020-05-28
Add wrapper module for Agda pretty-printer

### LINKED ISSUE #29: Add wrapper module for Agda pretty-printer
### Description

We are using the [`Pretty`](http://hackage.haskell.org/package/wl-pprint-text-1.2.0.1/docs/Text-PrettyPrint-Leijen-Text.html#t:Pretty) type class from the [`wl-pprint-text`](http://hackage.haskell.org/package/wl-pprint-text) library for pretty printing (see also [`FreeC.Pretty`](https://github.com/FreeProving/free-compiler/blob/master/src/lib/FreeC/Pretty.hs)). The Agda library uses a different pretty printer. Similar to the pretty printer for the Haskell AST (see [`FreeC.Frontend.Haskell.Pretty`](https://github.com/FreeProving/free-compiler/blob/master/src/lib/FreeC/Frontend/Haskell/Pretty.hs)) we will have to provide a `Pretty` instance that wraps the corresponding pretty printer from the Agda library.

### Dependencies

 - #30

JonasHoefer:
Adds a `Pretty` instance for Agda AST nodes, as well as some (provisional) smart constructors in `FreeC.Backend.Agda.Syntax` to test the pretty printer. These constructors will probably change when work on #31 starts. Closes #29

### DIFF
--- a/src/lib/FreeC/Backend/Agda/Pretty.hs
+++ b/src/lib/FreeC/Backend/Agda/Pretty.hs
@@ -14,22 +14,23 @@ module FreeC.Backend.Agda.Pretty
   )
 where
 
--- We need just he pretty instances from 'Agda.Syntax.Concrete.Pretty'.
+-- We just need the pretty instances from 'Agda.Syntax.Concrete.Pretty'.
 import qualified Agda.Syntax.Concrete.Pretty    ( )
 import qualified Agda.Utils.Pretty             as Agda
 
 import qualified FreeC.Backend.Agda.Syntax     as Agda
 import           FreeC.Pretty
 
+
 -- | Wrapper data type that makes any 'Agda.Pretty' instance pretty printable.
 newtype PrettyAgda a = PrettyAgda { unPrettyAgda :: a }
 
--- | Pretty instance for all Agda internal pretty printable types.
+-- | Pretty instance for all Agda internal, pretty printable types.
 instance (Agda.Pretty a) => Pretty (PrettyAgda a) where
   pretty = prettyString . Agda.prettyShow . unPrettyAgda
 
 -- | Helper function for printing 'Agda.Pretty' printable types @a@ to a 'Doc'
---   used by our pretty printer.
+--   from the pretty printing library, which is used by the compiler.
 prettyAgda :: (Agda.Pretty a) => a -> Doc
 prettyAgda = pretty . PrettyAgda
 
@@ -37,6 +38,6 @@ prettyAgda = pretty . PrettyAgda
 instance Pretty Agda.Expr where
   pretty = prettyAgda
 
--- | 'Agda.Declaration's  are a common AST node.
+-- | 'Agda.Declaration's are a common AST node.
 instance Pretty Agda.Declaration where
   pretty = prettyAgda
--- a/src/lib/FreeC/Backend/Agda/Pretty.hs
+++ b/src/lib/FreeC/Backend/Agda/Pretty.hs
@@ -2,7 +2,7 @@
 
 -- | This module contains 'Pretty' instances for the Agda AST.
 --
---   The Agda has its own 'Agda.Pretty' class in 'Agda.Utils.Pretty'. Therefore
+--   The Agda library has its own 'Agda.Pretty' class in "Agda.Utils.Pretty". Therefore
 --   we cannot instance every @a@ with an @Agda.Pretty@ instance with our
 --   'Pretty' instance without generating overlapping instances for basic types
 --   (e.g. 'String').
--- a/src/lib/FreeC/Backend/Agda/Pretty.hs
+++ b/src/lib/FreeC/Backend/Agda/Pretty.hs
@@ -14,8 +14,6 @@ module FreeC.Backend.Agda.Pretty
   )
 where
 
--- import qualified Agda.Syntax.Concrete.Generic  as Agda
-
 -- We need just he pretty instances from 'Agda.Syntax.Concrete.Pretty'.
 import qualified Agda.Syntax.Concrete.Pretty    ( )
 import qualified Agda.Utils.Pretty             as Agda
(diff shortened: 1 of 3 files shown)


src/lib/FreeC/Backend/Agda/Pretty.hs
```diff
@@ -0,0 +1,44 @@
+{-# OPTIONS_GHC -Wno-orphans #-}
+
+-- | This module contains 'Pretty' instances for the Agda AST.
+--
+--   The Agda has its own 'Agda.Pretty' class in 'Agda.Utils.Pretty'. Therefore
```
just95: `Agda.Utils.Pretty` is a module. Thus, double quotes are needed in Haddock to link to it.

```suggestion
--   The Agda library has its own 'Agda.Pretty' class in "Agda.Utils.Pretty". Therefore
```

src/lib/FreeC/Backend/Agda/Pretty.hs
```diff
@@ -0,0 +1,44 @@
...
+--   we cannot instance every @a@ with an @Agda.Pretty@ instance with our
+--   'Pretty' instance without generating overlapping instances for basic types
+--   (e.g. 'String').
+--   We introduce the wrapper datatype 'PrettyAgda' to avoid the overlapping
+--   instances and explicitly declare 'Pretty' instances for for common AST
```
just95:
```suggestion
--   instances and explicitly declare 'Pretty' instances for common AST
```

src/lib/FreeC/Backend/Agda/Pretty.hs
```diff
@@ -0,0 +1,44 @@
...
+--   nodes.
+module FreeC.Backend.Agda.Pretty
+  ( prettyAgda
+  )
+where
+
+-- import qualified Agda.Syntax.Concrete.Generic  as Agda
```
just95: When commenting out code, try to use block comments to separate them from regular comments. Is there a reason why you kept this import declaration?

```suggestion
{- import qualified Agda.Syntax.Concrete.Generic  as Agda -}
```

src/lib/FreeC/Backend/Agda/Syntax.hs
```diff
@@ -3,7 +3,59 @@
...
+simpleImport modName = Agda.Import
+  Agda.NoRange
+  modName
+  Nothing
+  Agda.DoOpen
+  (Agda.ImportDirective Agda.NoRange Agda.UseEverything [] [] Nothing)
+
+-- | Creates a (not qualified) Agda variable name from a 'String'.
+name :: String -> Agda.Name
+name ident = Agda.Name Agda.NoRange Agda.InScope [Agda.Id ident]
```
just95: I would move this to the other smart constructors for names and maybe add some sections like the following.

```haskell
-- Names                                                                     --

…

-- Imports                                                                   --

…

-- Expressions                                                               --

…
```

just95 (changes requested):

src/lib/FreeC/Backend/Agda/Syntax.hs
```diff
@@ -3,7 +3,71 @@
...
 module FreeC.Backend.Agda.Syntax
   ( module Agda.Syntax.Concrete
+  , module Agda.Syntax.Common
+  , module Agda.Syntax.Position
+  , name
+  , intLiteral
+  , simpleImport
+  , lambda
+  , qname
+  , qname'
```
just95: I usually try to organize the export list in the same way I've organized the rest of the module.

```suggestion
    -- * Identifiers
  , name
  , qname
  , qname'
    -- * Imports
  , simpleImport
    -- * Expressions
  , intLiteral
  , lambda
```

src/lib/FreeC/Backend/Agda/Syntax.hs
```diff
@@ -3,7 +3,71 @@
...
+-------------------------------------------------------------------------------
+
+-- | Creates a (not qualified) Agda variable name from a 'String'.
+name :: String -> Agda.Name
+name ident = Agda.Name Agda.NoRange Agda.InScope [Agda.Id ident]
+
+-- | Create a qualified name given a local variable 'Name' and a list of module
+--   'Name's.
+qname :: Agda.Name -> [Agda.Name] -> QName
+qname n = foldr Agda.Qual (Agda.QName n)
```
just95: Since the module name is usually written before the name, I'd swap the arguments of `qname`.

just95 (changes requested):

src/lib/FreeC/Backend/Agda/Syntax.hs
```diff
@@ -3,7 +3,74 @@
...
 where
 
 import           Agda.Syntax.Concrete
+import           Agda.Syntax.Position
+import           Agda.Syntax.Common
+
+import qualified Agda.Syntax.Common            as Agda
+import qualified Agda.Syntax.Concrete          as Agda
+import qualified Agda.Syntax.Literal           as Agda
+import qualified Agda.Syntax.Position          as Agda
```
just95: Importing the modules both qualified and unqualified has the unfortunate side effect that the `Agda` prefix can be forgotten without GHC complaining (for example, I just noticed that `QName` in `qname` is not qualified). Since it could be confusing if we mix `QName`  and `Agda.QName` but only unqualified imports can be reexported, the best option is probably to remove the qualified imports and not qualify identifiers from the Agda library in this module. I'll followup with the corresponding change to `FreeC.Backend.Coq.Syntax`.

just95 (changes requested):

src/lib/FreeC/Backend/Agda/Syntax.hs
JonasHoefer: This sounds reasonable. I noticed the same thing but decided proceed in the same way as in `FreeC.Backend.Coq.Syntax`. Another advantage is that the code becomes more readable, because the smart constructors usually just provide lots of default arguments, which all had to be prefixed.

just95 (approved):

PR #39, opened 2020-05-28
Fix unification of rigid type variables

### LINKED ISSUE #13: Type inference unifies rigid type variables
### Description

Our type inference algorithm does not fail when two rigid type variables bound by a type signature are matched with each other.

### Steps to Reproduce

1. Consider the following Haskell module

    ```haskell
    module Test where

    foo :: a -> b -> (a, a)
    foo x y = (x, y)
    ```

2. Compile `Test` with `freec` and `ghc` using the following commands.

    ```bash
    freec ./Test.hs
    ghc ./Test.hs
    ```

#### Expected behavior

In the example above the type signature specifies that `y` is of type `b` but `(x, y)` is of type `(a, a)` and thus `y` is also of type `a`. Since the type variables `a` and `b` may be distinct, `foo` is not type correct. Thus both `freec` and `ghc` should reject the definition.

#### Actual behavior

The GHC rightly rejects the definition of `foo` with the following error message.

```
Test.hs:4:15: error:
    • Couldn't match expected type ‘a’ with actual type ‘b’
      ‘b’ is a rigid type variable bound by
        the type signature for:
          foo :: forall a b. a -> b -> (a, a)
        at Test.hs:3:1-23
      ‘a’ is a rigid type variable bound by
        the type signature for:
          foo :: forall a b. a -> b -> (a, a)
        at Test.hs:3:1-23
    • In the expression: y
      In the expression: (x, y)
      In an equation for ‘foo’: foo x y = (x, y)
    • Relevant bindings include
        y :: b (bound at Test.hs:4:7)
        x :: a (bound at Test.hs:4:5)
        foo :: a -> b -> (a, a) (bound at Test.hs:4:1)
  |
4 | foo x y = (x, y)
  |               ^
```

However, `freec` just unifies `a` and `b` and produces the following Coq module.

```coq
(* module Test *)
From Base Require Import Free.
From Base Require Import Prelude.

Definition foo (Shape : Type) (Pos : Shape -> Type) {a : Type} 
   (x : Free Shape Pos a) (y : Free Shape Pos a)
   : Free Shape Pos (Pair Shape Pos a a)
 := @Pair_ Shape Pos a a x y.
```

### Versions

 - `freec --version`:
   `The Free Compiler, version 0.1.0.0 (v0.1.0.0-527-gfe1b9b0, ghc 8.6, linux, x86_64)`
 - `ghc --version`: 8.6.5
 - `cabal --version`: 2.4.1.0
 - OS: Ubuntu 19.10

just95:
The unification algorithm that is used by the type inference pass was extended to report a fatal error if a rigid type variable bound by the type signature of a function is matched with another type. Closes #13

### DIFF
--- a/src/lib/FreeC/Frontend/Haskell/Simplifier.hs
+++ b/src/lib/FreeC/Frontend/Haskell/Simplifier.hs
@@ -26,6 +26,7 @@ where
 import           Control.Monad                  ( unless
                                                 , when
                                                 )
+import           Data.Composition               ( (.:) )
 import           Data.List.Extra                ( concatUnzip3 )
 import           Data.Maybe                     ( fromJust
                                                 , fromMaybe
@@ -39,6 +40,7 @@ import           FreeC.Frontend.IR.PragmaParser
 import qualified FreeC.IR.Base.Prelude         as IR.Prelude
 import           FreeC.IR.Reference             ( freeTypeVars )
 import           FreeC.IR.SrcSpan
+import           FreeC.IR.Subterm               ( findFirstSubterm )
 import qualified FreeC.IR.Syntax               as IR
 import           FreeC.Monad.Converter
 import           FreeC.Monad.Reporter
@@ -480,9 +482,30 @@ simplifyTypeSchema (HSE.TyForall srcSpan (Just binds) Nothing typeExpr) = do
 -- Without explicit @forall@.
 simplifyTypeSchema typeExpr = do
   typeExpr' <- simplifyType typeExpr
-  let srcSpan  = IR.typeSrcSpan typeExpr'
-      typeArgs = map (IR.TypeVarDecl NoSrcSpan) (freeTypeVars typeExpr')
+  let srcSpan         = IR.typeSrcSpan typeExpr'
+      typeArgIdents   = freeTypeVars typeExpr'
+      typeArgSrcSpans = map (findTypeArgSrcSpan typeExpr') typeArgIdents
+      typeArgs        = zipWith IR.TypeVarDecl typeArgSrcSpans typeArgIdents
   return (IR.TypeSchema srcSpan typeArgs typeExpr')
+ where
+  -- | Finds the first occurrence of the type variable with the given name.
+  --
+  --   Returns 'NoSrcSpan' if 'findTypeArgSrcSpan'' returns @Nothing@.
+  --   Since the type arguments have been extracted using 'freeTypeVars',
+  --   'findTypeArgSrcSpan'' should never return @Nothing@.
+  findTypeArgSrcSpan :: IR.Type -> IR.TypeVarIdent -> SrcSpan
+  findTypeArgSrcSpan = fromMaybe NoSrcSpan .: flip findTypeArgSrcSpan'
+
+  -- | Like 'findTypeArgSrcSpan' but returns @Nothing@ if there is
+  --   no such type variable.
+  findTypeArgSrcSpan' :: IR.TypeVarIdent -> IR.Type -> Maybe SrcSpan
+  findTypeArgSrcSpan' = fmap IR.typeSrcSpan .: findFirstSubterm . isTypeVar
+
+  -- | Tests whether the given type is the type variable with the given name.
+  isTypeVar :: IR.TypeVarIdent -> IR.Type -> Bool
+  isTypeVar typeVarIdent (IR.TypeVar _ typeVarIdent') =
+    typeVarIdent == typeVarIdent'
+  isTypeVar _ _ = False
 
 -- | Simplifies the a type expression.
 simplifyType :: HSE.Type SrcSpan -> Simplifier IR.Type
--- a/src/lib/FreeC/Frontend/Haskell/Simplifier.hs
+++ b/src/lib/FreeC/Frontend/Haskell/Simplifier.hs
@@ -481,9 +481,7 @@ simplifyTypeSchema (HSE.TyForall srcSpan (Just binds) Nothing typeExpr) = do
 simplifyTypeSchema typeExpr = do
   typeExpr' <- simplifyType typeExpr
   let srcSpan  = IR.typeSrcSpan typeExpr'
-      typeArgs = map
-        (IR.TypeVarDecl NoSrcSpan . fromJust . IR.identFromQName)
-        (freeTypeVars typeExpr')
+      typeArgs = map (IR.TypeVarDecl NoSrcSpan) (freeTypeVars typeExpr')
   return (IR.TypeSchema srcSpan typeArgs typeExpr')
 
 -- | Simplifies the a type expression.
--- a/src/lib/FreeC/IR/Reference.hs
+++ b/src/lib/FreeC/IR/Reference.hs
@@ -31,6 +31,7 @@ where
 import           Data.Composition               ( (.:) )
 import qualified Data.Foldable                 as OSet
                                                 ( toList )
+import           Data.Maybe                     ( fromJust )
 import           Data.Set                       ( Set )
 import qualified Data.Set                      as Set
 import           Data.Set.Ordered               ( OSet
@@ -92,6 +93,12 @@ refScope = fst . unwrapRef
 -- | Unwraps the given reference and discards the scope information.
 refName :: Ref -> IR.QName
 refName = snd . unwrapRef
+
+-- | Like 'refName' but unwraps the identifier of the name.
+--
+--   Fails if the given reference is a symbol.
+refIdent :: Ref -> String
+refIdent = fromJust . IR.identFromQName . refName
 
 -------------------------------------------------------------------------------
 -- Reference sets                                                            --
@@ -286,13 +293,13 @@ withoutArgs args set =
 -------------------------------------------------------------------------------
 
 -- | The type variables that occur freely in the given node from left to right.
-freeTypeVars :: HasRefs a => a -> [IR.QName]
-freeTypeVars = map refName . filter (isVarRef .&&. isTypeRef) . refs
+freeTypeVars :: HasRefs a => a -> [IR.TypeVarIdent]
+freeTypeVars = map refIdent . filter (isVarRef .&&. isTypeRef) . refs
 
 -- | The type variables that occur freely in the given node.
-freeTypeVarSet :: HasRefs a => a -> Set IR.QName
+freeTypeVarSet :: HasRefs a => a -> Set IR.TypeVarIdent
 freeTypeVarSet =
-  Set.map refName . Set.filter (isVarRef .&&. isTypeRef) . OSet.toSet . refSet
+  Set.map refIdent . Set.filter (isVarRef .&&. isTypeRef) . OSet.toSet . refSet
 
 -------------------------------------------------------------------------------
 -- Free variables                                                            --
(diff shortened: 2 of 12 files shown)


src/lib/FreeC/Frontend/Haskell/Simplifier.hs
```diff
@@ -480,11 +482,30 @@ simplifyTypeSchema (HSE.TyForall srcSpan (Just binds) Nothing typeExpr) = do
...
+ where
+  -- | Finds the first occurrence of the type variable with the given name.
+  --
+  --   Returns 'NoSrcSpan' if 'findTypeArgSrcSpan'' returns @Nothing@.
+  --   Since the type arguments have been extracted using 'freeTypeVars',
+  --   'findTypeArgSrcSpan'' should never return @Nothing@.
+  findTypeArgSrcSpan :: IR.Type -> IR.TypeVarIdent -> SrcSpan
+  findTypeArgSrcSpan = fromMaybe NoSrcSpan .: flip findTypeArgSrcSpan'
+
+  -- | Like 'findTypeArgSrcSpan' but returns @Nothing@ if there is
```
stu204767, line 499:
```suggestion
  -- | Like 'findTypeArgSrcSpan', but returns @Nothing@ if there is
```

src/lib/FreeC/IR/Reference.hs
```diff
@@ -93,6 +94,12 @@ refScope = fst . unwrapRef
 refName :: Ref -> IR.QName
 refName = snd . unwrapRef
 
+-- | Like 'refName' but unwraps the identifier of the name.
```
stu204767, line 97:
```suggestion
-- | Like 'refName', but unwraps the identifier of the name.
```

src/lib/FreeC/Pass/TypeInferencePass.hs
```diff
@@ -758,15 +811,18 @@ applyFuncDeclVisibly funcDecl = withLocalTypeAssumption $ do
...
+abstractTypeArgs :: [IR.TypeVarIdent] -> IR.FuncDecl -> IR.FuncDecl
+abstractTypeArgs typeArgIdents funcDecl =
+  let
+    typeArgs                   = map (IR.UnQual . IR.Ident) typeArgIdents
+    IR.TypeSchema _ _ typeExpr = fromJust (IR.funcDeclTypeSchema funcDecl)
+    (IR.TypeSchema _ typeArgs' _, subst) =
+      abstractTypeSchema' typeArgs typeExpr
+    funcDecl' = applySubst subst funcDecl
+  in
+    funcDecl' { IR.funcDeclTypeArgs = typeArgs' }
```
stu204767, line 825:
```suggestion
  in funcDecl' { IR.funcDeclTypeArgs = typeArgs' }
```

src/test/FreeC/Monad/Class/Testable.hs
```diff
@@ -83,14 +84,23 @@ shouldSucceedWith = flip (shouldReturnWith' (const "<expectation>")) id
 -- | Sets the expectation that the given computation fails without returning
 --   a value.
 shouldFail :: (Show a, MonadTestable m err) => m a -> Expectation
-shouldFail = flip shouldFailWith (const (return ()))
+shouldFail = flip shouldFailWith expectAnyError
+
+-- | Like 'shouldFail' but if the given computation does not fail, the
```
stu204767, line 89:
```suggestion
-- | Like 'shouldFail', but if the given computation does not fail, the
```

src/test/FreeC/Monad/Class/Testable.hs
```diff
@@ -83,14 +84,23 @@ shouldSucceedWith = flip (shouldReturnWith' (const "<expectation>")) id
...
+shouldFailPretty :: (Pretty a, MonadTestable m err) => m a -> Expectation
+shouldFailPretty mx = shouldFailWith' showPretty mx expectAnyError
 
 -- | Sets the expectation returned by the given function for the error
 --   that was produced by the given computation instead of a value.
 shouldFailWith
   :: (Show a, MonadTestable m err) => m a -> (err -> Expectation) -> Expectation
 shouldFailWith = shouldFailWith' show
 
+-- | Auxillary function for 'shouldFailWith' that allows arbitrary errors.
```
stu204767:
```suggestion
-- | Auxiliary function for 'shouldFailWith' that allows arbitrary errors.
```

src/test/FreeC/Pass/TypeInferencePassTests.hs
```diff
@@ -185,241 +185,275 @@ testTypeInferencePass = describe "FreeC.Analysis.TypeInference" $ do
...
-                ]
-              )
-              [ "length @a @b (xs :: List a) :: Prelude.Integer = case xs of {"
-                ++ "    Nil -> if true @b then 0 else 1;"
-                ++ "    Cons (x :: a) (xs' :: List a) ->"
-                ++ "      succ (length @a @b xs')"
-                ++ "  }"
-              ]
+  context "non-recursive functions" $ do
+    it "infers the types of simple simple non-recursive correctly"
```
stu204767:
```suggestion
    it "infers the types of simple non-recursive functions correctly"
```

stu204767 (changes requested):
Well written code, I couldn't find any major issues.
All of the suggestions I made are minor: Some typos, some (in my opinion) missing commas and one unnecessary line break.

src/lib/FreeC/Pass/TypeInferencePass.hs
just95: This suggestion cannot be applied directly since Brittany would change it back. I'm not exactly sure why Brittany wants to change the formatting in this case, though. Moving the binding of `typeArgs` down to the binding of `funcDecl'` seems to work. However, I usually try to sort `let`-bindings topologically. Thus, I'd stick with with the current formatting.

nbun (approved):
Looks good to me!

src/lib/FreeC/IR/Subterm.hs
```diff
@@ -261,6 +263,13 @@ findSubterms :: Subterm a => (a -> Bool) -> a -> [a]
 findSubterms predicate term =
   filter predicate (map (fromJust . selectSubterm term) (allPos term))
 
+-- | Gets the first subterm of the given expression that satisfies the
+--   provided predicate.
+--
+--   Return @Nothing@ if there is no such subterm.
+findFirstSubterm :: Subterm a => (a -> Bool) -> a -> Maybe a
+findFirstSubterm = listToMaybe .: findSubterms
+
```
ichistmeinname, line 272: After reading this definition, I was wondering why the used function `findSubterms` uses `fromJust` rather than `catMaybes` to get rid of `Nothing`s afterwards. Do you really need the run-time error of `fromJust` to be triggered here? Yes, it's safe, because all positions of `term` should yield a valid sub term, but still I'm not sure this is the best way to go here.

Additionally, I was wondering if using [`First`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Semigroup.html#t:First) isn't the usual way to define a function like `findFirstSubTerm`, but perhaps it only overcomplicates the code -- not sure about that.

src/lib/FreeC/Frontend/Haskell/Simplifier.hs
```diff
@@ -480,11 +482,30 @@ simplifyTypeSchema (HSE.TyForall srcSpan (Just binds) Nothing typeExpr) = do
...
+  --   no such type variable.
+  findTypeArgSrcSpan' :: IR.TypeVarIdent -> IR.Type -> Maybe SrcSpan
+  findTypeArgSrcSpan' = fmap IR.typeSrcSpan .: findFirstSubterm . isTypeVar
+
+  -- | Tests whether the given type is the type variable with the given name.
+  isTypeVar :: IR.TypeVarIdent -> IR.Type -> Bool
+  isTypeVar typeVarIdent (IR.TypeVar _ typeVarIdent') =
+    typeVarIdent == typeVarIdent'
+  isTypeVar _ _ = False
```
ichistmeinname, line 508: Brittany does not want these function rules to be aligned, huh?

ichistmeinname (changes requested):

src/lib/FreeC/IR/Subterm.hs
just95: I think the primary reason for using `fromJust` was consistency with `findSubtermPos` which cannot be expressed using `catMaybes`. The potential run-time error could actually be beneficial in this case. It allows us to spot mistakes in the implementation while `catMaybes` would silently ignore such bugs.

A better solution could be to add an alternative to `selectSubterm` that does not return a `Maybe` but prints are more useful error message than `fromJust` by mentioning the position of the subterm that could not be found. Such a function would be also useful since the pattern of removing `Maybe` from the result of `seletSubterm` can be found in [other parts of the code](https://github.com/FreeProving/free-compiler/blob/master/src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithHelpers.hs#L151) as well.

---

I'm not sure how you would use `First` in this case. It would be possible to replace `listToMaybe` with concatenation over the `Maybe (First a)` monoid. But as you've said already, that would just over complicate the code.

```haskell
import           Data.Semigroup                 ( First(..) )

findFirstSubterm :: Subterm a => (a -> Bool) -> a -> Maybe a
findFirstSubterm = fmap getFirst . mconcat . map (Just . First) .: findSubterms
```

src/lib/FreeC/Frontend/Haskell/Simplifier.hs
just95: As far as I know, Brittany does not align rules if one of them spans multiple lines.

src/lib/FreeC/Frontend/Haskell/Simplifier.hs
just95: According to [this article](https://www.grammarly.com/blog/comma-before-but/) a comma is put before *but* only if it connects two independent clauses. 
If the comment was `This function behaves like 'findTypeArgSrcSpan', but it returns @Nothing@ if there is no such type variable.`, I'd agree that a comma should be placed before *but* since both halves could be sentences on their own. However, in this case I'm not sure what's the correct punctuation. `Like 'findTypeArgSrcSpan'` doesn't sound like a full sentence to me. However, I could be wrong. @ichistmeinname  what do you think is the proper punctuation in this case?

src/lib/FreeC/IR/Subterm.hs
ichistmeinname: Should we add an issue for the modification of `selectSubterm` with a custom error message?

ichistmeinname (approved):

src/lib/FreeC/IR/Subterm.hs
just95: I've added an enhancement proposal #52.

src/lib/FreeC/Frontend/Haskell/Simplifier.hs
stu204767: I also looked at this article during my review and think that the comma should be there:
`Like 'findTypeArgSrcSpan'` may not be a full English sentence, but in this context it stands for a full sentence like the one you described and is therefore an independent clause.
This can also be seen by assuming `Like 'findTypeArgSrcSpan'.` were the entire comment for this function. While it would be strange to have two functions doing the same thing from a programming point of view, this would still be a sensible comment from a linguistic point of view.

stu204767 (approved):

PR #40, opened 2020-05-29
Implement error effect in Coq #18

### LINKED ISSUE #18: Implement error effect in Coq
## Description

Add a container instance that allows us to model computations in the [`Error`](https://github.com/FreeProving/free-compiler/blob/master/base/Free/Instance/Error.v)  monad to the `base` library.

## Motivation

Currently the `base` librarary contains `Container` instances for the [`Maybe`](https://github.com/FreeProving/free-compiler/blob/master/base/Free/Instance/Maybe.v) and [`Identity`](https://github.com/FreeProving/free-compiler/blob/master/base/Free/Instance/Identity.v) monads. Partial functions require an instance of the [`Partial`](https://github.com/FreeProving/free-compiler/blob/master/base/Free/Partial.v) type class. `Maybe` is the only instance of this type class at the moment.
However, sometimes the user wants to show that a specific error cannot occur or that only specific errors occur in their program. For this purpose, the `Maybe` monad is not sufficient since the `Nothing` constructor carries no information about the concrete error.
The `base` library contains a template for the [`Error`](https://github.com/FreeProving/free-compiler/blob/master/base/Free/Instance/Error.v) monad already but is currently not implemented.
The goal is to implement the `Error` monad such that it can be used as an alternative to `Maybe`.

## Tasks

 - [x] Add suitable definitions for `Shape` and `Pos` to `Error.v`.
 - [x] Implement the `Partial` type class for your definitions of `Shape` and `Pos`.
 - [x] Add an example for how to use the `Error` monad in proofs to the `example` directory.

## Getting Started

Have a look at the [accompanying code](https://github.com/FreeProving/free-proving-code) from the publication [One Monad to Prove Them All](https://arxiv.org/abs/1805.08059).

MarvinLira:
- Add implementation as described in [here](https://github.com/FreeProving/free-proving-code/blob/master/src/Effects.v).
- Add an example function in haskell that throws an error and a simple QuickCheck property for that function.
- Add proofs concerning error messages in addition to the proof of the QuickCheck property in the corresponding Coq file.

### DIFF
--- a/base/Free/Instance/Error.v
+++ b/base/Free/Instance/Error.v
@@ -1,5 +1,7 @@
 (** * Definition of the error monad in terms of the free monad. *)
 
+Require Import Coq.Strings.String.
+
 From Base Require Import Free.
 From Base Require Import Free.Util.Void.
 
@@ -16,7 +18,13 @@ Module Error.
       impure msg (fun (p : Pos msg) => match p with end).
   End Monad.
 
-  (* TODO Partial instance for the error monad. *)
+  (* Partial instance for the error monad. *)
+  Open Scope string_scope.
+  Instance Partial : Partial  (Shape string) Pos := {
+    undefined := fun {A : Type}                => ThrowError "undefined";
+    error     := fun {A : Type} (msg : string) => ThrowError msg
+  }.
+  Close Scope string_scope.
 End Error.
 
 (* The type and smart constructor should be visible to other modules
--- a/base/Free/Instance/Error.v
+++ b/base/Free/Instance/Error.v
@@ -1,11 +1,12 @@
 (** * Definition of the error monad in terms of the free monad. *)
 
 From Base Require Import Free.
+From Base Require Import Free.Util.Void.
 
 Module Error.
-  (* TODO Container instance that corresponds to [Const]. *)
-  (* Definition Shape : Type := (* ... *). *)
-  (* Definition Pos (s : Shape) : Type := (* ... *). *)
+  (* Container instance that corresponds to [Const]. *)
+  Definition Shape (E : Type) : Type := E.
+  Definition Pos {E : Type} (s : Shape E) : Type := Void.
 
   (* TODO Type synonym and smart constructors for the error monad. *)
   Module Import Monad.
--- a/base/Free/Instance/Error.v
+++ b/base/Free/Instance/Error.v
@@ -20,7 +20,7 @@ Module Error.
 
   (* Partial instance for the error monad. *)
   Open Scope string_scope.
-  Instance Partial : Partial  (Shape string) Pos := {
+  Instance Partial : Partial (Shape string) Pos := {
     undefined := fun {A : Type}                => ThrowError "undefined";
     error     := fun {A : Type} (msg : string) => ThrowError msg
   }.
--- a/base/Free/Instance/Error.v
+++ b/base/Free/Instance/Error.v
@@ -1,7 +1,5 @@
 (** * Definition of the error monad in terms of the free monad. *)
 
-Require Import Coq.Strings.String.
-
 From Base Require Import Free.
 From Base Require Import Free.Util.Void.
 
@@ -19,12 +17,10 @@ Module Error.
   End Monad.
 
   (* Partial instance for the error monad. *)
-  Open Scope string_scope.
   Instance Partial : Partial (Shape string) Pos := {
-    undefined := fun {A : Type}                => ThrowError "undefined";
+    undefined := fun {A : Type}                => ThrowError "undefined"%string;
     error     := fun {A : Type} (msg : string) => ThrowError msg
   }.
-  Close Scope string_scope.
 End Error.
 
 (* The type and smart constructor should be visible to other modules
--- a/base/Free/Instance/Error.v
+++ b/base/Free/Instance/Error.v
@@ -8,9 +8,12 @@ Module Error.
   Definition Shape (E : Type) : Type := E.
   Definition Pos {E : Type} (s : Shape E) : Type := Void.
 
-  (* TODO Type synonym and smart constructors for the error monad. *)
+  (* Type synonym and smart constructors for the error monad. *)
   Module Import Monad.
-
+    Definition Error (E A : Type) := Free (Shape E) Pos A.
+    Definition NoError {E A : Type} (x : A) : Error E A := pure x.
+    Definition ThrowError {E A : Type} (msg : E) : Error E A :=
+      impure msg (fun (p : Pos msg) => match p with end).
   End Monad.
 
   (* TODO Partial instance for the error monad. *)
--- a/example/Proofs/ConsUncons.hs
+++ b/example/Proofs/ConsUncons.hs
@@ -15,7 +15,7 @@ import           Data.List                      ( head )
 
 unconsE :: [a] -> (a, [a])
 unconsE ls = case ls of
-  []       -> error "uncons: empty list"
+  []       -> error "unconsE: empty list"
   (x : xs) -> (x, xs)
 
 -- Next we can define a simple property about this function, e.g., its relation
--- a/example/Proofs/ConsUncons.hs
+++ b/example/Proofs/ConsUncons.hs
@@ -24,7 +24,7 @@ unconsE ls = case ls of
 prop_cons_unconsE :: (Eq a, Show a) => a -> [a] -> Property
 prop_cons_unconsE x xs = unconsE (x : xs) === (x, xs)
 
--- And we define a property thats validity depends on the model used in the proof.
+-- And we define a property whose validity depends on the model used in the proof.
 
 fst :: (a, b) -> a
 fst p = case p of
--- a/example/Proofs/ConsUncons.hs
+++ b/example/Proofs/ConsUncons.hs
@@ -1,0 +1,23 @@
+-- | This example demonstrates how to use the translation of an error
+--   throwing function to prove properties about this function and about the
+--   possible errors that can be thrown.
+--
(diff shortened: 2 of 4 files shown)


base/Free/Instance/Error.v
```diff
@@ -1,18 +1,30 @@
 (** * Definition of the error monad in terms of the free monad. *)
 
+Require Import Coq.Strings.String.
+
```
just95: Since `Free` exports `Coq.Strings.String`, this import is not necessary. We should probably also remove the import from `Maybe.v`.

base/Free/Instance/Error.v
```diff
@@ -1,18 +1,30 @@
...
   End Monad.
 
-  (* TODO Partial instance for the error monad. *)
+  (* Partial instance for the error monad. *)
+  Open Scope string_scope.
+  Instance Partial : Partial (Shape string) Pos := {
+    undefined := fun {A : Type}                => ThrowError "undefined";
+    error     := fun {A : Type} (msg : string) => ThrowError msg
+  }.
+  Close Scope string_scope.
```
just95: Is there a reason why you use `Open Scope string_scope.` instead of `"undefined"%string`?

example/Proofs/ConsUncons.hs
```diff
@@ -0,0 +1,23 @@
...
+--   possible errors that can be thrown.
+--
+--   This example uses an error throwing implementation of the @uncons@
+--   function.
+
+module Proofs.ConsUncons where
+
+import           Test.QuickCheck
+
+-- First we have to define the @uncons@ function in an error throwing way.
```
just95: In the comments and in the error message you wrote `uncons` instead of `unconsE`.

example/Proofs/ConsUnconsProofs.v
```diff
@@ -0,0 +1,58 @@
...
+From Generated Require Import Proofs.ConsUncons.
+
+(* The proof for the QuickCheck property differs only in that way from proofs
+   about functions that do not return an error, that an [Partial] instance for
+   [Shape] and [Pos] is given. *)
+Lemma cons_unconsE : quickCheck prop_cons_unconsE.
+Proof.
+  intros Shape Pos Partial A fx fxs.
+  reflexivity.
+Qed.
```
just95: This is a good introductory example. However, you don't actually use the `Error` monad here. This proof holds for any monad with a `Partial` instance.  Thus, I think that you should add another proof to highlight the differences between `Maybe` and `Error`. 
For example, a property that involves several different error messages could be interesting. You could show for example that `fst (unconsE xs) === head xs` holds in the `Maybe` monad but not in the `Error` monad (since the error messages differ).

example/Proofs/ConsUnconsProofs.v
```diff
@@ -0,0 +1,58 @@
...
+  Lemma uncons_only_empty_list_error : forall (A : Type) (l : List _ _ A),
+    (exists (result : Pair _ _ A (List _ _ A)), unconsE _ _ _ (NoError l) = NoError result) \/
+    (                                           unconsE _ _ _ (NoError l) = EmptyListError).
+  Proof.
+    intros A l.
+    destruct l as [ | fx fxs ].
+    - right. reflexivity.
+    - left. simpl. exists (pair_ fx fxs). reflexivity.
+  Qed.
+End ErrorMessages.
```
just95: Please end source files with a newline.

just95 (changes requested):

example/Proofs/ConsUnconsProofs.v
```diff
@@ -0,0 +1,58 @@
...
+  reflexivity.
+Qed.
+
+Section ErrorMessages.
+
+  (* To prove facts about the error messages we can write an abbreviation for
+     an [error] with a specific message. *)
+  Open Scope string_scope.
+  Definition EmptyListError {A : Type} := @error _ _ _ A "uncons: empty list".
+  Close Scope string_scope.
```
ichistmeinname: I'd suggest to use `"string"%string`-syntax for all error messages instead of opening and closing scopes.

example/Proofs/ConsUnconsProofs.v
```diff
@@ -0,0 +1,58 @@
...
+  (* If we weren't looking for an actual [error] but for an [undefined] in haskell
+     we could use the following definition. *)
+  Open Scope string_scope.
+  Definition Undefined {A : Type} := @undefined _ _ _ A.
+  Close Scope string_scope.
+
+  (* Now we can define and prove the lemma that using [unconsE] with an empty
+     list results in an [EmptyListError] *)
+  Lemma nil_unconsE_empty_list_error : forall (A : Type),
+  @unconsE _ _ _ A (Nil _ _) = EmptyListError.
```
ichistmeinname: Here's an indentation missing.

base/Free/Instance/Error.v
```diff
@@ -1,18 +1,30 @@
...
-  (* Definition Pos (s : Shape) : Type := (* ... *). *)
+  (* Container instance that corresponds to [Const]. *)
+  Definition Shape (E : Type) : Type := E.
+  Definition Pos {E : Type} (s : Shape E) : Type := Void.
 
-  (* TODO Type synonym and smart constructors for the error monad. *)
+  (* Type synonym and smart constructors for the error monad. *)
   Module Import Monad.
-
+    Definition Error (E A : Type) := Free (Shape E) Pos A.
```
ichistmeinname, line 13: Not sure about this, but can this mix of implicit and explicit type arguments cause problems for generated code? 
Not specific to your PR, but more of a general question: how do I know which arguments to make explicit and which not; are these intricacies already documented?

ichistmeinname (changes requested):

base/Free/Instance/Error.v
MarvinLira: This file compiles correctly and doesn't cause errors when using it in my example. If you're referring to the code generated by freec, this should also cause no problems as freec only uses the `Partial` class and uses `@` to make all arguments explicit.
When deciding which arguments to make explicit, I make those arguments explicit that I would also need in haskell code and the remaining arguments implicit.

example/Proofs/ConsUnconsProofs.v
```diff
@@ -0,0 +1,78 @@
...
+(* The first QuickCheck property holds for any [Partial] instance. Therefore
+   its proof differs only in that way from proofs about functions that do not
+   return an error, that an [Partial] instance for [Shape] and [Pos] is given. *)
+Lemma cons_unconsE : quickCheck prop_cons_unconsE.
+Proof.
+  intros Shape Pos Partial A fx fxs.
+  reflexivity.
+Qed.
+
+(* The second QuickCheck property holds provable for the [Maybe] instance of [Partial]. *)
```
just95: I think `provable` should be an adverb here.

```suggestion
(* The second QuickCheck property holds provably for the [Maybe] instance of [Partial]. *)
```

base/Free/Instance/Error.v
ichistmeinname: Ah, right, thanks for clarification!

ichistmeinname (approved):
Good job!

example/Proofs/ConsUnconsProofs.v
```diff
@@ -0,0 +1,78 @@
...
+  intro H.
+  specialize (H bool (Nil _ _)).
+  discriminate H.
+Qed.
+
+Section ErrorMessages.
+
+  (* To prove facts about the error messages we can write an abbreviation for
+     an [error] with a specific message. *)
+  Definition EmptyListError {A : Type} := @error _ _ _ A "unconsE: empty list"%string.
```
just95: The reason `freec` uses explicit arguments for `Partial` instances is that implicit type class arguments behave counter-intuitively in Coq. In this case the `Error` instance is selected since it's the last instance imported at the top of the file.

```coq
From Base Require Import Free Free.Instance.Maybe Free.Instance.Error Test.QuickCheck Prelude.

…

Check EmptyListError.
(* ==> EmptyListError 
           : Free (Error.Shape string) Error.Pos ?A
       where
       ?A : [ |- Type]
*)
```

Suppose that we decide to sort imports in Coq alphabetically in the future as we do in Haskell already.

```coq
From Base Require Import Free Free.Instance.Error Free.Instance.Maybe Prelude Test.QuickCheck.
```

Now the import of the `Maybe` instance occurs last. Thus, the `Maybe` instance is inferred by Coq.

```coq
Check EmptyListError.
(* ==> EmptyListError
            : Free Maybe.Shape Maybe.Pos ?A
       where
       ?A : [ |- Type]
*)
```

However, all proofs below still hold in the `Maybe` monad  (including `unconsE_only_empty_list_error` which uses `NoError`) and the error goes unnoticed. Thus, I think that the `Partial` instance should be specified explicitly everywhere.

example/Proofs/ConsUncons.hs
```diff
@@ -0,0 +1,34 @@
...
+  []       -> error "unconsE: empty list"
+  (x : xs) -> (x, xs)
+
+-- Next we can define a simple property about this function, e.g., its relation
+-- to the constructor of an non-empty list.
+
+prop_cons_unconsE :: (Eq a, Show a) => a -> [a] -> Property
+prop_cons_unconsE x xs = unconsE (x : xs) === (x, xs)
+
+-- And we define a property thats validity depends on the model used in the proof.
```
just95:
```suggestion
-- And we define a property whose validity depends on the model used in the proof.
```

just95 (changes requested):

just95 (approved):

maclement (approved):
I have nothing to add or criticize. Well done!

PR #41, opened 2020-05-31
Add bang patterns to IR

### LINKED ISSUE #23: Add bang patterns to IR
## Description

By adding abstractions inspired by GHC's [`BangPatterns`](https://downloads.haskell.org/~ghc/8.6.5/docs/html/users_guide/glasgow_exts.html#extension-BangPatterns) language extension to the intermediate representation (IR), the translation of recursive functions to Coq can be simplified.

## Motivation

The `Environment` currently contains a field [`envDecArgs`](https://freeproving.github.io/free-compiler/docs/master/FreeC-Environment.html#v:envDecArgs). This field is used for two purposes.

 1. If the input module contains a function declaration

     ```haskell
     f x₁ … xᵢ … xₙ = e
     ```

     and one of the following pragmas 

     ```haskell
     {-# FreeC f DECREASES ON xᵢ #-}
     {-# FreeC f DECREASES ON ARGUMENT i #-}
     ```

     the `envDecArgs` map contains an entry that maps the name of `f` to `(i, "xᵢ")`. This information is used by [`identifyDecArgs`](https://freeproving.github.io/free-compiler/docs/master/FreeC-Backend-Coq-Analysis-DecreasingArguments.html#v:identifyDecArgs) to skip termination checking for `f` and assume `xᵢ` is its decreasing argument.

 2. Regardless of whether there is such an annotation for the decreasing argument or not, the result of `identifyDecArgs` is used to [split recursive functions](https://github.com/FreeProving/free-compiler/blob/010fba5d414a9184be056c30f6ced8bd7f1629ac/src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithHelpers.hs#L77) into helper and main functions. The `envDecArgs` field is manipulated such that the resulting non-recursive main function does not have an entry anymore (if one was added by a pragma before) and an entry is added for each recursive helper function. This information is used during the [translation of function applications](https://github.com/FreeProving/free-compiler/blob/010fba5d414a9184be056c30f6ced8bd7f1629ac/src/lib/FreeC/Backend/Coq/Converter/Expr.hs#L109) to insert a bind operation for calls to helper functions.

That `envDecArgs` has multiple purposes is not a good design decision and the management of the decreasing argument map that has to be performed in the second step is quite complicated and prone to errors . Thus, the goal is to get rid of the second use case.

What we actually want to do in the second use case is to mark that the helper functions is strict in its decreasing argument. In Haskell there is actually a way to express that a function is strict in specific arguments using the [`BangPatterns`](https://downloads.haskell.org/~ghc/8.6.5/docs/html/users_guide/glasgow_exts.html#extension-BangPatterns) language extension. If the IR supported bang patterns for function declarations, they could be used to model the strictness of helper functions.

## Tasks

 - [x] Add bang patterns to the IR.
    + [x] Extend the IR AST.
    + [x] Extend the IR parser.
    + [x] Extend the IR pretty-printer.
 - [x] Add information about strict arguments to [function environment entries](https://freeproving.github.io/free-compiler/docs/master/FreeC-Environment-Entry.html#v:FuncEntry).
 - [x] Update the translation of function applications and other translation rules.
 - [x] Remove uses of `envDecArgs` which are not related to the first use case.

> **Note:** The goal is to add bang patterns to IR not to add bang pattern support to the Haskell front end. `BangPatterns` are a Haskell language extension and will thus not supported by `freec`. However, don't focus on recursive functions in your test cases. Test you implementation with bang patterns in arbitrary positions. Most importantly, bang patterns should also be supported in `case` expressions.

### LINKED ISSUE #57: Main functions are wronly inlined as well
### Description

When a function `f` calls another function `g` for which (as expected) no corresponding helper function `g_0` exists inlining is still performed using `g` which leads to type errors.

### Steps to Reproduce

Compile the following Haskell module.

```haskell
module Left where

data Tree a = Leaf a | Node (Tree a) (Tree a)

{-# FreeC left DECREASES ON ARGUMENT 1 #-}
--left and left' are mutually recursive
left :: Tree Bool -> Bool
left t = left' t

left' :: Tree Bool -> Bool
left' (Leaf n    ) = n
left' (Node t1 t2) = left t1

--goLeft is an entry function into the mutual recursive circle
goLeft :: Tree Bool -> Bool
goLeft (Leaf n    ) = n
goLeft (Node t1 t2) = left t1
```

#### Expected behavior

The function `left` should simply be called on a the monadic argument.

```coq
Definition goLeft (Shape : Type) (Pos : Shape -> Type) (a7
    : Free Shape Pos (Tree Shape Pos (Bool Shape Pos)))
   : Free Shape Pos (Bool Shape Pos) :=
  a7 >>=
  (fun a7_0 =>
     match a7_0 with
     | leaf a8 => a8
     | node a10 a11 => left Shape Pos a10
     end).
```

#### Actual behavior

The `goLeft` function wrongly tries to inline `left`

```coq
Definition goLeft (Shape : Type) (Pos : Shape -> Type) (a7
    : Free Shape Pos (Tree Shape Pos (Bool Shape Pos)))
   : Free Shape Pos (Bool Shape Pos) :=
  a7 >>=
  (fun a7_0 =>
     match a7_0 with
     | leaf a8 => a8
     | node a10 a11 =>
         a10 >>= (fun (a10_0 : Tree Shape Pos (Bool Shape Pos)) => left Shape Pos a10_0)
     end).
```

### Versions

 - Git commit: 87ab7e5f5b74302d44e740c5ed2e19ce959bc9bf
 - `ghc --version`: 8.6.5
 - `cabal --version`: 3.2.0.0
 - `coqc --version`: 8.9.0
 - OS: Windows 10

Daniel-Teut:
Extends the IR of the compiler to add bang patterns and change the generation of helper functions, so that these bang patterns are used instead of the environment. Additionally change the function translation to also use the bang patterns. Closes #23.


src/lib/FreeC/IR/Syntax/Expr.hs
```diff
@@ -276,21 +276,30 @@ prettyExprPred' _ (Undefined _ _      ) = prettyString "undefined"
...
+--
+--   The left hand side can be annotated by a @!@.
 data Alt = Alt
   { altSrcSpan :: SrcSpan
   , altConPat  :: ConPat
   , altVarPats :: [VarPat]
   , altRhs     :: Expr
+  , altIsStrict  :: Bool
   }
  deriving (Eq, Show)
```
just95: After reading #23 again, I've got to admit that my wording was a bit ambiguous. With "bang patterns should also be supported in `case` expressions" I actually meant only variable patterns and not constructor patterns in `case` expressions. I originally did't consider constructor patterns since bang patterns only have an effect if they precede a variable pattern. Thus, I think that `altIsStrict` can be removed again. I'm sorry for the inconvenience. I'll try to be more clear in future issue descriptions.

src/lib/FreeC/Backend/Coq/Converter/Arg.hs
```diff
@@ -84,7 +84,7 @@ convertArg (IR.VarPat srcSpan ident maybeArgType) = do
 --   annotated with the converted type but the type is not lifted to the
 --   @Maybe@ monad.
 convertPureArg :: IR.VarPat -> Converter Coq.Binder
-convertPureArg (IR.VarPat srcSpan ident maybeArgType) = do
+convertPureArg (IR.VarPat srcSpan ident maybeArgType _) = do
   ident'        <- renameAndDefineVar srcSpan True ident maybeArgType
   maybeArgType' <- mapM convertType' maybeArgType
   generateArgBinder ident' maybeArgType'
```
just95: One advantage of the bang pattern notation is that we would only need `convertArg`. If `varPatIsStrict` is `True`, `convertArg` should behave like `convertPureArg`.

src/lib/FreeC/Backend/Coq/Converter/Expr.hs
```diff
@@ -117,7 +117,7 @@ convertExpr' (IR.Var srcSpan name _) typeArgs args = do
                     (return (Coq.app (Coq.Qualid qualid) partialArg))
       -- Is this a recursive helper function?
       Just arity   <- inEnv $ lookupArity IR.ValueScope name
-      mDecArgIndex <- inEnv $ lookupDecArgIndex name
+      mDecArgIndex <- inEnv $ lookupFirstStrictArgIndex name
```
just95: Even though there is at most one strict argument at the moment, you shouldn't make that assumption here since we may decide to use bang patterns for other purposes in the future. If a function is strict in two arguments, for example, both arguments must be bound before the function is invoked. Also add test cases for this behavior.

src/lib/FreeC/Backend/Coq/Converter/Expr.hs
```diff
@@ -295,6 +295,6 @@ convertConPat (IR.ConPat srcSpan ident) varPats = do
 --
 --   The types of variable patterns are not annotated in Coq.
 convertVarPat :: IR.VarPat -> Converter Coq.Pattern
-convertVarPat (IR.VarPat srcSpan ident maybeVarType) = do
+convertVarPat (IR.VarPat srcSpan ident maybeVarType _) = do
   ident' <- renameAndDefineVar srcSpan False ident maybeVarType
```
just95: The flag of `renameAndDefineVar` that is set to `False` here indicates whether a value is in the `Free` monad or not. If the variable is annotated with a bang pattern, it should not be lifted. Thus, you have to pass `varPatIsStrict` to `renameAndDefineVar` such that the compiler knows that the variable needs to be wrapped with `pure` if it occurs in a position where a lifted value is expected. Also a bind needs to be inserted in `convertAlt` to implement the semantic of the bang pattern. Add test cases for bang patterns in `case` expressions.

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithHelpers.hs
```diff
@@ -153,16 +148,18 @@ transformRecFuncDecl (IR.FuncDecl srcSpan declIdent typeArgs args maybeRetType e
...
       argTypeMap       = Map.fromList (zip argNames argTypes)
       helperArgTypeMap = boundVarTypeMap `Map.union` argTypeMap
       helperArgTypes =
         map (join . (`Map.lookup` helperArgTypeMap)) helperArgNames
-      helperArgs = zipWith
+      helperArgs = zipWith3
         (IR.VarPat NoSrcSpan . fromJust . IR.identFromQName)
         helperArgNames
         helperArgTypes
+        (map (== decArg) helperArgNames)
```
just95: This could be a problem if arguments of the original IR function are marked as strict already.

src/lib/FreeC/IR/Similar.hs
```diff
@@ -363,7 +363,7 @@ instance Similar IR.Alt where
 --   > ———————————  ——————————————————————————
 --   >  Γ ⊢ x ≈ y    Γ ⊢ (x :: τ) ≈ (y :: τ')
 instance Similar IR.VarPat where
-  similar' (IR.VarPat _ _ t1) (IR.VarPat _ _ t2) = similar' t1 t2
+  similar' (IR.VarPat _ _ t1 _) (IR.VarPat _ _ t2 _) = similar' t1 t2
```
just95: Two variable patterns are not similar if only one of them has a bang pattern. The following two additional inference rules should hold.

```haskell
--   > —————————————  ————————————————————————————
--   >  Γ ⊢ !x ≈ !y    Γ ⊢ !(x :: τ) ≈ !(y :: τ')
```

just95 (changes requested):

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithHelpers.hs
```diff
@@ -153,16 +149,23 @@ transformRecFuncDecl (IR.FuncDecl srcSpan declIdent typeArgs args maybeRetType e
         helperArgNames = Set.toList (usedVars `Set.intersection` boundVars)
 
     -- Determine the type of helper function's arguments and its return type.
+    -- Additionally the decreasing argument is marked as strict.
```
stu204767:
```suggestion
    -- Additionally, the decreasing argument is marked as strict.
```

src/lib/FreeC/IR/Similar.hs
```diff
@@ -362,8 +362,15 @@ instance Similar IR.Alt where
 --   >                      Γ ⊢ τ ≈ τ'
 --   > ———————————  ——————————————————————————
 --   >  Γ ⊢ x ≈ y    Γ ⊢ (x :: τ) ≈ (y :: τ')
+--
+--   If one of the patterns has a bang pattern the other one needs one aswell.
```
stu204767:
```suggestion
--   If one of the patterns has a bang pattern, the other one needs one as well.
```

src/lib/FreeC/IR/Syntax/Expr.hs
```diff
@@ -321,11 +330,13 @@ instance Pretty ConPat where
 -- | A variable pattern used as an argument to a function, lambda abstraction
 --   or constructor pattern.
 --
---   The variable pattern can optionally have a type signature.
+--   The variable pattern can optionally have a type signature
+--   or can be annotated by a @!@.
```
stu204767:
```suggestion
--   or be annotated by a @!@.
```

src/test/FreeC/IR/SimilarTests.hs
```diff
@@ -268,6 +268,10 @@ testSimilarFuncDecls = context "function declarations" $ do
     fx <- expectParseTestFuncDecl "f (x :: Foo) = x"
     fy <- expectParseTestFuncDecl "f y = y"
     fx `shouldNotBeSimilarTo` fy
+  it "functions with strict and non strict arguments are not similar" $ do
```
stu204767:
```suggestion
  it "functions with strict and non-strict arguments are not similar" $ do
```

src/test/FreeC/Frontend/IR/ParserTests.hs
```diff
@@ -368,56 +368,107 @@ testIfExprParser = context "if expressions" $ do
...
       `shouldParse` IR.Case
                       NoSrcSpan
                       s
-                      [ IR.Alt NoSrcSpan fooPat [] x
-                      , IR.Alt NoSrcSpan barPat [] y
+                      [ IR.Alt NoSrcSpan fooPat [] x False
+                      , IR.Alt NoSrcSpan barPat [] y False
                       ]
                       Nothing
+  it "accepts case expressions with one strict and one non strict alternative"
```
stu204767:
```suggestion
  it "accepts case expressions with one strict and one non-strict alternative"
```

src/test/FreeC/Frontend/IR/ParserTests.hs
```diff
@@ -368,56 +368,107 @@ testIfExprParser = context "if expressions" $ do
...
                             Nothing
+  it "accepts case expressions with strict variable patterns" $ do
+    "case s of { Foo !x !y -> x }"
+      `shouldParse` IR.Case
+                      NoSrcSpan
+                      s
+                      [IR.Alt NoSrcSpan fooPat [xPatStrict, yPatStrict] x False]
+                      Nothing
+  it
+      "accepts case expressions with strict variable patterns and strict alternative"
```
stu204767:
```suggestion
  it "accepts case expressions with strict variable patterns and a strict alternative"
```
(The line break might be added again by Brittany)

src/test/FreeC/Frontend/IR/ParserTests.hs
```diff
@@ -368,56 +368,107 @@ testIfExprParser = context "if expressions" $ do
...
+                          ]
+                          Nothing
   it "accepts case expressions with type annotated variable patterns" $ do
     "case s of { Foo (x :: a) -> x }"
       `shouldParse` IR.Case NoSrcSpan
                             s
-                            [IR.Alt NoSrcSpan fooPat [xPat'] x]
+                            [IR.Alt NoSrcSpan fooPat [xPat'] x False]
                             Nothing
+  it "accepts case expressions with strict type annotated variable patterns"
```
stu204767:
```suggestion
  it "accepts case expressions with strict type-annotated variable patterns"
```

src/test/FreeC/Frontend/IR/ParserTests.hs
```diff
@@ -508,22 +559,38 @@ testIntLiteralParser = context "integer literals" $ do
...
-  it "accepts function declarations with a type annotated arguments" $ do
+  it "accepts function declarations with multiple strict arguments" $ do
+    "f !x !y = x"
+      `shouldParse` IR.FuncDecl NoSrcSpan
+                                f
+                                []
+                                [xPatStrict, yPatStrict]
+                                Nothing
+                                x
+  it "accepts function declarations with a type annotated argument" $ do
```
stu204767:
```suggestion
  it "accepts function declarations with a type-annotated argument" $ do
```

src/test/FreeC/Frontend/IR/ParserTests.hs
```diff
@@ -508,22 +559,38 @@ testIntLiteralParser = context "integer literals" $ do
...
     "f (x :: a) = x" `shouldParse` IR.FuncDecl NoSrcSpan f [] [xPat'] Nothing x
+  it "accepts function declarations with a strict type annotated argument" $ do
```
stu204767:
```suggestion
  it "accepts function declarations with a strict type-annotated argument" $ do
```

stu204767 (changes requested):
I made some minor suggestions (mostly missing hyphens in comments), but your code looks very good in total.

src/test/FreeC/Test/Environment.hs
```diff
@@ -151,12 +152,14 @@ defineTestVar nameStr = do
...
-defineTestFunc' :: Bool -> String -> Int -> String -> Converter String
-defineTestFunc' partial nameStr arity typeStr = do
+defineTestFunc nameStr arity =
+  defineTestFunc' False (replicate arity False) nameStr arity
+
+-- | Like 'defineTestFunc' but the second argument controls whether the
+--   defined function is partial or not. The first argument controls the
+--   strictness of the function arguments.
+defineTestFunc' :: Bool -> [Bool] -> String -> Int -> String -> Converter String
+defineTestFunc' partial areStrict nameStr arity typeStr = do
```
just95: The documentation does not match the implementation.

```suggestion
-- | Like 'defineTestFunc' but the first argument controls whether the
--   defined function is partial or not. The second argument controls the
--   strictness of the function arguments.
defineTestFunc' :: Bool -> [Bool] -> String -> Int -> String -> Converter String
defineTestFunc' partial areStrict nameStr arity typeStr = do
```

src/test/FreeC/Test/Environment.hs
```diff
@@ -176,7 +180,13 @@ defineTestFunc' partial nameStr arity typeStr = do
...
 --   Returns the Coq identifier assigned to the function.
 definePartialTestFunc :: String -> Int -> String -> Converter String
-definePartialTestFunc = defineTestFunc' True
+definePartialTestFunc nameStr arity =
+  defineTestFunc' True (replicate arity False) nameStr arity
+
+-- | Like 'defineTestFunc' but also allows to mark arguments as strict in the
+--   first argument.
+defineStrictTestFunc :: [Bool] -> String -> Int -> String -> Converter String
+defineStrictTestFunc = defineTestFunc' False
```
just95: Since the list contains one element for every argument, the `arity` wouldn't have to be passed to this function explicitly.

```haskell
defineStrictTestFunc :: String -> [Bool] -> String -> Converter String
```

src/lib/FreeC/IR/Syntax/Expr.hs
```diff
@@ -321,11 +321,13 @@ instance Pretty ConPat where
...
 --   or constructor pattern.
 --
---   The variable pattern can optionally have a type signature.
+--   The variable pattern can optionally have a type signature
+--   or be annotated by a @!@.
 data VarPat = VarPat
   { varPatSrcSpan :: SrcSpan
   , varPatIdent   :: String
   , varPatType    :: Maybe Type
+  , varPatIsStrict  :: Bool
```
just95:
```suggestion
data VarPat = VarPat
  { varPatSrcSpan   :: SrcSpan
  , varPatIdent     :: String
  , varPatType      :: Maybe Type
  , varPatIsStrict  :: Bool
```

src/lib/FreeC/IR/Similar.hs
```diff
@@ -362,8 +362,15 @@ instance Similar IR.Alt where
...
+--
+--   If one of the patterns has a bang pattern, the other one needs one as well.
+--
+--   >                        Γ ⊢ τ ≈ τ'
+--   > —————————————  ————————————————————————————
+--   >  Γ ⊢ !x ≈ !y    Γ ⊢ !(x :: τ) ≈ !(y :: τ')
 instance Similar IR.VarPat where
-  similar' (IR.VarPat _ _ t1) (IR.VarPat _ _ t2) = similar' t1 t2
+  similar' (IR.VarPat _ _ t1 s1) (IR.VarPat _ _ t2 s2) renaming =
+    s1 == s2 && similar' t1 t2 renaming
```
just95: In the other `Similar` instances a point-free notation is used.

```suggestion
  similar' (IR.VarPat _ _ t1 s1) (IR.VarPat _ _ t2 s2) =
    const (s1 == s2) .&&. similar' t1 t2
```

src/lib/FreeC/Backend/Coq/Converter/Expr.hs
```diff
@@ -115,26 +116,18 @@ convertExpr' (IR.Var srcSpan name _) typeArgs args = do
...
-            generateApplyN arity callee (before ++ decArg' : after)
+      Just arity         <- inEnv $ lookupArity IR.ValueScope name
+      Just strictArgs    <- inEnv $ lookupStrictArgs name
+      -- Add type annotations for strict arguments.
+      Just typeArgIdents <- inEnv $ lookupTypeArgs IR.ValueScope name
+      Just argTypes      <- inEnv $ lookupArgTypes IR.ValueScope name
+      let typeArgNames = map (IR.UnQual . IR.Ident) typeArgIdents
+          subst = composeSubsts (zipWith singleSubst typeArgNames typeArgs)
+          decArgTypes = map (applySubst subst) argTypes
+      decArgTypes' <- mapM (mapM convertType') decArgTypes
```
just95: The name `decArgTypes` doesn't really fit anymore.

just95 (changes requested):

src/lib/FreeC/Environment.hs
```diff
@@ -138,6 +137,20 @@ removeDecArg :: IR.QName -> Environment -> Environment
...
   env { envDecArgs = Map.delete funcName (envDecArgs env) }
 
+-------------------------------------------------------------------------------
+-- Modifying entries in the environment                                      --
+-------------------------------------------------------------------------------
+
+-- | Changes the Coq identifier of the entry with the given name in the given
+--   scope to the given identifier.
+--
+--   If such an entry does not exists, the environment is not changed.
```
stu204767:
```suggestion
--   If such an entry does not exist, the environment is not changed.
```

src/lib/FreeC/Backend/Coq/Converter/Free.hs
```diff
@@ -61,6 +61,40 @@ genericApply func effectArgs implicitArgs args
 generatePure :: Coq.Term -> Converter Coq.Term
 generatePure = return . Coq.app (Coq.Qualid Coq.Base.freePureCon) . (: [])
 
+-- | Generates a Coq expression that binds the given values to fresh variables
+--   if the position in the given @[Bool]@ list is @True@.
+--
+--   The fresh variables and the not bound values are passed to the given
+--   function in their original order. The fresh variables are not visible
+--   outside this function. If a value is a variable, the name of that variable
+--   is used as the prefix of the corresponding fresh variable. Otherwise the
```
stu204767:
```suggestion
--   is used as the prefix of the corresponding fresh variable. Otherwise, the
```

stu204767 (changes requested):

stu204767 (approved):

src/lib/FreeC/Backend/Coq/Converter/Expr.hs
```diff
@@ -115,26 +116,18 @@ convertExpr' (IR.Var srcSpan name _) typeArgs args = do
...
-            generateApplyN arity callee (before ++ decArg' : after)
+      Just arity         <- inEnv $ lookupArity IR.ValueScope name
+      Just strictArgs    <- inEnv $ lookupStrictArgs name
+      -- Add type annotations for strict arguments.
+      Just typeArgIdents <- inEnv $ lookupTypeArgs IR.ValueScope name
+      Just argTypes      <- inEnv $ lookupArgTypes IR.ValueScope name
+      let typeArgNames = map (IR.UnQual . IR.Ident) typeArgIdents
+          subst = composeSubsts (zipWith singleSubst typeArgNames typeArgs)
+          strictArgTypes = map (applySubst subst) argTypes
+      strictArgTypes' <- mapM convertType' strictArgTypes
```
just95: The name `strictArgTypes` isn't ideal either. Even though only the elements are used where `strictArgs` is set to `True`, the list itself does not just contain types of `strictArgs` but of all `args`. I'd suggest the following.

```suggestion
      argTypes' <- mapM (convertType' . applySubst subst) argTypes
```

src/test/FreeC/Test/Environment.hs
```diff
@@ -176,7 +180,14 @@ defineTestFunc' partial nameStr arity typeStr = do
...
 --   Returns the Coq identifier assigned to the function.
 definePartialTestFunc :: String -> Int -> String -> Converter String
-definePartialTestFunc = defineTestFunc' True
+definePartialTestFunc nameStr arity =
+  defineTestFunc' True (replicate arity False) nameStr arity
+
+-- | Like 'defineTestFunc' but also allows to mark arguments as strict in the
+--   first argument.
+defineStrictTestFunc :: [Bool] -> String -> String -> Converter String
+defineStrictTestFunc areStrict nameStr =
```
just95: I would prefer the order of arguments to be swapped such that the function is consistent with how you usually write a function declaration in IR, i.e, `<name> <args> :: <type>`.

```suggestion
defineStrictTestFunc :: String -> [Bool] -> String -> Converter String
defineStrictTestFunc nameStr areStrict =
```

src/test/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithHelpersTests.hs
```diff
@@ -501,6 +501,83 @@ testConvertRecFuncDeclWithHelpers = context "with helper functions" $ do
...
+        "List"           <- defineTestTypeCon "List" 1
+        ("nil" , "Nil" ) <- defineTestCon "Nil" 0 "forall a. List a"
+        ("cons", "Cons") <- defineTestCon "Cons"
+                                          2
+                                          "forall a. a -> List a -> List a"
+        "interleave" <- defineTestFunc "interleave"
+                                       2
+                                       "forall a. List a -> List a -> List a"
+        shouldConvertWithHelpersTo
+            [ "interleave @a (xs :: List a) !(ys :: List a) :: List a ="
```
just95, line 551: I think it would be also interesting to have an example where the bang pattern precedes the decreasing argument. Such an example could reveal a problem in the generation of the `struct` annotation.

just95 (changes requested):

just95 (approved):

PR #42, opened 2020-06-01
Add check for complete pattern matching #21

### LINKED ISSUE #21: Add check for complete pattern matching
## Description

The compiler should report an error if a `case` expression does not contain exactly one alternative for every constructor of the matched data type.

## Motivation

One assumption of the compiler is that pattern matching is complete, i.e., there is one alternative for every constructor of the matched data type in a `case` expression. If the [pattern-matching compiler](https://github.com/FreeProving/haskell-src-transformations/) is enabled, the `case` completion phase will add missing alternatives. However, pattern-matching compilation is optional and even if it is enabled, there are [known bugs](https://github.com/FreeProving/haskell-src-transformations/issues/7) in the pattern-matching compiler that cause the generation of code that violates the assumption. Since the compiler does not check the assumption, the generated Coq code is invalid if the assumption is not met. For a user it may be difficult to understand the error reported by Coq. Thus, a check should be implemented into the Free Compiler directly.

## Tasks

 - [x] Add a field to [environment entries for data type declarations](https://freeproving.github.io/free-compiler/docs/master/FreeC-Environment-Entry.html#v:DataEntry) with the names of its constructors.
     + [x] Update the [encoder](https://freeproving.github.io/free-compiler/docs/master/FreeC-Environment-ModuleInterface-Encoder.html) and [decoder](https://freeproving.github.io/free-compiler/docs/master/FreeC-Environment-ModuleInterface-Decoder.html) for module interface files accordingly. 
     + [x] Don't forget to bump the module interface file format  version number.
     + [x] Identify where such entries are added to the environment and initialize the new field appropriatly.
 - [x] Add a compiler [`Pass`](https://freeproving.github.io/free-compiler/docs/master/FreeC-Pass.html#t:Pass) to the compiler [`pipeline`](https://freeproving.github.io/free-compiler/docs/master/FreeC-Pipeline.html#v:pipeline) that performs the check. Perform at least the following checks and report a fatal error if any check fails.
    + [x] The type of the `case` expression's scrutinee is a data type (i.e., not a function).
    + [x] There is one alternative for every constructor of the data type.
    + [x] There are not multiple alternatives for the same constructor of the data type.
 - [x] Add documentation and test cases for your compiler `Pass`.

maclement:
Add a field to environment entries for data type declarations with the names of its constructors.
Add a compiler pass to the compiler pipeline that performs the check. 
Add documentation and test cases for your compiler pass.


MajaRet:
Right now, the check does not consider nested case expressions or case expressions that are used as another case expression's scrutinee. 

For example, the check accepts the following definition:

```
data Foobar = Foo | Bar
f x = case x of 
           Foo -> case x of 
                           Foo -> Foo
           Bar -> Bar 
```
As well as the following definition.

```
f x = case (case x of 
                     Foo -> Foo) of 
            Foo -> Foo 
            Bar -> Bar  
```

MajaRet (changes requested):
Right now, the check does not consider nested `case` expressions or `case` expressions that are used as another `case` expression's scrutinee. 

For example, the check accepts the following definition:

```
data Foobar = Foo | Bar
f x = case x of 
           Foo -> case x of 
                      Foo -> Foo
           Bar -> Bar 
```
The following definition is accepted as well.

```
f x = case (case x of 
                Foo -> Foo) of 
            Foo -> Foo 
            Bar -> Bar  
```

src/lib/FreeC/Environment/Entry.hs
```diff
@@ -1,6 +1,6 @@
...
---   in the environment.
+-- | This module contains data types that are used to store information



+--   about declared functions, (type) variables and (type) constructors



+--   in the environment.
```
just95: The line endings have to be fixed. Not sure why the pipeline wasn't triggered when this PR was marked as "Ready for Review".

just95 (changes requested):

src/lib/FreeC/Environment/Entry.hs
maclement: I will fix all line endings. Somehow my fix for brittany didn't work aswell.

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,123 @@
...









+--   >                  ,Some b -> Some (True)}
```
just95: The alternatives of `case` expressions are separated by semicolons, not commas. IR `case` expressions that span multiple lines are formatted as follows in the remaining documentation and test cases.

```haskell
--   > redundant :: Just Bool -> Just Bool
--   > redundant (on :: Just Bool) = case (on :: Just Bool) of {
--   >     Some a -> Some False;
--   >     None   -> None;
--   >     Some b -> Some True
--   >   }
```

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,123 @@
...
+  checkPatternExpr :: IR.Expr -> Converter ()
+  checkPatternExpr (IR.Case srcSpan exprScrutinee exprAlts _) = do
+    checkPatternExpr exprScrutinee
+    mapM_ (checkPatternExpr . IR.altRhs) exprAlts
+    case IR.exprType (exprScrutinee) >>= getTypeConName of
+      Nothing       -> failedPatternCheck srcSpan
+      Just typeName -> do
+        env <- getEnv
+        let altConNames = map (IR.conPatName . IR.altConPat) exprAlts
+        case lookupEntry IR.TypeScope typeName env of
```
just95:
```suggestion
        maybeEntry <- inEnv $ lookupEntry IR.TypeScope typeName
        let altConNames = map (IR.conPatName . IR.altConPat) exprAlts
        case maybeEntry of
```

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,123 @@
...


+checkPatternFuncDecl :: IR.FuncDecl -> Converter ()
+checkPatternFuncDecl funcDecl = checkPatternExpr (IR.funcDeclRhs funcDecl)
+ where
+  checkPatternExpr :: IR.Expr -> Converter ()
+  checkPatternExpr (IR.Case srcSpan exprScrutinee exprAlts _) = do
+    checkPatternExpr exprScrutinee
+    mapM_ (checkPatternExpr . IR.altRhs) exprAlts
+    case IR.exprType (exprScrutinee) >>= getTypeConName of
```
just95:
```suggestion
    case IR.exprType exprScrutinee >>= getTypeConName of
```

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,123 @@
...
+          Just entry | isDataEntry entry ->
+            performCheck (entryConsNames entry) altConNames srcSpan
+          Just entry | isTypeSynEntry entry ->
+            case
+                getTypeConName (entryTypeSyn entry)
+                  >>= \e -> lookupEntry IR.TypeScope e env
+              of
+                Just entry' ->
+                  performCheck (entryConsNames entry') altConNames srcSpan
+                _ -> failedPatternCheck srcSpan
```
just95: I don't think that this handles situations like the following correctly.

```haskell
type Foo = Bar
type Bar = Baz
data Baz = Baz

case (x :: Foo) of { … }
```

I would recommend to have a look at [`expandAllTypeSynonyms` from `FreeC.IR.TypeSynExpansion`](https://freeproving.github.io/free-compiler/docs/master/FreeC-IR-TypeSynExpansion.html#v:expandAllTypeSynonyms).

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,123 @@
...









+--   for each constructor of the corresponding type.
```
just95: Haddock comments should start with a single sentence summary. Additional details are provided in subsequent paragraphs.

```suggestion
-- | Checks that all functions of a given module have complete pattern matching.
--
--   The pattern matching is complete if there is exactly one case alternative
--   for each constructor of the corresponding type.
```

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,123 @@
...









+
```
just95, line 63: What are pre- and post- conditions of this pass? Is the AST modified in some way? In which cases are errors reported?

```haskell
--   = Specification
--
--   == Preconditions
--
--   TODO
--
--   == Translation
--
--   TODO
--
--   == Postconditions
--
--   TODO
--
--   == Error cases
--
--   * TODO

```

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,123 @@
...




+completePatternPass :: Pass IR.Module
+completePatternPass ast = do
+  mapM_ checkPatternFuncDecl (IR.modFuncDecls ast)
+  return ast
+
+-- | Checks a funcDecl for incomplete Patterns
```
just95: If you use Haskell identifiers in text, use markup for links `'…'` or code spans `@…@`.

```suggestion
-- | Checks a 'IR.FuncDecl' for incomplete patterns.
```

In this case you could also give a textual description instead. 

```suggestion
-- | Checks a function declaration for incomplete patterns.
```

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,123 @@
...






+checkPatternFuncDecl :: IR.FuncDecl -> Converter ()
+checkPatternFuncDecl funcDecl = checkPatternExpr (IR.funcDeclRhs funcDecl)
+ where
+  checkPatternExpr :: IR.Expr -> Converter ()
```
just95, line 98: Even though they are not visible in the generated Haddock documentation, I'd also add documentation comments to local declarations in `where` blocks.

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,123 @@
...
+  failedPatternCheck :: SrcSpan -> Converter ()
+  failedPatternCheck srcSpan =
+    reportFatal
+      $  Message srcSpan Error
+      $  "Incomplete pattern in function: "
+      ++ stringFromName (IR.funcDeclName funcDecl)
+
+  stringFromName :: IR.Name -> String
+  stringFromName (IR.Ident  s) = s
+  stringFromName (IR.Symbol s) = s
```
just95: Use [`showPretty` from `FreeC.Pretty`](https://freeproving.github.io/free-compiler/docs/master/FreeC-Pretty.html#v:showPretty) to embed IR AST nodes in error messages.

```suggestion
      ++ showPretty (IR.funcDeclName funcDecl)
```

just95 (changes requested):

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,154 @@
...





+  unfoldTypeSyn :: EnvEntry -> [IR.ConName] -> SrcSpan -> Converter ()
+  unfoldTypeSyn entry altConNames srcSpan | isDataEntry entry = do
+    performCheck (entryConsNames entry) altConNames srcSpan
+  unfoldTypeSyn entry altConNames srcSpan | isTypeSynEntry entry = do
+    case getTypeConName (entryTypeSyn entry) of
```
just95: Consider the following type synonym.

```haskell
type Id a = a
```

The pattern matching in the following `case` expression is complete, but no `DataEntry` will be found by `unfoldTypeSyn`.

```haskell
case (x :: Id [a]) of
  []     -> …
  x : xs -> …
```

Try to reuse [`expandAllTypeSynonyms`](https://freeproving.github.io/free-compiler/docs/master/FreeC-IR-TypeSynExpansion.html#v:expandAllTypeSynonyms).

just95 (changes requested):

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,154 @@
...

+--







+--   The type of all checked expressions has to be annotaded.
```
MajaRet:
```suggestion
--   The type of all checked expressions has to be annotated.
```

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,154 @@
...









+--   for each constructor of the corresponding type.
```
MajaRet: Since pattern matching has to be complete in each occurrence of a `case` expression on the right-hand side of a function declaration, perhaps mention that _all_ occurring case expressions must have exactly one case alternative for each constructor of the corresponding type.

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,154 @@
...

+--







+--   declarations have complete patten macthing. A pattern ist complete if there
```
MajaRet:
```suggestion
--   declarations have complete pattern matching. A pattern is complete if there
```

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,154 @@
...









+--   The following declarations
```
MajaRet:
```suggestion
--   The module consisting of the following declarations
```

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,154 @@
...









+--   should not pass the check.
```
MajaRet:
```suggestion
--   should not pass the check because the @case@ expression is missing an alternative for the constructor @Nothing@.
```

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,154 @@
...









+--   The following declaration with redundant alternavies
```
MajaRet:
```suggestion
--   The following declaration with redundant alternatives
```

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,154 @@
...









+--   >     Some b -> Some (True)}
```
MajaRet: Shouldn't the constructors be `Just` and `Nothing`? The brackets around `True` and `False` aren't necessary, either.
```suggestion
--   >     Just a -> Just False;
--   >     Nothing -> Nothing;
--   >     Just b -> Just True}
```

src/lib/FreeC/Pass/CompletePatternPass.hs
maclement: Oh, I searched for something similar to `expandAllTypeSynonyms` but I couldn't find it.

MajaRet (changes requested):
Aside from some small things I marked (plus the remaining Windows line endings), the code and documentation look good to me!

src/lib/FreeC/Pass/CompletePatternPass.hs
maclement: Yes you are right. Seems like an old version of this example.

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -1,44 +1,46 @@
 -- | This module contains a compiler pass that checks if all function
---   declarations have complete patten macthing. A pattern ist complete if there
+--   declarations have complete pattern macthing. A pattern is complete if there
```
MajaRet:
```suggestion
--   declarations have complete pattern matching. A pattern is complete if there
```

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -1,44 +1,46 @@
...
+--   should not pass the check because the @case@ expression has two alternatives
+--   for the constructor @Just@.
 --
 --   == Example 3
 --   The following declaration where the scrutinee is a function
 --
 --   > case_id = case \x -> x  of
 --
---   should not pass the check.
+--   should not pass the check because the type of the scrutinee.
```
MajaRet:
```suggestion
--   should not pass the check because functions are not permitted as scrutinees.
```

MajaRet (changes requested):
There are still some (very minor) things left, but if you change those, I'll be quite happy : )

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,146 @@
...
+--   The Environment has to contain the names of all constructors for
+--   all used data types.
+--   Additionally, the environment should contain entries for all used type
+--   synonyms.
+--
+--   == Translation
+--
+--   This pass only performs a check and therefore does not change the module.
+--   == Postconditions
+--
```
just95: I'd leave a blank line before and after headings (including after the headings in the example section).

```suggestion
--   This pass only performs a check and therefore does not change the module.
--
--   == Postconditions
--
```

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,146 @@
...
+    tau' <- expandAllTypeSynonyms tau
+    case getTypeConName tau' of
+      Nothing       -> failedPatternCheck srcSpan
+      Just typeName -> do
+        maybeEntry <- inEnv $ lookupEntry IR.TypeScope typeName
+        let altConNames = map (IR.conPatName . IR.altConPat) exprAlts
+        case maybeEntry of
+          Just entry | isDataEntry entry ->
+            performCheck (entryConsNames entry) altConNames srcSpan
+          _ -> failedPatternCheck srcSpan
```
just95: The error message printed by `failedPatternCheck` does not describe the actual error in this case very well. The catch all case applies if `maybeEntry` is `Nothing` (i.e., `typeName` is not in the environment) or the entry is not a `DataEntry`. The case that the entry is missing can be handled using [`lookupEntryOrFail`](https://freeproving.github.io/free-compiler/docs/master/FreeC-Environment-LookupOrFail.html#v:lookupEntryOrFail). That the entry is a `DataEntry` should already be guaranteed by `expandAllTypeSynonyms` (i.e., it cannot be a `TypeSynEntry`) and `getTypeConName` (i.e., it cannot be a `TypeVarEntry`). Thus, I think you can omit the second check. However, you should state the assumption in a comment.

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,146 @@
...
+-- | Checks a function declaration for incomplete patterns.
+checkPatternFuncDecl :: IR.FuncDecl -> Converter ()
+checkPatternFuncDecl funcDecl = checkPatternExpr (IR.funcDeclRhs funcDecl)
+ where
+  -- | Checks an expression for incomplete patterns.
+  checkPatternExpr :: IR.Expr -> Converter ()
+  checkPatternExpr (IR.Case srcSpan exprScrutinee exprAlts _) = do
+    checkPatternExpr exprScrutinee
+    mapM_ (checkPatternExpr . IR.altRhs) exprAlts
+    let tau = fromJust $ IR.exprType exprScrutinee -- is safe beacause all types are annotated
```
just95: Try to write full sentences in comments. I'd also write the comment before the line. Writing comments at the end of lines just tempts one to write shorter comments (especially due to the 80 character limit).

just95 (changes requested):

MajaRet (approved):

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,150 @@
...
+--   > case_id = case \x -> x  of
+--
+--   should not pass the check because functions are not permitted as scrutinees.
+--
+--   = Specification
+--
+--   == Preconditions
+--
+--   The type of all checked expressions has to be annotated.
+--   The Environment has to contain the names of all constructors for
```
just95: If you mean the `Environment` data type write, `'Environment'` in Haddock.

```suggestion
--   The 'Environment' has to contain the names of all constructors for
```

If you mean the work "environment", it shouldn't be capitalized.

```suggestion
--   The environment has to contain the names of all constructors for
```

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -0,0 +1,150 @@
...
+--
+--   == Example 2
+--
+--   The following declaration with redundant alternatives
+--
+--   > redundant :: Just Bool -> Just Bool
+--   > redundant (on :: Just Bool) = case (on :: Just Bool) of {
+--   >     Just a  -> Just False;
+--   >     Nothing -> Nothing;
+--   >     Just b  -> Just True}
```
just95:
```suggestion
--   >     Just b  -> Just True

```

just95 (changes requested):

just95 (approved):

PR #43, opened 2020-06-01
Add Partial to reserved identifiers #16

### LINKED ISSUE #16: Custom declarations of `Partial` are not renamed automatically
### Description

The compiler produces a type error when compiling the following example program.

```haskell
module NList where

data Partial a = Value a
               | Undefined

data NList a = Cons a (Partial (NList a))

toNList :: a -> [a] -> NList a
toNList val [] = Cons val Undefined
toNList val [x] = Cons x Undefined
toNList val (x:xs) = Cons x (Value (toNList val xs))
```

### Steps to Reproduce

1. compile `NList.hs` with `freec -o NList NList.hs`
2. open created coq file in coqide
3. go through definitions with coq

#### Expected behavior

The created file should be free of type errors.

#### Actual behavior

When trying to add the function `toNList_1` the following type error occurs:

```
In environment
Shape : Type
Pos : Shape -> Type
a_0 : Type
a19_0 : Free Shape Pos a_0
The term "Partial Shape Pos" has type "Type -> Type" which should be Set, Prop or Type.
```

### Versions

 - `freec --version`: version 0.1.0.0
 - `ghc --version`: version 8.6.5
 - `cabal --version`:  3.2.0.0
 - `coqc --version`:  8.11.0
 - OS: Ubuntu 19.10
### Known workarounds
None

stu204146-kiel:


src/lib/FreeC/Backend/Coq/Base.hs
```diff
@@ -56,6 +56,10 @@ freeArgs =
 -- Partiality                                                                --
 -------------------------------------------------------------------------------
 
+-- | The name of the type class @Partial@
```
Daniel-Teut:
```suggestion
-- | The Coq identifier for the @Partial@ type class.
```
I would change the wording here slightly to be more in line with the rest of the comments.

Daniel-Teut (changes requested):

Daniel-Teut (approved):

just95 (approved):

PR #44, opened 2020-06-01
Remove `Maybe` from `Environment` entries #19

### LINKED ISSUE #19: Remove `Maybe` from `Environment` entries
## Description

The `entryArgTypes` and `entryReturnType` fields of the [`EnvEntry`](https://freeproving.github.io/free-compiler/docs/master/FreeC-Environment-Entry.html#t:EnvEntry) data type should contain `IR.Type`s instead of `Maybe IR.Type`s.

## Motivation

The`entryArgTypes` and `entryReturnType` fields  had to be lifted to the  `Maybe` monad because the environment contains entries for generated helper functions but not all of their argument and return types were known. Since the compiler supports type inference now, we also know the types of helper functions. Thus, it should be possible to remove `Maybe` from those two fields.

## Tasks

 - [x] Update the `EnvEntry` data type.
 - [x] Fix resulting type errors.
 - [x] Investigate whether the compiler generates untyped AST nodes after type inference.
 - [x] If there are such untyped nodes, update the code such that their type is properly annotated.

stu204767:
As described in #19, it is no longer neccessary to encapsulate the `entryArgTypes` and `entryReturnType` fields of the [environment entries](https://freeproving.github.io/free-compiler/docs/master/FreeC-Environment-Entry.html#t:EnvEntry) in the `Maybe` type, because the type inference can infer these type now. This pull request removes these `Maybe`s, fixes the resulting type errors and adds tests by @marvin2706 regarding these changes. Closes #19.


src/lib/FreeC/Backend/Coq/Converter/Expr.hs
```diff
@@ -132,7 +132,7 @@ convertExpr' (IR.Var srcSpan name _) typeArgs args = do
           let typeArgNames = map (IR.UnQual . IR.Ident) typeArgIdents
               subst = composeSubsts (zipWith singleSubst typeArgNames typeArgs)
               decArgType = applySubst subst (argTypes !! index)
-          decArgType' <- mapM convertType' decArgType
+          decArgType' <- mapM convertType' (Just decArgType)
```
nbun: Instead of `mapM`-ing into a `Just`, you can apply the function directly.

src/lib/FreeC/Pass/DefineDeclPass.hs
```diff
@@ -113,8 +115,8 @@ defineFuncDecl funcDecl = do
     { entrySrcSpan       = IR.funcDeclSrcSpan funcDecl
     , entryArity         = length (IR.funcDeclArgs funcDecl)
     , entryTypeArgs = map IR.typeVarDeclIdent (IR.funcDeclTypeArgs funcDecl)
-    , entryArgTypes      = map IR.varPatType (IR.funcDeclArgs funcDecl)
-    , entryReturnType    = IR.funcDeclReturnType funcDecl
+    , entryArgTypes = map (fromJust . IR.varPatType) (IR.funcDeclArgs funcDecl)
```
nbun, line 118: I'm assuming that the `fromJust` ist needed due to the `Map.lookup` operation yielding `Maybe`. Would it make sense to switch to a total map instead?

nbun (changes requested):

src/lib/FreeC/Pass/DefineDeclPass.hs
stu204767: I don't see a `Map.lookup` here. A [`FuncDecl`](https://freeproving.github.io/free-compiler/docs/master/FreeC-IR-Syntax-FuncDecl.html#t:FuncDecl) contains a list of [`VarPat`](https://freeproving.github.io/free-compiler/docs/master/FreeC-IR-Syntax-Expr.html#t:VarPat)s, in which a [`Type`](https://freeproving.github.io/free-compiler/docs/master/FreeC-IR-Syntax-Type.html#t:Type) is contained in a `Maybe`. `IR.varPatType` and `IR.funcDeclArgs` are just the record field names.
Or do you mean something else?

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithSections.hs
```diff
@@ -466,8 +466,8 @@ updateTypeSig mgu constTypeVars argTypeMap returnTypeMap funcDecl = do
   let allTypeArgs = map IR.typeVarIdent typeArgVars
       entry'      = entry { entryArity         = length args
                           , entryTypeArgs      = allTypeArgs \\ constTypeVars
-                          , entryArgTypes      = argTypes
-                          , entryReturnType    = returnType
+                          , entryArgTypes      = map fromJust argTypes
```
nbun, line 467: > I'm assuming that `fromJust` is needed due to the `Map.lookup` operation yielding `Maybe`. Would it make sense to switch to a total map instead?

The above comment was referring to this line since `argTypes` is built using `Map.lookup`.

src/lib/FreeC/Pass/DefineDeclPass.hs
nbun: Sorry, my comment was referencing another change but somehow ended up here. I have moved the comment there.

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithSections.hs
stu204767: Do you mean changing `argTypeMap` to a map that is total by having a default value? I don't think that would be useful here, because it would be difficult to come up with a useful default value other than `error`.
The other kind of total maps I have seen only work with Enum-types as keys, which we definitely can't use here.

However, `fromJust` could be avoided by using the [`(!)`](https://hackage.haskell.org/package/containers-0.6.2.1/docs/Data-Map-Strict.html#v:-33-)-operator instead of `lookup` to build `argTypes` and `returnType`. This would slightly reduce the length of the code, but the location where errors potentially occur would be less explicit, so I don't know if this change is desired.

MarvinLira (approved):

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithSections.hs
nbun: I see. The default value I had in mind was indeed `error`. If we assume that the lookup never fails and, thus, we do not need the ability to distinguish between success and failure at the call site, I would prefer a total map. In addition, moving `error` to the map would allow adding the key to the error message. However, I am not sure whether this information is actually helpful.

stu204767: Okay. But to add the key to the error message, we would need a default function of type `k -> a` instead of a default value of type `a` (assuming a map of type `Map k a`). The packages providing data types for total maps I have found (like [this](https://hackage.haskell.org/package/total-map)) only use default values instead of functions.
I would therefore suggest to avoid using an additional package and changing the type of the map itself, but instead use a new `lookup`-like function that uses `Map.lookup` and returns the result of this lookup without `Maybe` or `error` with an error message containing the key.
(The existing function `Map.findWithDefault` can't be used because it too works with default values and not functions).

just95: Since the code in this module is very lengthy already, I'd argue against complicating it even more by tweaking some internal error messages. The use of `fromJust` looks safe to me in this scenario. I think that it is more important that fields of environment entries are evaluated strictly such that potential errors are thrown in the first place. However, that's beyond the scope of issue #19.

just95 (approved):

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithSections.hs
nbun: So the information about the key and map for which the lookup failed is not useful? My issue with `fromJust` is that you essentially lose all information associated with how the error occurred and only know that _something_ went wrong. But this is probably a discussion worth having in its own issue, I'll mark this as resolved here.

nbun (approved):

ichistmeinname:
@stu204767 Due to some merges, there are new conflicts that need to be tackled! Furthermore, please add a description for this merge and a reference to the issue that can be closed after the merge.

PR #45, opened 2020-06-02
 Add interface for Agda base library

### LINKED ISSUE #31: Add interface for Agda base library

JonasHoefer:
Adds the module `FreeC.Backend.Agda.Base`, which contains Haskell representations for important parts of the Agda base library. At the moment `>>=` is missing, because there are multiple ways to define operators. Closes #31


src/lib/FreeC/Backend/Agda/Base.hs
```diff
@@ -0,0 +1,75 @@
...
+impureConName = Agda.name "impure"
+
+impure :: Agda.Expr -> Agda.Expr
+impure = Agda.app $ Agda.Ident $ Agda.qname [Agda.name "Free"] impureConName
+
+shape :: Agda.Name
+shape = Agda.name "S"
+
+position :: Agda.Name
+position = Agda.name "P"
```
ichistmeinname: These type arguments are named `Shape` and `Pos` in the Coq backend.

ichistmeinname (changes requested):

just95 (approved):

ichistmeinname (approved):
Good work!

PR #46, opened 2020-06-02
Avoid unnecessary monadic lifting of partial applications

### LINKED ISSUE #3: Functions defined as other functions are wrapped too often
When defining functions as other (partial) functions the compiled code adds too many monad wrappings.

For example the accumulative reverse

```haskell
rev :: [a] -> [a]
rev = rev' []

rev' :: [a] -> [a] ->[a]
rev' xs [] = xs
rev' xs (y:ys) = rev' (y:xs) ys
```

becomes

```coq
Definition rev (Shape : Type) (Pos : Shape -> Type) {a : Type}
   : Free Shape Pos (Free Shape Pos (List Shape Pos a) ->
           Free Shape Pos (List Shape Pos a)) :=
  pure (fun x_0 => @rev' Shape Pos a (@Nil Shape Pos a) x_0).
```

but should be 

```coq
Definition rev (Shape : Type) (Pos : Shape -> Type) {a : Type}
   : Free Shape Pos (List Shape Pos a) ->
           Free Shape Pos (List Shape Pos a) :=
  pure (fun x_0 => @rev'_0 Shape Pos a (@Nil Shape Pos a) x_0).
```

or 

```coq
... :=    @rev' Shape Pos a (@Nil Shape Pos a).
```

MajaRet:
Previously, the compiler performed eta abstractions any time it encountered a partial
application on a function declaration's right-hand side. These abstractions were done
by transforming the partial application into a lambda abstraction that includes the
missing arguments.

When this happens on the top level (i.e. a function is defined as a partial application
of another function or constructor), this leads to an additional lifting into the Free
monad.
This can be avoided by applying the right-hand side to the missing arguments and
adding them to the left-hand side of the function rule explicitly.

The change is implemented in the compiler's `EtaConversionPass`.

Closes #3


MajaRet:
I'm not sure why the Brittany check fails. Running the check-formatting.sh script changed 22 files for some reason (hence the 22 changed files), but the check still fails.

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -31,17 +44,27 @@
...
 --
---   On the right-hand sides of function declarations, all of the largest
---   sub-expressions of the form
+--   Assume that we have the following function declaration. 
+--
+--   > f e₁ … eₖ = g @α₁ … @αₚ e₁ … eₘ
+--
+--   where @g@ is the name of an @n@-ary constructor or function declaration and 
+--   @m < n@. This declaration is then replaced by 
+--   > f e₁ … eₖ x₍ₘ₊₁₎ … xₙ = @α₁ … @αₚ g e₁ … eₘ x₍ₘ₊₁₎ … xₙ
```
just95: Just a small typo.

```suggestion
--   > f e₁ … eₖ x₍ₘ₊₁₎ … xₙ = g @α₁ … @αₚ e₁ … eₘ x₍ₘ₊₁₎ … xₙ
```

Very good job on the documentation btw!

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -78,27 +104,67 @@ etaConversionPass ast = do
...
+--   The missing top-level arguments are also added to the left-hand side of the 
+--   declaration and the function's type and the environment are updated accordingly. 
 etaConvertFuncDecl :: IR.FuncDecl -> Converter IR.FuncDecl
 etaConvertFuncDecl funcDecl = do
-  rhs' <- etaConvertExpr (IR.funcDeclRhs funcDecl)
-  return funcDecl { IR.funcDeclRhs = rhs' }
+  let rhs = IR.funcDeclRhs funcDecl
+  xs   <- generateFullArgumentList rhs
+  rhs' <- etaConvertTopLevelRhs xs rhs
+  -- compute the function's new (uncurried) type
```
just95: All comments should be proper sentences.

```suggestion
  -- Compute the function's new (uncurried) type.
```

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -78,27 +104,69 @@ etaConversionPass ast = do
...
+-- | Applies a function declaration's or constructor's right-hand side to its 
+--   missing arguments.
+--   Regular η-conversions are performed for the right-hand side's proper subexpressions 
+--   via 'etaConvertSubExprs'.
+etaConvertTopLevelRhs :: [IR.VarPat] -> IR.Expr -> Converter IR.Expr
+etaConvertTopLevelRhs argPats expr = localEnv $ do
+  expr' <- etaConvertSubExprs expr
+  if null argPats
+    then return expr'
+    else return (IR.app NoSrcSpan expr' argExprs)
```
just95: You don't need to test whether there are any arguments. `IR.app` returns the expression if the argument list is empty.

```suggestion
  return (IR.app NoSrcSpan expr' argExprs)
```

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -78,27 +104,69 @@ etaConversionPass ast = do
...
+  where argExprs = map IR.varPatToExpr argPats
```
just95: Since you have a `do` block above, I'd use a `let` statement in this case.

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -78,27 +104,69 @@ etaConversionPass ast = do
...
+--   The missing top-level arguments are also added to the left-hand side of the 
+--   declaration and the function's type and the environment are updated accordingly. 
 etaConvertFuncDecl :: IR.FuncDecl -> Converter IR.FuncDecl
 etaConvertFuncDecl funcDecl = do
-  rhs' <- etaConvertExpr (IR.funcDeclRhs funcDecl)
-  return funcDecl { IR.funcDeclRhs = rhs' }
+  let rhs = IR.funcDeclRhs funcDecl
+  xs          <- generateFullArgumentList rhs
+  returnType' <- inEnv
+    $ lookupReturnType IR.ValueScope (IR.funcDeclQName funcDecl)
```
just95: I'd recommend using the return type annotation from `funcDecl` directly.

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -78,27 +104,69 @@ etaConversionPass ast = do
...
+  let (args, returnType) = splitReturnType (length xs) (fromJust returnType')
+  -- compute the function's new arguments and add them to the argument list 
+  let newArgs = zipWith
+        (\t i -> IR.VarPat { IR.varPatSrcSpan = NoSrcSpan
+                           , IR.varPatIdent   = i
+                           , IR.varPatType    = Just t
+                           }
+        )
+        args
+        xs
```
just95: This should be equivalent to the following.

```suggestion
  let newArgs = zipWith (IR.VarPat NoSrcSpan) xs (map Just args)
```

A better name for `args` could be `newArgTypes` and for `xs` a name like `newArgIdents` would be more descriptive.

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -169,3 +238,26 @@ arityOf (IR.Undefined _ _      ) = return 0
...
+--   @τ₁ -> (… -> (τₙ -> τ))@, which is true for the return type of 
+--   a partially defined function with @n@ missing arguments.
+--   The function returns a tuple consisting of a list of the argument
+--   types @[τ₁, …, τₙ]@ and the return type @τ@. 
+splitReturnType :: Int -> IR.Type -> ([IR.Type], IR.Type)
+splitReturnType 0 t = ([], t)
+splitReturnType n t =
+  let (arg , r    ) = IR.splitFuncType t 1
+      (args, rType) = splitReturnType (n - 1) r
+  in  (arg ++ args, rType)
```
just95: I don't get how this is different from `IR.splitFuncType` (except that the arguments are flipped).

Also, the assumption that the type has the form `τ₁ -> (… -> (τₙ -> τ))` is not necessarily true in the presence of type synonyms. The [`TypeSignaturePass`](https://github.com/FreeProving/free-compiler/blob/4627f930212624096a61ad051b557c0690f4d838/src/lib/FreeC/Pass/TypeSignaturePass.hs#L187) deals with this issue by expanding type synonyms on demand. I think that the most reliable solution would be to move the `etaConversionPass` above the `typeSignaturePass` in the compiler [`pipeline`](https://github.com/FreeProving/free-compiler/blob/4627f930212624096a61ad051b557c0690f4d838/src/lib/FreeC/Pipeline.hs#L25). This would have the advantage, that you wouldn't have to annotate any types or modify environment entries. However, this change would be more involved and is not really in the scope of a first issue. Thus, I'd suggest that we just recycle the definition of `FreeC.Pass.TypeSignaturePass.splitFuncType` for now.

src/test/FreeC/Pass/EtaConversionPassTests.hs
```diff
@@ -32,6 +33,20 @@ shouldEtaConvert inputStr expectedOutputStr = do
...
+        _ <- defineTestTypeCon "Foo" 0
+        _ <- defineTestFunc "f" 0 "Foo -> Foo"
+        _ <- defineTestFunc "g" 2 "Foo -> Foo -> Foo"
+        "f = g Foo" `shouldEtaConvert` "f (y :: Foo) :: Foo = g Foo y"
+
+-- These are the old tests, but since the testing interface has changed
+-- and we now test whole function declarations instead of right-hand sides,
+-- these tests don't work anymore.
+-- I'll leave them here for reference for now, but they should be removed later.
+{-
```
just95: All of the old test cases should still work. I'd suggest to export `etaConvertExpr` again and move the new and existing test cases to their own [`context`](https://hackage.haskell.org/package/hspec-2.7.1/docs/Test-Hspec.html#v:context)es.

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -78,27 +104,69 @@ etaConversionPass ast = do
...
+                           }
+        )
+        args
+        xs
+  let vars' = IR.funcDeclArgs funcDecl ++ newArgs
+  -- compute the new right-hand side
+  rhs'       <- etaConvertTopLevelRhs newArgs rhs
+  -- update the environment with the new type and arguments
+  Just entry <- inEnv $ lookupEntry IR.ValueScope (IR.funcDeclQName funcDecl)
+  modifyEnv $ addEntry entry { entryArity      = length vars'
```
just95: Good catch that the environment entries need to be updated : )

However, when the arity of a function in the environment changes, additional η-conversions become necessary. Therefore, η-conversion has to be performed recursively.

just95 (changes requested):

src/lib/FreeC/Pass/EtaConversionPass.hs
MajaRet: I did that at first, but then the unit tests didn't work because apparently, `IR.funcDeclReturnType` contains `Nothing` when you define a test function.

just95: Okay, that makes sense. However, in the test cases you could just annotate the return type and add to the preconditions of the pass that the function declarations must be type annotated.

```haskell
  "f :: Foo -> Foo = g Foo" `shouldEtaConvert` "f (y :: Foo) :: Foo = g Foo y"
```

src/lib/FreeC/Pass/EtaConversionPass.hs
MajaRet: It isn't; I simply misunderstood what `IR.splitFuncType` actually does. I've changed it so that `etaConvertFuncDecl` simply uses `IR.splitFuncType` now.

src/test/FreeC/Pass/EtaConversionPassTests.hs
```diff
@@ -18,47 +18,70 @@ import           FreeC.Test.Expectations
...
+      "f x y z" `shouldEtaConvert` "f x y z"
+    it "eta-converts under applied functions" $ shouldSucceedWith $ do
+      _ <- defineTestTypeCon "Foo" 0
+      _ <- defineTestFunc "f" 2 "Foo -> Foo -> Foo"
+      "f x" `shouldEtaConvert` "\\y -> f x y"
+    it "leaves application of local variables unchanged"
+      $                  shouldSucceedWith
+      $                  do
+                           "\\(f :: a -> b -> c) x -> f x"
       `shouldEtaConvert` "\\(f :: a -> b -> c) x -> f x"
```
just95: In this case it probably better to write the expectation in prefix notation such that Brittany does not mess up the formatting.

src/lib/FreeC/Pass/EtaConversionPass.hs
just95: Ah, okay. If you have a suggestion for how to improve the documentation of `IR.splitFuncType` (or any other function), don't hesitate to rephrase it : )

ichistmeinname (approved):
I really like the effort you put into the documentation, great work!

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -237,8 +317,15 @@ arityOf (IR.Lambda _ _ _ _     ) = return 0
...
+minimumM :: [Converter Int] -> Converter Int
+minimumM []         = return 0
+minimumM (mx : mxs) = foldr minM mx mxs
+
+-- Calculates the minimum of two integers lifted into the Converter monad.
+minM :: Converter Int -> Converter Int -> Converter Int
+minM m1 m2 = do
+  i1 <- m1
+  i2 <- m2
+  return (if i1 <= i2 then i1 else i2)
```
ichistmeinname: It's a pity that there's no `minimumM` or `minimumA` function in some Haskell library. Nevertheless, as your version of `minM` is basically a lifted version of `min`, we can rewrite it as one fold that yields `return 0`. I'd suggest to then outsource this function into a `Util` module that should exist somewhere already (ping @just95). For that sake, I generalised the function from `Converter Int` to `m a`, where  `a` is `Ord` (for the usage of `min`) and `Bounded` (in order to yield the smallest value otherwise).

```suggestion
minimumM :: (Ord a, Bounded a) => [m a] -> m a
minimumM = foldr (liftM2 min) (return minBound)
```

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -209,6 +276,19 @@ etaConvertSubExprs' expr = do
 -- Arity                                                                     --
 -------------------------------------------------------------------------------
 
+-- | Find the minimum number of missing arguments (the arity of the expressions) among the alternatives
+--   for the right-hand side of a function declaration. 
+--   All non-zero arities on the right-hand side of a function declaration 
+--   have the same arity.
+findMinMissingArguments :: IR.Expr -> Converter Int
+findMinMissingArguments (IR.If _ _ e1 e2 _) =
+  minM (findMinMissingArguments e1) (findMinMissingArguments e2)
```
ichistmeinname: We can get rid of calling `findMinMissingArguments` twice!

```suggestion
findMinMissingArguments :: IR.Expr -> Converter Int
findMinMissingArguments (IR.If _ _ e1 e2 _) =
  minimumM (map findMinMissingArguments [e1, e2])
```

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -106,57 +132,98 @@ etaConversionPass ast = do
...
+-- are fully applied.
+etaConvertTopLevel argPats expr = localEnv $ do
   let argExprs = map IR.varPatToExpr argPats
-  return $ IR.app NoSrcSpan expr' argExprs
+  -- apply expression to missing arguments and perform eta-conversion on the 
+  -- resulting expression 
+  etaConvertExpr $ IR.app NoSrcSpan expr argExprs
+
+-- | Calls @etaConvertTopLevel@ on all alternatives in an if or case expression. 
+etaConvertTopLevel' :: [IR.VarPat] -> IR.Expr -> Converter IR.Expr
```
ichistmeinname: Can we think of a better name here?

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -106,57 +132,98 @@ etaConversionPass ast = do
...
-  -- Compute the function's new (uncurried) type. Assumes that funcDecl's return type has already been inferred.
-  let (newArgTypes, returnType) = IR.splitFuncType
-        (fromJust $ IR.funcDeclReturnType funcDecl)
-        (length newArgIdents)
+  newArgNumber <- findMinMissingArguments rhs
+  -- Only perform top-level eta conversion when all alternatives for right-hand 
+  -- sides are missing at least one argument.
+  if newArgNumber == 0
+    then do
+      rhs' <- etaConvertExpr (IR.funcDeclRhs funcDecl)
```
ichistmeinname:
```suggestion
      rhs' <- etaConvertExpr rhs
```

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -106,57 +132,98 @@ etaConversionPass ast = do
...
 --   declaration and the function's type and the environment are updated accordingly. 
 etaConvertFuncDecl :: IR.FuncDecl -> Converter IR.FuncDecl
 etaConvertFuncDecl funcDecl = do
   let rhs = IR.funcDeclRhs funcDecl
-  newArgIdents <- generateFullArgumentList rhs
-  -- Compute the function's new (uncurried) type. Assumes that funcDecl's return type has already been inferred.
-  let (newArgTypes, returnType) = IR.splitFuncType
-        (fromJust $ IR.funcDeclReturnType funcDecl)
-        (length newArgIdents)
+  newArgNumber <- findMinMissingArguments rhs
```
ichistmeinname, line 172: I'm currently regretting reading the file bottom to top instead of the other way around : ) I now have the impression that `findMinMissingArguments` and `arityOf` do not need to be monadic (i.e., of type `Converter <something>`) -- there's not context needed to compute the needed value. Can you confirm or refute that impression?

ichistmeinname (changes requested):

src/lib/FreeC/Pass/EtaConversionPass.hs
MajaRet: `arityOf` needs to look up the arity of a function in the current environment, so it does need to produce a value inside the `Converter` monad. 
`inEnv` has the type `MonadConverter m => (Environment -> a) -> m a`. 
Since `findMinMissingArguments` uses `arityOf`, it also needs to produce a monadic value.

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -142,10 +272,24 @@ etaConvertSubExprs' expr = do
...
 
+-- | Find the minimum number of missing arguments (the arity of the expressions) among the alternatives
+--   for the right-hand side of a function declaration. 
+--   All non-zero arities on the right-hand side of a function declaration 
+--   have the same arity.
+findMinMissingArguments :: IR.Expr -> Converter Int
+findMinMissingArguments (IR.If _ _ e1 e2 _) =
+  minimumM (map findMinMissingArguments [e1, e2])
+findMinMissingArguments (IR.Case _ _ alts _) =
+  minimumM $ map (findMinMissingArguments . IR.altRhs) alts
```
just95: We can avoid to add `minimumM` and a monad utility module by writing the following.

```haskell
minimum <$> mapM (findMinMissingArguments . IR.altRhs) alts
```

src/lib/FreeC/Pass/EtaConversionPass.hs
ichistmeinname: Ah, I see. I missed that from the latest diff I was looking at!

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -51,20 +98,30 @@
...
 import           FreeC.Environment
 import           FreeC.Environment.Fresh
+import           FreeC.Environment.Entry
 import           FreeC.IR.SrcSpan
 import           FreeC.IR.Subterm
 import qualified FreeC.IR.Syntax               as IR
 import           FreeC.Monad.Converter
 import           FreeC.Pass
+-- temporary import; ideally, this pass should be moved before the TypeSignaturePass.
+import           FreeC.Pass.TypeSignaturePass   ( splitFuncType )
```
just95: Could you make this a `TODO` comment such that we don't forget that we have to address this in the future?

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -78,18 +135,91 @@ etaConversionPass ast = do
...
 etaConvertFuncDecl funcDecl = do
-  rhs' <- etaConvertExpr (IR.funcDeclRhs funcDecl)
-  return funcDecl { IR.funcDeclRhs = rhs' }
+  let rhs = IR.funcDeclRhs funcDecl
+  newArgNumber <- findMinMissingArguments rhs
+  -- Only perform top-level eta conversion when all alternatives for right-hand sides are missing at least one argument.
+  if newArgNumber == 0
+    then do
+      rhs' <- etaConvertExpr rhs
+      return funcDecl { IR.funcDeclRhs = rhs' }
```
just95: Shouldn't the `else` branch cover this case already?

just95 (changes requested):

src/lib/FreeC/Pass/EtaConversionPass.hs
MajaRet: Yes, you're right. I thought this way, we'd avoid unnecessary function evaluations and the 'updating' of the environment, but thinking about it again, it just causes different unnecessary function evaluations (like computing the arity of the right-hand side again).

src/lib/FreeC/Pass/EtaConversionPass.hs
MajaRet: Good point!

ichistmeinname: I had a version using `liftM` as well but refrained from this suggestion afterwards ; ) So, yes, please go ahead with this suggestion!

jo-hanna1997 (commented):
Your implementation looks good and I like the extension for if-expressions on top-level.
An interesting case occurs when in both if-branches functions with different arity occur.
For example:
```
--assume g as arity 2
g :: Foo -> Foo -> Foo 
--assume h as arity 1
h:: Foo -> (Foo -> Foo)

f :: Bool -> Foo -> Foo
f b = if b then g Foo else h Foo
```
Currently this will be transformed into:

`f b = if b then (\x -> g Foo x) else h Foo`

But we could also over apply `h` which is no harm to obtain

`f b x = if b the g Foo x else h Foo x` 

Furthermore I do think a recursive call is neccessary if we consider mutually recursive functions.
Let's say we have `g` as above and 
```
k :: Foo -> Foo
k = f

f :: Foo -> Foo
f = g (k Foo)
```
Then a first run would update `f` to `f x = g (k Foo) x`.
Now `f` has arity 1 but in `k` `f` is applied to no arguments. 
With repeated runs this should be resolved and yield
`k = (\x -> f x)`

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -78,18 +133,83 @@ etaConversionPass ast = do
...
+  -- apply expression to missing arguments and perform eta-conversion on the 
+  -- resulting expression 
+  etaConvertExpr $ IR.app NoSrcSpan expr argExprs
+
+-- | Calls @etaConvertTopLevel@ on all alternatives in an if or case expression. 
+etaConvertAlternatives :: [IR.VarPat] -> IR.Expr -> Converter IR.Expr
+etaConvertAlternatives argPats expr = do
+  -- The first child term of an if or case expression is the condition/the scrutinee
+  -- and should remain unchanged.
+  let (e : subterms) = childTerms expr
```
ichistmeinname, line 236: Are we sure that this binding can never fail, that is, `childTerms expr` never yields an empty list?

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -78,18 +133,83 @@ etaConversionPass ast = do
...
+  subterms' <- mapM (etaConvertTopLevel argPats) subterms
+  let Just expr' = replaceChildTerms expr (e : subterms')
```
ichistmeinname, line 238: Maybe we should discuss, if we want to just `fromJust` directly or use the let-binding like here. Any arguments on your or @just95's side?

ichistmeinname (changes requested):

src/lib/FreeC/Pass/EtaConversionPass.hs
MajaRet: It should never yield an empty list. `etaConvertAlternatives` is only used when `expr` is an if or case expression, and `childTerms` is defined as follows for those kinds of expressions (in `FreeC.IR.Subterm`):
```
  childTerms (IR.If _ e1 e2 e3 _        ) = [e1, e2, e3]
  childTerms (IR.Case   _ expr alts _   ) = expr : map IR.altRhs alts
```
So the list always has at least one element.

src/lib/FreeC/Pass/EtaConversionPass.hs
just95: In this case, I'd keep it as is for now. The issue will be addressed in #52 anyway.

ichistmeinname (approved):

src/lib/FreeC/Pass/EtaConversionPass.hs
just95: I agree that it is safe to assume that the list cannot be empty. The documentation of `childTerms` states that the list contains all direct child nodes. Since `if` and `case` expressions both have at least one direct child, the list is guaranteed  to be non-empty by the specification in this situation. However, `childTerms`  does not really provide any guarantee that the nodes are sorted from left to right. That's just how it happens to be implemented right now. Thus, I would recommend to perform the recursion in `etaConvertTopLevel` explicitly.

MajaRet: Okay, that makes sense. I'll change it then.

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -51,45 +97,133 @@
...
--- | Applies 'etaConvertExpr' to the right-hand side of the given function
---   declaration.
+-- | Make sure all occurring functions are fully applied by calling 'etaConvertFuncDecl' on each function declaration. 
+--
+--   If the type of a function declaration changes due to a top-level 
+--   η-conversion, the procedure is performed recursively on all 
+--   previously converted function declarations. 
+--   This ensures that all functions, including mutually-recursive 
+--   functions, are fully applied correctly. 
+etaConvertFuncDecls :: [IR.FuncDecl] -> [IR.FuncDecl] -> Converter [IR.FuncDecl]
```
just95, line 153: It would be nice if you could add an explanation of the function arguments to the Haddock documentation.

See [Function Arguments](https://haskell-haddock.readthedocs.io/en/latest/markup.html#function-arguments).

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -51,45 +97,133 @@
...
+  funcDecls' <- etaConvertFuncDecls (IR.modFuncDecls ast) []
   return ast { IR.modFuncDecls = funcDecls' }
 
 -------------------------------------------------------------------------------
 -- Function declarations                                                     --
 -------------------------------------------------------------------------------
 
--- | Applies 'etaConvertExpr' to the right-hand side of the given function
---   declaration.
+-- | Make sure all occurring functions are fully applied by calling 'etaConvertFuncDecl' on each function declaration.
```
just95: Documentation of function declarations should start with a third-person verb.

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -51,45 +97,133 @@
...
+--   functions, are fully applied correctly. 
+etaConvertFuncDecls :: [IR.FuncDecl] -> [IR.FuncDecl] -> Converter [IR.FuncDecl]
+etaConvertFuncDecls []         newFuncDecls = return newFuncDecls
+etaConvertFuncDecls (fd : fds) newFuncDecls = do
+  newFuncDecl <- etaConvertFuncDecl fd
+  if IR.funcDeclReturnType newFuncDecl /= IR.funcDeclReturnType fd
+    then etaConvertFuncDecls (newFuncDecls ++ (newFuncDecl : fds)) []
+    else etaConvertFuncDecls fds (newFuncDecls ++ [newFuncDecl])
+
+-- | Depending on the presence or absence of missing top-level arguments, applies 'modifyTopLevel' or 'etaConvertExpr' to the right-hand side of the given function declaration to ensure all functions
```
just95: This line looks way to long. Make sure no comments exceed the maximum line length of 80 characters.

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -51,45 +97,133 @@
...
+etaConvertAlternatives :: [IR.VarPat] -> IR.Expr -> Converter IR.Expr
+etaConvertAlternatives argPats expr = do
+  -- The first child term of an if or case expression is the condition/the scrutinee
+  -- and should remain unchanged.
+  let (e : subterms) = childTerms expr
+  subterms' <- mapM (etaConvertTopLevel argPats) subterms
+  let Just expr' = replaceChildTerms expr (e : subterms')
+  return expr'
+
+
```
just95: There should be at most one blank line between two code blocks. There are multiple redundant blank lines in this file and the tests that can be removed.

just95 (changes requested):
Looks good to me. There are just some formatting and documentation related issues that I would like you to address before we merge this.

jo-hanna1997:
It would be nice to know how you decided to proceed, converning the comment I wrote, so that I can adjust the tests.

MajaRet:
> It would be nice to know how you decided to proceed, converning the comment I wrote, so that I can adjust the tests.

Yes, of course, sorry. I added the recursion (by introducing a new function `etaConvertFuncDecls`). I also made the other change you suggested, but I did it in another branch called 'issue-3-overapplied' because I'm not completely sure that the version with over-applications is always equivalent in the presence of effects. 
All that changed there is that the two occurrences of `minimum` were changed to `maximum`.

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -22,26 +22,72 @@
...
+--   Both types of η-conversion may also be applied to the same expression.
+--   For example, the function 
+--   
+--  > f :: τ₁ -> τ₂ -> τ₃ > τ 
+--  > f e₁ = case e₁ of
+--  >           c₁ -> g 
+--  >           c₂ -> h 
 --
---   where @x₍ₘ₊₁₎ … xₙ@ are @n-m@ fresh variables.
+--  where c₁ and c₂ are the constructors of τ₁,
```
just95:
```suggestion
--  where @c₁@ and @c₂@ are the constructors of @τ₁@, 
```

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -51,45 +97,142 @@
...
 import           FreeC.Environment.Fresh
+import           FreeC.Environment.Entry
 import           FreeC.IR.SrcSpan
 import           FreeC.IR.Subterm
 import qualified FreeC.IR.Syntax               as IR
 import           FreeC.Monad.Converter
 import           FreeC.Pass
+-- temporary import; TODO: move this pass before the TypeSignaturePass.
+import           FreeC.Pass.TypeSignaturePass   ( splitFuncType )
+
```
just95, line 126:
```suggestion
```

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -22,26 +22,72 @@
...
+--   > f e₁ … eₖ = g @α₁ … @αₚ e₁ … eₘ
+--
+--   where @g@ is the name of an @n@-ary constructor or function declaration and 
+--   @m < n@. This declaration is then replaced by 
+--   > f e₁ … eₖ x₍ₘ₊₁₎ … xₙ = g @α₁ … @αₚ e₁ … eₘ x₍ₘ₊₁₎ … xₙ
+--   
+--   where x₍ₘ₊₁₎ … xₙ are @n-m@ fresh variables. 
+--
+--   If a function rule has several alternatives for the right-hand side through 
+--   -- possibly nested -- if or case expressions, the number of added arguments
```
just95: Are the `--` intentional (en- or em-dashes) or has something gone wrong during formatting? I'd suggest to use Unicode (i.e., `–` or `—`) if you want to insert a dash.

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -22,26 +22,72 @@
...
---   > \x₍ₘ₊₁₎ … xₙ -> f @α₁ … @αₚ e₁ … eₘ x₍ₘ₊₁₎ … xₙ
+--   where @y₍ₗ₊₁₎ … yₚ@ are @p-l@ fresh variables. 
+-- 
+--   Both types of η-conversion may also be applied to the same expression.
+--   For example, the function 
+--   
+--  > f :: τ₁ -> τ₂ -> τ₃ > τ 
+--  > f e₁ = case e₁ of
+--  >           c₁ -> g 
+--  >           c₂ -> h
```
just95: Maybe IR notation should be used here.

```suggestion
--  > f :: τ₁ -> τ₂ -> τ₃ > τ 
--  > f e₁ = case e₁ of {
--  >     c₁ -> g;
--  >     c₂ -> h 
--  >   }
```

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -146,6 +289,17 @@ etaConvertSubExprs' expr = do
 -- Arity                                                                     --
 -------------------------------------------------------------------------------
 
+-- | Find the minimum number of missing arguments among the alternatives
```
just95:
```suggestion
-- | Finds the minimum number of missing arguments among the alternatives
```

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -51,45 +97,142 @@
...
+    else etaConvertFuncDecls fds (newFuncDecls ++ [newFuncDecl])
+
+-- | Applies appropriate η-conversions to a function declaration.
+--
+-- Depending on the presence or absence of missing top-level arguments,
+-- the function uses 'etaConvertTopLevel' or 'etaConvertExpr' to ensure 
+-- all functions and constructors on the right-hand side are fully applied. 
+-- The missing top-level arguments are also added to the left-hand
+-- side of the declaration and the function's type and the environment
+-- are updated accordingly.
```
just95: This paragraph should be aligned with the subject line. There is another line in 207 that needs to be aligned.
After aligning the comments, make sure that the 80 character limit is still met. Some lines are still slightly too long right now.

```suggestion
--   Depending on the presence or absence of missing top-level arguments,
--   the function uses 'etaConvertTopLevel' or 'etaConvertExpr' to ensure 
--   all functions and constructors on the right-hand side are fully applied. 
--   The missing top-level arguments are also added to the left-hand
--   side of the declaration and the function's type and the environment
--   are updated accordingly. 
```

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -51,45 +97,142 @@
...
+  etaConvertAlternatives argPats expr
+etaConvertTopLevel argPats expr@(IR.Case _ _ _ _) =
+  etaConvertAlternatives argPats expr
+-- If there is only one alternative, apply it to the newly-added arguments, then 
+-- apply @etaConvertExpr@ to it to make so the expression and its sub-expressions 
+-- are fully applied.
+etaConvertTopLevel argPats expr = localEnv $ do
+  let argExprs = map IR.varPatToExpr argPats
+  -- apply expression to missing arguments and perform eta-conversion on the 
+  -- resulting expression
```
just95:
```suggestion
  -- Apply expression to missing arguments and perform eta-conversion on the 
  -- resulting expression.
```

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -22,26 +22,72 @@
...
 --   == Translation
 --
---   On the right-hand sides of function declarations, all of the largest
+--   Assume that we have the following function declaration. 
+--
+--   > f e₁ … eₖ = g @α₁ … @αₚ e₁ … eₘ
+--
+--   where @g@ is the name of an @n@-ary constructor or function declaration and 
+--   @m < n@. This declaration is then replaced by 
+--   > f e₁ … eₖ x₍ₘ₊₁₎ … xₙ = g @α₁ … @αₚ e₁ … eₘ x₍ₘ₊₁₎ … xₙ
```
just95: Some markup languages have problems with consecutive block-lvel elements. I don't know how Haddock handles this case but I'd insert a blank line between block-level elements just to be safe.

```suggestion
--
--   > f e₁ … eₖ x₍ₘ₊₁₎ … xₙ = g @α₁ … @αₚ e₁ … eₘ x₍ₘ₊₁₎ … xₙ
```

just95 (changes requested):

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -22,26 +22,75 @@
...
+--  > f :: τ₁ -> τ₂ -> τ₃ > τ 
+--  > f e₁ = case e₁ of {
+--  >           c₁ -> g; 
+--  >           c₂ -> h 
+--  >   }
+--
+--  where @c₁@ and @c₂@ are the constructors of @τ₁@, 
+--  @g@ is a binary function of type @τ₂ -> τ₃ > τ@, and
+--  @h@ is a unary function of type @τ₂ -> (τ₃ > τ)@, 
+--  will be converted to
```
just95: I just noticed that the dash is missing from some function arrows.

```suggestion
--  > f :: τ₁ -> τ₂ -> τ₃ -> τ
--  > f e₁ = case e₁ of {
--  >           c₁ -> g;
--  >           c₂ -> h 
--  >   }
--
--  where @c₁@ and @c₂@ are the constructors of @τ₁@,
--  @g@ is a binary function of type @τ₂ -> τ₃ -> τ@, and
--  @h@ is a unary function of type @τ₂ -> (τ₃ -> τ)@,
--  will be converted to
```

While writing the suggestion above, I've also noticed that there seem to be some trailing spaces. I'd recommend to configure your editor to remove trailing whitespace automatically.

src/lib/FreeC/Pass/EtaConversionPass.hs
```diff
@@ -22,26 +22,75 @@
...
+--  >   }
+--
+--  where @c₁@ and @c₂@ are the constructors of @τ₁@, 
+--  @g@ is a binary function of type @τ₂ -> τ₃ > τ@, and
+--  @h@ is a unary function of type @τ₂ -> (τ₃ > τ)@, 
+--  will be converted to 
+-- 
+--  > f e₁ x = case e₁ of
+--  >           c₁ -> \y -> g x y
+--  >           c₂ -> h x
```
just95: My [previous comment](https://github.com/FreeProving/free-compiler/pull/46#discussion_r440710789) about using IR notation in examples also applies to this code block. It may also be a good idea to write the type annotations in IR notation to demonstrate that `modifyTopLevel` manipulates the type annotations.

```haskell
f (e₁ :: τ₁) :: τ₂ -> τ₃ -> τ = case e₁ of {
    c₁ -> g;
    c₂ -> h 
  }

f (e₁ :: τ₁) (x :: τ₂) :: τ₃ -> τ = case e₁ of {
    c₁ -> \y -> g x y
    c₂ -> h x 
  }
```

just95 (changes requested):

jo-hanna1997 (approved):

just95 (approved):

PR #48, opened 2020-06-03
Add field for Agda identifiers to environment entries

### LINKED ISSUE #47: Add field for Agda identifiers to environment entries
The [environment entires](https://github.com/FreeProving/free-compiler/blob/master/src/lib/FreeC/Environment/Entry.hs) contain Coq identifiers. This allows to connect the Coq with the IR identifiers and ensure the validity of names. For the implementation of the Agda backend we will also need this functionality.

JonasHoefer:
I added `agdaIdent` fields to the environment entries. They are initialized together with the corresponding coq fields. I had to add preliminary names in the [`Prelude.toml`](https://github.com/FreeProving/free-compiler/blob/master/base/Prelude.toml) to implement the initialization in [`Decoder.hs`](https://github.com/FreeProving/free-compiler/blob/master/src/lib/FreeC/Environment/ModuleInterface/Decoder.hs).


just95 (changes requested):
Looks good to me, I'd just increase the version number of the module interface file format.

ichistmeinname (approved):

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithHelpers.hs
```diff
@@ -185,6 +185,7 @@ transformRecFuncDecl (IR.FuncDecl srcSpan declIdent typeArgs args maybeRetType e
       , entryIsPartial     = partial
       , entryName          = helperName
       , entryIdent         = undefined -- filled by renamer
+      , entryAgdaIdent     = undefined -- filled by renamer
```
ichistmeinname, line 188: There's not better way to model the missing information in that phase? Seems hacky to me.

JonasHoefer: This question would also interest me. I implemented the Agda identifiers only analogous to to the Coq ones and just accepted the `undefined`s.

ichistmeinname: Is it only necessary for `FuncEntry` or is a similar work-around used of the other entries as well @just95?

just95: The same hack is used for every entry since every entry has a `entryIdent` field that is filled by the `renameAndAddEntry` function. In theory it would be possible to ensure on type level, that the fields are filled correctly.

```haskell
data EnvEntry ident = FuncEntry { …, entryIdent :: ident } | …

-- The following corresponds to the current `EnvEntry` type.
type CoqEntry = EnvEntry Coq.Qualid

-- If we have such parameterized entries, we could define Agda identifiers 
-- like this but this would also mean that the `Environment`, `Converter` and
-- all `Pass`es need to be parameterized. That's why I've argued for a separate
-- `entryAgdaIdent` field in the past.
type AgdaEntry = EnvEntry Agda.QName 

-- The renamer can now take an entry that has not yet been renamed and return 
-- an entry that contains a Coq identifier.
type UnrenamedEntry = EnvEntry ()
filledByRenamer :: ()
renameAndAddEntry :: UnrenamedEntry -> Converter CoqEntry

-- This function would be used as follows. It does not need the `undefined` hack anymore.
renameAndAddEntry FuncDecl { …, entryIdent = filledByRenamer }
```

JonasHoefer: To avoid large scale refactoring for the moment I am also in favour of keeping the current version of the record (at least until we make some progress in the Agda backend and know what data is needed).

just95 (approved):

PR #51, opened 2020-06-04
Refine inlining recursively and adapt function ordering in Coq code

### LINKED ISSUE #7: Generated Coq code defines functions in uncompilable order and misses an argument unpacking
### Description

The generated code defines two functions (a called and a calling one) in the wrong order, such that the called one is not in scope. Simply interchanging the functions solves this.

Moreover one of the generated functions rightly defines a local function `f_0` but never calls it. Instead the function `f` (not in scope) is called and the unpacking scheme using `>>=` is ignored.

### Steps to Reproduce

Compile the following Haskell code:

```haskell
module ZigZag where

data Tree a = Leaf a | Node (Tree a) (Tree a)


{-# FreeC zigzag DECREASES ON ARGUMENT 1 #-}
zigzag :: Tree Bool -> Bool
zigzag t = zig t

zig:: Tree Bool -> Bool
zig (Leaf n) = n
zig (Node t1 t2) = zag t1

zag:: Tree Bool -> Bool
zag (Leaf n) = n
zag (Node t1 t2) = zigzag t2
```
with the following command

```bash
freec --transform-pattern-matching ZigZag.hs
```

#### Expected behavior

The example above should produce valid Coq code.

#### Actual behavior

In the generated helper function `zag_0`, the definition of `zigzag` is inlined but the definition of `zig` is not inlined recursively. The main function for `zig` is not yet defined at this point.

```coq
Fixpoint zag_0 (Shape : Type) (Pos : Shape -> Type) (a7
                 : Tree Shape Pos (Bool Shape Pos)) {struct a7} : Free Shape Pos (Bool Shape
                                                                                       Pos)
           := match a7 with
              | leaf a8 => a8
              | node a10 a11 => zig Shape Pos a11 (* <-- not inlined correctly *)
              end
with zig_0 (Shape : Type) (Pos : Shape -> Type) (a0
             : Tree Shape Pos (Bool Shape Pos)) {struct a0} : Free Shape Pos (Bool Shape
                                                                                   Pos)
       := match a0 with
          | leaf a1 => a1
          | node a3 a4 =>
              a3 >>= (fun (a3_0 : Tree Shape Pos (Bool Shape Pos)) => zag_0 Shape Pos a3_0)
          end.
```

Additionally, the generated main functions occur out of order.

```coq
Definition zag (Shape : Type) (Pos : Shape -> Type) (a7
    : Free Shape Pos (Tree Shape Pos (Bool Shape Pos)))
   : Free Shape Pos (Bool Shape Pos) :=
  a7 >>= (fun (a7_0 : Tree Shape Pos (Bool Shape Pos)) => zag_0 Shape Pos a7_0).

Definition zigzag (Shape : Type) (Pos : Shape -> Type) (t
    : Free Shape Pos (Tree Shape Pos (Bool Shape Pos)))
   : Free Shape Pos (Bool Shape Pos) :=
  zig Shape Pos t. (* <-- `zig` is used here but defined later *)

Definition zig (Shape : Type) (Pos : Shape -> Type) (a0
    : Free Shape Pos (Tree Shape Pos (Bool Shape Pos)))
   : Free Shape Pos (Bool Shape Pos) :=
  a0 >>= (fun (a0_0 : Tree Shape Pos (Bool Shape Pos)) => zig_0 Shape Pos a0_0).
```

### Versions

 - `freec --version`: 0.1.0.0
 - `ghc --version`: 8.6.5
 - `cabal --version`: 2.4.1.0
 - `coqc --version`: 8.9.0
 - OS: Windows

### Known workarounds

Replacing the last line by `zag (Node t1 t2) = zig t2` works just fine.

jo-hanna1997:
As suggested in #7 one run of function inlining applied to the helper function might not be sufficient. 
This is due to possibly long chains of functions beeing defined as other (partial) functions. The inlining process is now repeated as long as no fixpoint is reached. Hereby it is assumed to be sufficient that at each position all available functions should be inlined at most once (in order to guarentee termination).
Furthermore main functions can still be linear dependent on one another and are now sorted correspondingly.


src/lib/FreeC/IR/Inlining.hs
```diff
@@ -76,71 +86,81 @@ inlineExpr decls = inlineAndBind
...
     Just (typeArgs, args, rhs) -> do
       (typeArgs', rhs' ) <- renameTypeArgs typeArgs rhs
       (args'    , rhs'') <- renameArgs args rhs'
       return
-        (map IR.typeVarDeclIdent typeArgs', map IR.varPatIdent args', rhs'')
+        ( map IR.typeVarDeclIdent typeArgs'
+        , map IR.varPatIdent      args'
+        , rhs''
+        , name
+        )
```
just95: I'm wondering whether it is actually necessary to add an additional return value to `inlineExpr'`. As far as I can tell, you have to return `name` because `inlineAndBind` removes the function with that name from the list of functions to inline in the recursive call. But couldn't the recursive call be performed in here?

jo-hanna1997:
The current tests do not consider that main functions (also independet once) will be reordered.
However this happens because an analysis of the dependency graph is now performed on main functions as well. Especially functions independent from one another will now occure in decreasing lexicographic ordering. The current tests do not consider this and I will therefore adjust them.

src/lib/FreeC/IR/Inlining.hs
```diff
@@ -18,6 +18,7 @@ import           FreeC.Monad.Converter
 import           FreeC.Monad.Reporter
 import           FreeC.Pretty
 
+
```
just95: Avoid multiple consecutive blank lines.

src/test/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithHelpersTests.hs
```diff
@@ -1,5 +1,5 @@
--- | This module contains tests for
---   "FreeC.Backend.Coq.Converter.FuncDecl.Rec.WithHelpers".
+-- | This module contains tests for
+--   "FreeC.Backend.Coq.Converter.FuncDecl.Rec.WithHelpers".
```
just95: There seems to be still some kind of problem with Brittany. As far as I can see, there is no difference between the two lines (other than the line endings, presumably).

just95 (approved):

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/NonRec.hs
```diff
@@ -3,15 +3,28 @@
...
 import           FreeC.Backend.Coq.Converter.Expr
 import           FreeC.Backend.Coq.Converter.FuncDecl.Common
 import qualified FreeC.Backend.Coq.Syntax      as Coq
+import           FreeC.IR.DependencyGraph
 import qualified FreeC.IR.Syntax               as IR
 import           FreeC.Monad.Converter
 
+
+-- | Converts non-recursive but possibly linear dependent Haskell functions
+--   into an ordered list of @Definiton@ sentences such that each definition
```
stu204146-kiel: Typo
Definition instead of Definiton

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/NonRec.hs
```diff
@@ -3,15 +3,28 @@
...
+--   only depends on definitions at preceeding list positions.
+convertNonRecFuncDecls :: [IR.FuncDecl] -> Converter [Coq.Sentence]
+convertNonRecFuncDecls decls =
+  let orderdDecls = concatMap
```
stu204146-kiel: Typo
```suggestion
  let orderedDecls = concatMap
```

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithHelpers.hs
```diff
@@ -1,5 +1,5 @@
 -- | This module contains functions for converting mutually recursive
---   function declarations by spliting when into one or more recursive helper
+--   function declarations by spliting them into one or more recursive helper
```
stu204146-kiel:
```suggestion
--   function declarations by splitting them into one or more recursive helper
```

stu204146-kiel (changes requested):

stu204146-kiel (approved):

PR #55, opened 2020-06-09
Ignore `Arbitrary` instances #15

### LINKED ISSUE #15: Ignore `Arbitrary` instances
## Description

`freec` currently ignores type class contexts such that Haskell modules can still be compiled with GHC. It would be nice if we could ignore instance declarations, too.

## Motivation

When the user wants to test QuickCheck properties in GHC, `Arbitrary` instances have to be defined for custom data types. Currently, `freec` does not support type classes and type class instance declarations. Thus, the following error message is reported if there are `Arbitrary` instance declarations.

```
example/Test.hs:5:1-7:0: error:
    Type classes are not supported!
   |
 5 | instance Arbitrary Test where
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```

As a work around, `Arbitrary` instances currently have to be commented out when compiling the module to Coq.

It should be straight forward to use the same mechanism that we use to ignore type class contexts to ignore all instance declarations. Thus, I'm going to add the <kbd>good first issue</kbd> label.

## Alternatives

 - Instead of ignoring all instance declarations, we could also just ignore `Arbitrary` instances. However, `Show` and `Eq` instances are also often needed in QuickCheck. Thus, it might be better to ignore all instance declarations.
 - `freec` supports custom pragmas of the form `{-# FreeC ... #-}`. We could add a custom pragma to ignore individual instance declarations in the future.

MarvinLira:
As it was suggested in #15, all declarations of type class instances are ignored using the same mechanism that is used to ignore type class contexts.
As deriving clauses on data definitions are already ignored, I also ignored standalone deriving declarations for consistency reasons.
Closes #15

### DIFF
--- a/src/lib/FreeC/Frontend/Haskell/Simplifier.hs
+++ b/src/lib/FreeC/Frontend/Haskell/Simplifier.hs
@@ -328,8 +328,12 @@ simplifyDecl decl@(HSE.GDataDecl _ _ _ _ _ _ _) =
 simplifyDecl decl@(HSE.GDataInsDecl _ _ _ _ _ _) =
   notSupported "GADT style declarations" decl
 simplifyDecl decl@(HSE.ClassDecl _ _ _ _ _) = notSupported "Type classes" decl
-simplifyDecl decl@(HSE.InstDecl _ _ _ _) = notSupported "Type classes" decl
-simplifyDecl decl@(HSE.DerivDecl _ _ _ _) = notSupported "Type classes" decl
+simplifyDecl decl@(HSE.InstDecl _ _ _ _) = do
+  skipNotSupported "Instance declarations" decl
+  return ([], [], [])
+simplifyDecl decl@(HSE.DerivDecl _ _ _ _) = do
+  skipNotSupported "Deriving declarations" decl
+  return ([], [], [])
 simplifyDecl decl@(HSE.DefaultDecl _ _) = notSupported "Type classes" decl
 simplifyDecl decl@(HSE.SpliceDecl _ _) = notSupported "Template Haskell" decl
 simplifyDecl decl@(HSE.TSpliceDecl _ _) = notSupported "Template Haskell" decl


just95 (approved):

src/lib/FreeC/Frontend/Haskell/Simplifier.hs
```diff
@@ -328,8 +328,12 @@ simplifyDecl decl@(HSE.GDataDecl _ _ _ _ _ _ _) =
 simplifyDecl decl@(HSE.GDataInsDecl _ _ _ _ _ _) =
   notSupported "GADT style declarations" decl
 simplifyDecl decl@(HSE.ClassDecl _ _ _ _ _) = notSupported "Type classes" decl
-simplifyDecl decl@(HSE.InstDecl _ _ _ _) = notSupported "Type classes" decl
-simplifyDecl decl@(HSE.DerivDecl _ _ _ _) = notSupported "Type classes" decl
+simplifyDecl decl@(HSE.InstDecl _ _ _ _) = do
+  skipNotSupported "Instance declarations" decl
+  return ([], [], [])
```
ichistmeinname, line 333: Would it make sense to move `return ([],[],[])` into the definition of `skipNotSupported`?

MarvinLira: I don't think so. Not in every position, where `skipNotSupported` is used, is `([],[],[])` the result to return. One could of course modify `skipNotSupported` such that it takes `([],[],[])` as a third argument and returns it, but this would not make the code much shorter.
Additionally, I would say that the purpose of this function should be to only produce a warning and not to return some value as it is also used within larger `do` blocks like in line 144.

ichistmeinname: Since the return type of `skipNotSupported` currently is `()`, all other usages of the function in a do-block just skip the result, which is still possible (and plausible) with the changed return value. Isn't it?

just95: @ichistmeinname GHC would complain in such cases and the result would have to be ignored with `_ <-` explicitly. Also, a return type of `([], [], [])` would lead to type errors when `skipNonSupported` is used in `when` or `unless`. I'd agree with @marvin2706 that the purpose of the function is just to produce a warning. If we really want to avoid the repeated `return ([], [], [])`s, I'd suggest to add a `skipNonSupportedDecl` function for that purpose.

ichistmeinname: I see, you're right. So maybe the code felt a bit less clumsy to me if all skipped constructors would be matched with a guard. Something like the following, where `describeSkip` yields the "description String" like `"Instance declarations" above.

```
simplifyDecl decl | isSkippable decl = skipNotSupported (describeSkip decl) decl
```

ichistmeinname: Then let's not overthink it and leave as it is.

stu204767 (approved):

PR #56, opened 2020-06-09
Implement translation of type expressions

### LINKED ISSUE #54: Implement translation of type expressions
Implement translation of type expressions (without `Size` annotations) using the `Concrete` representation.

JonasHoefer:
Adds translation for non recursive function signatures. Creates a base for converting further module entries. Blocked by #48. Closes #54

### DIFF
--- a/src/exe/Main.hs
+++ b/src/exe/Main.hs
@@ -148,7 +148,6 @@ convertInputModule haskellAst = do
     agdaAst <- liftConverter $ Agda.convertModule haskellAst
     traceM $ showPretty agdaAst
     traceM "\n--------\n"
-    traceM undefined
     return (modName, coqAst)
 
 -------------------------------------------------------------------------------
--- a/src/exe/Main.hs
+++ b/src/exe/Main.hs
@@ -23,8 +23,6 @@ import           FreeC.Application.Option.Help
 import           FreeC.Application.Option.Version
 import           FreeC.Application.Options
 import           FreeC.Application.Options.Parser
-import qualified FreeC.Backend.Agda.Converter  as Agda
-import           FreeC.Backend.Agda.Pretty      ( )
 import qualified FreeC.Backend.Coq.Base        as Coq.Base
 import           FreeC.Backend.Coq.Converter    ( convertModule )
 import           FreeC.Backend.Coq.Pretty
@@ -51,8 +49,6 @@ import           FreeC.Pretty                   ( putPrettyLn
                                                 , showPretty
                                                 , writePrettyFile
                                                 )
-
-import           Debug.Trace                    ( traceM )
 
 -------------------------------------------------------------------------------
 -- Main                                                                      --
@@ -145,9 +141,6 @@ convertInputModule haskellAst = do
   reportApp $ do
     loadRequiredModules haskellAst
     coqAst  <- liftConverter $ convertModule haskellAst
-    agdaAst <- liftConverter $ Agda.convertModule haskellAst
-    traceM $ showPretty agdaAst
-    traceM "\n--------\n"
     return (modName, coqAst)
 
 -------------------------------------------------------------------------------
--- a/src/exe/Main.hs
+++ b/src/exe/Main.hs
@@ -140,7 +140,7 @@ convertInputModule haskellAst = do
     else putDebug $ "Compiling " ++ showPretty modName
   reportApp $ do
     loadRequiredModules haskellAst
-    coqAst  <- liftConverter $ convertModule haskellAst
+    coqAst <- liftConverter $ convertModule haskellAst
     return (modName, coqAst)
 
 -------------------------------------------------------------------------------
--- a/src/exe/Main.hs
+++ b/src/exe/Main.hs
@@ -23,6 +23,8 @@ import           FreeC.Application.Option.Help
 import           FreeC.Application.Option.Version
 import           FreeC.Application.Options
 import           FreeC.Application.Options.Parser
+import qualified FreeC.Backend.Agda.Converter  as Agda
+import           FreeC.Backend.Agda.Pretty      ( )
 import qualified FreeC.Backend.Coq.Base        as Coq.Base
 import           FreeC.Backend.Coq.Converter    ( convertModule )
 import           FreeC.Backend.Coq.Pretty
@@ -49,6 +51,8 @@ import           FreeC.Pretty                   ( putPrettyLn
                                                 , showPretty
                                                 , writePrettyFile
                                                 )
+
+import           Debug.Trace                    ( traceM )
 
 -------------------------------------------------------------------------------
 -- Main                                                                      --
@@ -140,7 +144,11 @@ convertInputModule haskellAst = do
     else putDebug $ "Compiling " ++ showPretty modName
   reportApp $ do
     loadRequiredModules haskellAst
-    coqAst <- liftConverter $ convertModule haskellAst
+    coqAst  <- liftConverter $ convertModule haskellAst
+    agdaAst <- liftConverter $ Agda.convertModule haskellAst
+    traceM $ showPretty agdaAst
+    traceM "\n--------\n"
+    traceM undefined
     return (modName, coqAst)
 
 -------------------------------------------------------------------------------
--- a/src/lib/FreeC/Backend/Agda/Base.hs
+++ b/src/lib/FreeC/Backend/Agda/Base.hs
@@ -89,7 +89,6 @@ shape = Agda.name "S"
 position :: Agda.Name
 position = Agda.name "P"
 
--- | Adds the @Shape@ and @Position@ type variables to a list of variables.
 addTVars :: [Agda.Name] -> [Agda.Name]
 addTVars ts = shape : position : ts
 
--- a/src/lib/FreeC/Backend/Agda/Base.hs
+++ b/src/lib/FreeC/Backend/Agda/Base.hs
@@ -9,6 +9,7 @@ module FreeC.Backend.Agda.Base
     -- * Free Monad
   , free
   , free'
+  , freeArgs
   , pure
   , impure
   , shape
@@ -45,36 +46,50 @@ imports = Agda.simpleImport $ Agda.qname [baseLibName] $ Agda.name "Free"
 -- Free Monad                                                                --
 -------------------------------------------------------------------------------
 
+-- | Identifier for the @Free@ monad.
 free :: Agda.Name
 free = Agda.name "Free"
 
+-- | Lifts a type in the free monad.
 free' :: Agda.Expr -> Agda.Expr
-free' e = foldl1
+free' = Agda.app $ freeArgs $ Agda.qname' free
+
+-- | Apply the @Shape@ and @Position@ argument to the given type constructor.
+--
+-- > c ↦ c @S @P
+freeArgs :: Agda.QName -> Agda.Expr
+freeArgs qname = foldl1
   Agda.app
-  [ Agda.Ident (Agda.qname' free)
+  [ Agda.Ident qname
   , Agda.Ident (Agda.qname' shape)
   , Agda.Ident (Agda.qname' position)
-  , e
   ]
 
+-- | Identifier for the @Pure@ constructor of the @Free@ monad.
 pureConName :: Agda.Name
 pureConName = Agda.name "pure"
 
+-- | Applies the @Pure@ constructor of the free monad to the given expression.
 pure :: Agda.Expr -> Agda.Expr
 pure = Agda.app $ Agda.Ident $ Agda.qname [Agda.name "Free"] pureConName
 
+-- | Identifier for the @Impure@ constructor of the @Free@ monad.
 impureConName :: Agda.Name
 impureConName = Agda.name "impure"
 
+-- | Applies the @Impure@ constructor of the free monad to the given expression.
 impure :: Agda.Expr -> Agda.Expr
 impure = Agda.app $ Agda.Ident $ Agda.qname [Agda.name "Free"] impureConName
 
+-- | Reserved name for the @Shape@ type variable.
 shape :: Agda.Name
 shape = Agda.name "S"
 
+-- | Reserved name for the @Position@ type variable.
 position :: Agda.Name
 position = Agda.name "P"
 
+-- | Adds the @Shape@ and @Position@ type variables to a list of variables.
 addTVars :: [Agda.Name] -> [Agda.Name]
 addTVars ts = shape : position : ts
 
--- a/src/lib/FreeC/Backend/Agda/Base.hs
+++ b/src/lib/FreeC/Backend/Agda/Base.hs
@@ -8,13 +8,9 @@ module FreeC.Backend.Agda.Base
   , imports
     -- * Free Monad
   , free
-  , free'
-  , freeArgs
   , pure
-  , impure
   , shape
   , position
-  , addTVars
     -- * reserved identifiers
   , reservedIdents
   )
@@ -50,47 +46,21 @@ imports = Agda.simpleImport $ Agda.qname [baseLibName] $ Agda.name "Free"
 free :: Agda.Name
 free = Agda.name "Free"
 
--- | Lifts a type in the free monad.
-free' :: Agda.Expr -> Agda.Expr
-free' = Agda.app $ freeArgs $ Agda.qname' free
+-- | Identifier for the @pure@ constructor of the @Free@ monad.
+pure :: Agda.Name
+pure = Agda.name "pure"
 
--- | Apply the @Shape@ and @Position@ argument to the given type constructor.
---
--- > c ↦ c @S @P
-freeArgs :: Agda.QName -> Agda.Expr
-freeArgs qname = foldl1
-  Agda.app
-  [ Agda.Ident qname
-  , Agda.Ident (Agda.qname' shape)
-  , Agda.Ident (Agda.qname' position)
-  ]
-
--- | Identifier for the @Pure@ constructor of the @Free@ monad.
-pureConName :: Agda.Name
-pureConName = Agda.name "pure"
-
--- | Applies the @Pure@ constructor of the free monad to the given expression.
-pure :: Agda.Expr -> Agda.Expr
-pure = Agda.app $ Agda.Ident $ Agda.qname [Agda.name "Free"] pureConName
-
--- | Identifier for the @Impure@ constructor of the @Free@ monad.
-impureConName :: Agda.Name
-impureConName = Agda.name "impure"
-
--- | Applies the @Impure@ constructor of the free monad to the given expression.
-impure :: Agda.Expr -> Agda.Expr
-impure = Agda.app $ Agda.Ident $ Agda.qname [Agda.name "Free"] impureConName
+-- | Identifier for the @impure@ constructor of the @Free@ monad.
+impure :: Agda.Name
+impure = Agda.name "impure"
 
 -- | Reserved name for the @Shape@ type variable.
 shape :: Agda.Name
-shape = Agda.name "S"
+shape = Agda.name "Shape"
 
--- | Reserved name for the @Position@ type variable.
+-- | Reserved name for the @Pos@ type variable.
 position :: Agda.Name
-position = Agda.name "P"
-
-addTVars :: [Agda.Name] -> [Agda.Name]
-addTVars ts = shape : position : ts
+position = Agda.name "Pos"
 
 -------------------------------------------------------------------------------
 -- Reserved identifiers                                                      --
@@ -100,4 +70,4 @@ addTVars ts = shape : position : ts
 --
 --   This does only include identifiers without corresponding Haskell name.
 reservedIdents :: [Agda.Name]
-reservedIdents = [free, pureConName, impureConName, shape, position]
+reservedIdents = [free, pure, impure, shape, position]
--- a/src/lib/FreeC/Backend/Agda/Base.hs
+++ b/src/lib/FreeC/Backend/Agda/Base.hs
@@ -8,10 +8,12 @@ module FreeC.Backend.Agda.Base
   , imports
     -- * Free Monad
   , free
+  , free'
   , pure
   , impure
   , shape
   , position
+  , addTVars
     -- * reserved identifiers
   , reservedIdents
   )
@@ -46,6 +48,15 @@ imports = Agda.simpleImport $ Agda.qname [baseLibName] $ Agda.name "Free"
 free :: Agda.Name
 free = Agda.name "Free"
 
+free' :: Agda.Expr -> Agda.Expr
+free' e = foldl1
+  Agda.app
+  [ Agda.Ident (Agda.qname' free)
+  , Agda.Ident (Agda.qname' shape)
+  , Agda.Ident (Agda.qname' position)
+  , e
+  ]
+
 pureConName :: Agda.Name
 pureConName = Agda.name "pure"
 
@@ -63,6 +74,9 @@ shape = Agda.name "S"
 
 position :: Agda.Name
 position = Agda.name "P"
+
+addTVars :: [Agda.Name] -> [Agda.Name]
+addTVars ts = shape : position : ts
 
 -------------------------------------------------------------------------------
 -- Reserved identifiers                                                      --
--- a/src/lib/FreeC/Backend/Agda/Converter/Free.hs
+++ b/src/lib/FreeC/Backend/Agda/Converter/Free.hs
@@ -1,0 +1,38 @@
+-- | This module contains auxilary functions that help to generate Coq code
+--   that uses the @Free@ monad.
+
+module FreeC.Backend.Agda.Converter.Free
+  ( generatePure
+  , free
+  , applyFreeArgs
+  , addFreeArgs
+  )
+where
+
+import qualified FreeC.Backend.Agda.Syntax     as Agda
+import qualified FreeC.Backend.Agda.Base       as Agda.Base
+
+-- | Applies the @pure@ constructor of the free monad to the given expression.
+generatePure :: Agda.Expr -> Agda.Expr
+generatePure =
+  Agda.app $ Agda.Ident $ Agda.qname [Agda.Base.baseLibName] Agda.Base.pure
+
+-- | Lifts a type in the free monad.
+free :: Agda.Expr -> Agda.Expr
+free = Agda.app $ applyFreeArgs $ Agda.qname' Agda.Base.free
+
+-- | Apply the @Shape@ and @Pos@ argument to the given type constructor.
+--
+-- > c ↦ c @S @P
+applyFreeArgs :: Agda.QName -> Agda.Expr
+applyFreeArgs qname = foldl1
+  Agda.app
+  [ Agda.Ident qname
+  , Agda.Ident (Agda.qname' Agda.Base.shape)
+  , Agda.Ident (Agda.qname' Agda.Base.position)
+  ]
+
+-- | Adds the resvered names for the free args @Shape@ and @Pos@ to a list of
+--   names.
+addFreeArgs :: [Agda.Name] -> [Agda.Name]
+addFreeArgs ts = Agda.Base.shape : Agda.Base.position : ts
--- a/src/lib/FreeC/Backend/Agda/Converter/Free.hs
+++ b/src/lib/FreeC/Backend/Agda/Converter/Free.hs
@@ -23,7 +23,7 @@ free = Agda.app $ applyFreeArgs $ Agda.qname' Agda.Base.free
 
 -- | Apply the @Shape@ and @Pos@ argument to the given type constructor.
 --
--- > c ↦ c @S @P
+-- > c ↦ c Shape Pos
 applyFreeArgs :: Agda.QName -> Agda.Expr
 applyFreeArgs qname = foldl1
   Agda.app
--- a/src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
+++ b/src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
@@ -1,4 +1,7 @@
-module FreeC.Backend.Agda.Converter.FuncDecl where
+module FreeC.Backend.Agda.Converter.FuncDecl
+  ( convertFuncDecl
+  )
+where
 
 import           Prelude                 hiding ( mod )
 
@@ -13,21 +16,25 @@ import           FreeC.Monad.Converter          ( Converter
                                                 , inEnv
                                                 )
 import           FreeC.Environment              ( lookupAgdaIdent )
-import           Debug.Trace
+import           FreeC.Environment.LookupOrFail ( lookupAgdaIdentOrFail )
 
+
+-- | Converts the given function declarations. Returns the declarations for the
+--   type signature and the definition (TODO).
 convertFuncDecl :: IR.FuncDecl -> Converter [Agda.Declaration]
 convertFuncDecl decl = sequence [convertSignature decl]
 
+-- | Converts the type signature of the given function to an Agda type
+--   declaration.
 convertSignature :: IR.FuncDecl -> Converter Agda.Declaration
 convertSignature (IR.FuncDecl _ ident tVars args retType _) =
-  Agda.funcSig <$> convertIdent ident <*> liftType tVars types
+  Agda.funcSig <$> lookupIdent ident <*> convertFunc_ tVars types
   where types = (IR.varPatType <$> args) `snoc` retType
 
-convertDef :: IR.FuncDecl -> Converter Agda.Declaration
-convertDef = undefined
-
-convertIdent :: IR.DeclIdent -> Converter Agda.Name
-convertIdent ident = do
+-- | Looks up the name of a Haskell function in the environment and converts it
+--   to an Agda name.
+lookupIdent :: IR.DeclIdent -> Converter Agda.Name
+lookupIdent ident = do
   let name = IR.declIdentName ident
   Just qualid <- inEnv $ lookupAgdaIdent IR.ValueScope name
   pure $ Agda.unqualify qualid
@@ -36,8 +43,9 @@ convertIdent ident = do
 -- Free Type Lifting                                                         --
 -------------------------------------------------------------------------------
 
-liftType :: [IR.TypeVarDecl] -> [Maybe IR.Type] -> Converter Agda.Expr
-liftType tVars ts = Agda.pi tVars' <$> convertFunc (fromJust <$> ts)
+-- | Converts a fully applied function.
+convertFunc_ :: [IR.TypeVarDecl] -> [Maybe IR.Type] -> Converter Agda.Expr
+convertFunc_ tVars ts = Agda.pi tVars' <$> convertFunc (fromJust <$> ts)
   where tVars' = Agda.Base.addTVars (Agda.name . IR.typeVarDeclIdent <$> tVars)
 
 -- | Lifts an n-ary function piece wise. Functions, which take all their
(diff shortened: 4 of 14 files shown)


src/lib/FreeC/Backend/Agda/Syntax.hs
```diff
@@ -70,8 +98,47 @@ intLiteral = Lit . LitNat NoRange
...
+-- | Creates an application AST node. Application is left associative and in
+--   in type expressions binds stronger than type arrow. For these cases paren-
+--   thesis are added automatically.
 --
---   @e a@
+--   > e a
 app :: Expr -> Expr -> Expr
-app l r = App NoRange l $ defaultNamedArg r
+app l r@(App _ _ _) = App NoRange l $ defaultNamedArg $ paren r
+app l r@(Fun _ _ _) = App NoRange l $ defaultNamedArg $ paren r
```
ichistmeinname: Predicate in guard condition vs. code duplication. Not sure what I find more readable though. Any comments or preferences about that after calling it out?

src/lib/FreeC/Environment.hs
```diff
@@ -188,6 +194,16 @@ usedIdents = concatMap entryIdents . Map.elems . envEntries
...
 
+-- | Gets a list of Agda identifiers for functions, (type/smart) constructors,
+--   (type/fresh) variables that were used in the given environment already.
+usedAgdaIdents :: Environment -> [Agda.QName]
+usedAgdaIdents = concatMap entryIdents . Map.elems . envEntries
+ where
+  entryIdents :: EnvEntry -> [Agda.QName]
+  entryIdents entry
+    | isConEntry entry = [entryAgdaIdent entry, entryAgdaSmartIdent entry]
+    | otherwise        = [entryAgdaIdent entry]
```
ichistmeinname: Again, not sure it adds readability, but perhaps it's more explicit about the parts all cases have in common.

```suggestion
  entryIdents entry = (if isConEntry entry then (entryAgdaSmartIdent entry :) else id) [entryAgdaIdent entry]
```

src/lib/FreeC/Environment/LookupOrFail.hs
```diff
@@ -44,6 +45,20 @@ lookupIdentOrFail srcSpan scope name = do
...
+--
+--   If an error is reported, it points to the given source span.
+lookupAgdaIdentOrFail
+  :: SrcSpan  -- ^ The source location where the identifier is requested.
+  -> IR.Scope    -- ^ The scope to look the identifier up in.
+  -> IR.QName -- ^ The Haskell identifier to look up.
+  -> Converter Agda.QName
+lookupAgdaIdentOrFail srcSpan scope name = do
+  entry <- lookupEntryOrFail srcSpan scope name
+  return (entryAgdaIdent entry)
```
ichistmeinname:
```suggestion
lookupAgdaIdentOrFail srcSpan scope name =
  entryAgdaIdent <$> lookupEntryOrFail srcSpan scope name
```

src/lib/FreeC/Environment/LookupOrFail.hs
```diff
@@ -44,6 +45,20 @@ lookupIdentOrFail srcSpan scope name = do
...
 
+-- | Looks up the Agda identifier for a Haskell function, (type)
+--   constructor or (type) variable with the given name or reports a fatal
+--   error message if the identifier has not been defined.
+--
+--   If an error is reported, it points to the given source span.
+lookupAgdaIdentOrFail
+  :: SrcSpan  -- ^ The source location where the identifier is requested.
+  -> IR.Scope    -- ^ The scope to look the identifier up in.
+  -> IR.QName -- ^ The Haskell identifier to look up.
```
ichistmeinname: Why doesn't brittany want to indent the comment? : /

```suggestion
  :: SrcSpan  -- ^ The source location where the identifier is requested.
  -> IR.Scope -- ^ The scope to look the identifier up in.
  -> IR.QName -- ^ The Haskell identifier to look up.
```

just95: Brittany doesn't format comments. They have to be aliged manually.

ichistmeinname: What a pity (for @JonasHoefer  ; ))!

src/exe/Main.hs
```diff
@@ -140,7 +144,10 @@ convertInputModule haskellAst = do
     else putDebug $ "Compiling " ++ showPretty modName
   reportApp $ do
     loadRequiredModules haskellAst
-    coqAst <- liftConverter $ convertModule haskellAst
+    coqAst  <- liftConverter $ convertModule haskellAst
+    agdaAst <- liftConverter $ Agda.convertModule haskellAst
+    traceM $ showPretty agdaAst
+    traceM "\n--------\n"
```
ichistmeinname: The tracing expressions should be removed from the final code.

src/lib/FreeC/Backend/Agda/Base.hs
```diff
@@ -43,27 +46,52 @@ imports = Agda.simpleImport $ Agda.qname [baseLibName] $ Agda.name "Free"
...
 
+-- | Identifier for the @Impure@ constructor of the @Free@ monad.
 impureConName :: Agda.Name
 impureConName = Agda.name "impure"
 
+-- | Applies the @Impure@ constructor of the free monad to the given expression.
 impure :: Agda.Expr -> Agda.Expr
 impure = Agda.app $ Agda.Ident $ Agda.qname [Agda.name "Free"] impureConName
 
+-- | Reserved name for the @Shape@ type variable.
```
ichistmeinname, line 57: Somehow I have the feeling that I wanted to rename `shape` and `position` in an earlier version of this file. Did we agree not to do it or was it just forgotten? I wanted to use the same names as in Coq.

src/lib/FreeC/Backend/Agda/Syntax.hs
```diff
@@ -29,17 +39,21 @@ import           Agda.Syntax.Position
...
 -- | Creates a (not qualified) Agda variable name from a 'String'.
 name :: String -> Name
-name ident = Name NoRange InScope [Id ident]
+name str = Name NoRange InScope [Id str]
 
 -- | Create a qualified identifier given a local identifier as 'Name' and a
 --   list of module 'Name's.
 qname :: [Name] -> Name -> QName
-qname modules ident = foldr Qual (QName ident) modules
+qname modules n = foldr Qual (QName n) modules
```
ichistmeinname: I have the feeling that we usually use `n` for `Interger` values. Perhaps we can use `name` as variable name, if we rename the function `name` to `varName` or something similar?

src/lib/FreeC/Backend/Agda/Syntax.hs
```diff
@@ -29,17 +39,21 @@ import           Agda.Syntax.Position
...
 
 -- | Creates a qualified name using an empty list of module names.
 qname' :: Name -> QName
 qname' = qname []
 
+nextQName :: QName -> QName
+nextQName (Qual n qn) = Qual (nextName n) qn
```
ichistmeinname: Not sure what happens here: in case of `Qual` the `ModName` is changed but not `qn`? Again, maybe we can rename `n` here to s.th. with more semantic meaning like `modName`.

ichistmeinname (changes requested):

src/lib/FreeC/Backend/Agda/Syntax.hs
JonasHoefer: Its just wrong. The correct definition is of course.
```haskell
nextQName :: QName -> QName
nextQName (Qual n qn) = Qual n $ nextQName qn
nextQName (QName n  ) = QName $ nextName n
```

ichistmeinname: That makes more sense indeed. That is, `ModName` is just a type synonym for `Name, which got us into trouble here ; )

Nevertheless, I'd like to see a better variable name than `n`. In the best case both occurrences of `n` have _different_ names as they stand for different entities.

src/lib/FreeC/Backend/Agda/Syntax.hs
JonasHoefer: `name` would be a good name, but it produces a shadow warning, because `name :: String -> Agda.Name` is defined in the same module. `varName` is problematic, because `Agda.QName`s are used for modules and functions as well. I will try to find a better name.
EDIT: typo

src/lib/FreeC/Backend/Agda/Base.hs
JonasHoefer: I think i created this branch before changing the names on the other branch. The variables are named `Shape` and `Position` on the [`dev-agda`](https://github.com/FreeProving/free-compiler/blob/dev-agda/src/lib/FreeC/Backend/Agda/Base.hs) branch. Since the other commit is newer the names from the `dev-agda` branch should persist (even though i am still in favour of `S` and `P`, because they shorten type signatures immensely).

src/lib/FreeC/Environment.hs
JonasHoefer: In this case we could use `guard` to avoid the `else id`.
```haskell
  entryIdents entry =
    entryAgdaIdent entry 
      : (guard (isConEntry entry) $> entryAgdaSmartIdent entry)
```

src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
```diff
@@ -0,0 +1,76 @@
...
+import qualified FreeC.Backend.Agda.Syntax     as Agda
+import           FreeC.Environment              ( lookupAgdaIdent )
+import           FreeC.Environment.LookupOrFail ( lookupAgdaIdentOrFail )
+import           FreeC.Environment.Renamer      ( renameAndDefineAgdaTypeVar )
+import qualified FreeC.IR.Syntax               as IR
+import           FreeC.Monad.Converter          ( Converter
+                                                , inEnv
+                                                , localEnv
+                                                )
+import           Prelude                 hiding ( mod )
```
just95: Remember to sort your imports.

src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
```diff
@@ -0,0 +1,76 @@
...
+--   > (τ → σ)’ = Free (Free τ’ → Free σ’)
+--   > α' = α    (τ σ)’ = τ’ σ’    C' = Ĉ @S @P
+convertType :: IR.Type -> Converter Agda.Expr
+convertType (IR.TypeVar s name) = Agda.Ident
+  <$> lookupAgdaIdentOrFail s IR.TypeScope (IR.UnQual (IR.Ident name))
+convertType (IR.TypeCon s name) =
+  Agda.Base.freeArgs <$> lookupAgdaIdentOrFail s IR.TypeScope name
+convertType (IR.TypeApp  _ l r) = Agda.app <$> convertType l <*> convertType r
+convertType (IR.FuncType _ l r) = freeFunc <$> convertType l <*> convertType r
+  where freeFunc = Agda.func `on` Agda.Base.free'
```
just95: Since the translation of types is needed for the translation of most other AST nodes, this function should be moved to its own module.

The comment should either explain what the "dagger translation" is or at least refer to an explanation. Furthermore, this does not look like the dagger translation to me. The implementation resembles the `*` translation. The translation rules in the comment don't look right to me. They resemble neither the `*` nor the dagger translation.

src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
```diff
@@ -0,0 +1,76 @@
...
+  Agda.funcSig <$> lookupIdent ident <*> convertFunc_ tVars types
+  where types = (IR.varPatType <$> args) `snoc` retType
+
+-- | Looks up the name of a Haskell function in the environment and converts it
+--   to an Agda name.
+lookupIdent :: IR.DeclIdent -> Converter Agda.Name
+lookupIdent ident = do
+  let name = IR.declIdentName ident
+  Just qualid <- inEnv $ lookupAgdaIdent IR.ValueScope name
+  pure $ Agda.unqualify qualid
```
just95: Is there a reason why you are using `pure` instead of `return` here?

src/lib/FreeC/Backend/Agda/Base.hs
```diff
@@ -43,27 +46,52 @@ imports = Agda.simpleImport $ Agda.qname [baseLibName] $ Agda.name "Free"
...
 pure :: Agda.Expr -> Agda.Expr
 pure = Agda.app $ Agda.Ident $ Agda.qname [Agda.name "Free"] pureConName
 
+-- | Identifier for the @Impure@ constructor of the @Free@ monad.
 impureConName :: Agda.Name
 impureConName = Agda.name "impure"
 
+-- | Applies the @Impure@ constructor of the free monad to the given expression.
 impure :: Agda.Expr -> Agda.Expr
 impure = Agda.app $ Agda.Ident $ Agda.qname [Agda.name "Free"] impureConName
```
just95: Try to use the same name in the comments as in the implementation (e.g., `@pure@` instead of `@Pure@`).

In the Coq back end, there is never the need to generate an `impure` constructor. I guess that this will also not be necessary in case of the Agda back end. If you still want to keep the `impure` smart constructor, it should probably have two arguments (one for the concrete shape and one for the position function).

The purpose of this module is to make it easier to change the names of the definitions in the base library. Instead of writing `Agda.name "Free"` you should reuse your definition of `free`.

src/lib/FreeC/Backend/Agda/Base.hs
```diff
@@ -43,27 +46,52 @@ imports = Agda.simpleImport $ Agda.qname [baseLibName] $ Agda.name "Free"
...
+-- | Reserved name for the @Shape@ type variable.
 shape :: Agda.Name
 shape = Agda.name "S"
 
+-- | Reserved name for the @Position@ type variable.
 position :: Agda.Name
 position = Agda.name "P"
 
+addTVars :: [Agda.Name] -> [Agda.Name]
+addTVars ts = shape : position : ts
```
just95: I think `addTVars` corresponds to [`genericApply`](https://freeproving.github.io/free-compiler/docs/master/FreeC-Backend-Coq-Converter-Free.html#v:genericApply), `pure` roughtly corresponds to [`generatePure`](https://freeproving.github.io/free-compiler/docs/master/FreeC-Backend-Coq-Converter-Free.html#v:generatePure) and I think that `freeArgs` is more similar to [`genericApply`](https://freeproving.github.io/free-compiler/docs/master/FreeC-Backend-Coq-Converter-Free.html#v:genericApply) from `FreeC.Backend.Coq.Converter.Free` than `freeArgs` from `FreeC.Backend.Coq.Base`. I'd suggest to add `FreeC.Backend.Agda.Converter.Free` for such functions. The `FreeC.Backend.Agda.Base` module should mainly contain the names themselves.

src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
```diff
@@ -0,0 +1,76 @@
...
+  Agda.unqualify <$> renameAndDefineAgdaTypeVar srcSpan name
+
+-- | Lifts an n-ary function piece wise. Functions, which take all their
+--   arguments aren't evaluated until they are fully applied. Therefore partial
+--   application cannot produce @undefined@.
+--
+--   > (τ₁ → τ₂ → … → τₙ)* = τ₁' → τ₂' → … → τₙ'
+convertFunc :: [IR.Type] -> Converter Agda.Expr
+convertFunc ts =
+  foldr1 Agda.func <$> mapM (fmap Agda.Base.free' . convertType) ts
```
just95: I wouldn't use `*` in the comment. That symbol is reserved by the `*` translation by Abel et al. which has a different semantic (i.e., `(τ₁ → τ₂ → … → τₙ)* = τ₁' → Free (τ₂' → Free (… → τₙ'))`).

src/lib/FreeC/Backend/Agda/Converter/Module.hs
```diff
@@ -0,0 +1,26 @@
...
+convertModule = moduleEnv . (runPipeline >=> convertModule')
+
+-- | Like 'convertModule'' but does not apply any compiler passes beforehand.
+convertModule' :: IR.Module -> Converter Agda.Declaration
+convertModule' (IR.Module _ name _ _ _ _ funcDecls) =
+  Agda.moduleDecl (convertModName name) <$> concatMapM convertFuncDecl funcDecls
+
+convertModName :: IR.ModName -> Agda.QName
+convertModName name = Agda.qname (init parts) (last parts)
+  where parts = Agda.name <$> splitOn "." name
```
just95, line 28: Add documentation comments for all modules and top-level declarations.

src/lib/FreeC/Backend/Agda/Syntax.hs
```diff
@@ -70,8 +108,46 @@ intLiteral = Lit . LitNat NoRange
...
+
+-- | Creates a pi type binding the given names as hidden variables.
+--
+--   > pi [α, β, γ, …] expr = ∀ {α} {β} {γ} … → expr
+pi :: [Name] -> Expr -> Expr
+pi decls = Pi [TBind NoRange (bName <$> decls) (Underscore NoRange Nothing)]
+
+-- | Helper function for creating bound, named arguments.
+bName :: Name -> NamedArg Binder
+bName n = Arg hiddenArgInfo $ Named Nothing $ Binder Nothing $ mkBoundName_ n
```
just95: I would maybe call this `hiddenArg` instead.

src/lib/FreeC/Backend/Agda/Syntax.hs
```diff
@@ -70,8 +108,46 @@ intLiteral = Lit . LitNat NoRange
...
+-------------------------------------------------------------------------------
+
+-- | A smart constructor for non dependent function types.
+func :: Expr -> Expr -> Expr
+func l@(Fun _ _ _) = Fun NoRange (defaultArg (paren l)) -- (->) is right assoc.
+func l             = Fun NoRange (defaultArg l)
+
+-- | Creates a pi type binding the given names as hidden variables.
+--
+--   > pi [α, β, γ, …] expr = ∀ {α} {β} {γ} … → expr
```
just95: Trailing ellipsis indicate that a list is infinite to me. Thus, I would write something like `pi [α₁, …, αₙ]` instead.

src/lib/FreeC/Environment.hs
```diff
@@ -184,9 +193,18 @@ usedIdents :: Environment -> [Coq.Qualid]
 usedIdents = concatMap entryIdents . Map.elems . envEntries
  where
   entryIdents :: EnvEntry -> [Coq.Qualid]
-  entryIdents entry
-    | isConEntry entry = [entryIdent entry, entrySmartIdent entry]
-    | otherwise        = [entryIdent entry]
+  entryIdents entry =
+    entryIdent entry : (guard (isConEntry entry) $> entrySmartIdent entry)
```
just95: I would suggest to write `entryIdent entry : [entrySmartIdent entry | isConEntry entry]`. This notation wouldn't require additional imports and does not obfuscate the code as much.

src/lib/FreeC/Backend/Agda/Syntax.hs
```diff
@@ -9,37 +9,54 @@ module FreeC.Backend.Agda.Syntax
...
+qname modules unQName = foldr Qual (QName unQName) modules
 
 -- | Creates a qualified name using an empty list of module names.
 qname' :: Name -> QName
 qname' = qname []
 
+-- | Creates a new qualified name, by appending a number or incrementing it.
+nextQName :: QName -> QName
+nextQName (Qual modName qName) = Qual modName $ nextQName qName
+nextQName (QName unQName     ) = QName $ nextName unQName
```
just95: Where does `nextName` come from? I have the impression that this should be part of the renamer.

just95 (changes requested):

src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
JonasHoefer: This `do` block only needs the `Applicative` instance. I think it was the result of moving some code around. I decided to change it and handle the identifier lookup better. As a result the `do` block is now gone.
```haskell
lookupIdent :: IR.DeclIdent -> Converter Agda.Name
lookupIdent (IR.DeclIdent srcSpan name) =
  Agda.unqualify <$> lookupAgdaIdentOrFail srcSpan IR.ValueScope name
```

src/lib/FreeC/Backend/Agda/Base.hs
JonasHoefer: I suggest the following:
- `addTVars` becomes `addFreeArgs` (at the moment it seems reasonable to define the free args as function adding them, because they are only used in that way)
- `freeArgs` becomes `applyFreeArgs`. 
- In our Agda translation type parameters are always hidden, therefore there is no need for a `genericApply`, because we have never have to apply types to functions.

src/lib/FreeC/Backend/Agda/Syntax.hs
JonasHoefer: [`Agda.Syntax.Concrete.Name`](https://github.com/agda/agda/blob/1b426c067e0c807a21db5b84b8d68fe85e54521c/src/full/Agda/Syntax/Concrete/Name.hs#L269). I moved `nextQName` to `FreeC.Environment.Renamer`.

src/lib/FreeC/Backend/Agda/Converter/Free.hs
```diff
@@ -0,0 +1,38 @@
...
+generatePure =
+  Agda.app $ Agda.Ident $ Agda.qname [Agda.Base.baseLibName] Agda.Base.pure
+
+-- | Lifts a type in the free monad.
+free :: Agda.Expr -> Agda.Expr
+free = Agda.app $ applyFreeArgs $ Agda.qname' Agda.Base.free
+
+-- | Apply the @Shape@ and @Pos@ argument to the given type constructor.
+--
+-- > c ↦ c @S @P
```
just95: Aren't curly braces used in Agda instead of `@` to pass implicit arguments explicitly?

```suggestion
-- > c ↦ c {Shape} {Pos}
```

The implementation on the other hand suggests, that `Shape` and `Pos` are passed as regular arguments.

```suggestion
-- > c ↦ c Shape Pos
```

src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
```diff
@@ -0,0 +1,52 @@
...
+-- | Looks up the name of a Haskell function in the environment and converts it
+--   to an Agda name.
+lookupValueIdent :: IR.DeclIdent -> Converter Agda.Name
+lookupValueIdent (IR.DeclIdent srcSpan name) =
+  Agda.unqualify <$> lookupAgdaIdentOrFail srcSpan IR.ValueScope name
+
+-- | Converts a fully applied function.
+convertFunc_ :: [IR.TypeVarDecl] -> [Maybe IR.Type] -> Converter Agda.Expr
+convertFunc_ tVars ts = Agda.pi <$> tVars' <*> convertFunctionType
+  (fromJust <$> ts) -- handled in #19
```
just95: In this case `ts` is a value and not a computation. Thus, I'd use `map` instead of `<$>`. The same applies to line 38.

src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
```diff
@@ -0,0 +1,52 @@
...
+--   type signature and the definition (TODO).
+convertFuncDecl :: IR.FuncDecl -> Converter [Agda.Declaration]
+convertFuncDecl decl = localEnv $ sequence [convertSignature decl]
+
+-- | Converts the type signature of the given function to an Agda type
+--   declaration.
+convertSignature :: IR.FuncDecl -> Converter Agda.Declaration
+convertSignature (IR.FuncDecl _ ident tVars args retType _) =
+  Agda.funcSig <$> lookupValueIdent ident <*> convertFunc_ tVars types
+  where types = (IR.varPatType <$> args) `snoc` retType
```
just95: As indicated by the use of `snoc` here and `foldr1` in `convertFunctionType`, the representation of a function type `τ₁ -> … -> τₙ -> τ` as a list `[τ₁, …, τₙ, τ]` is not ideal. As the return type `τ` has a special role, it should have its own argument in `convertFunctionType`.

src/lib/FreeC/Backend/Agda/Converter/Type.hs
```diff
@@ -0,0 +1,87 @@
...
+convertConstructorType types =
+  -- We can use the @star@ translation for the data type, because only the name
+  -- and the application of type and @Size@ variables have to be translated.
+  Agda.fun <$> convertFunctionType (init types) <*> star (last types)
+
+-- | Utility function for introducing a new Agda type variable to the current
+--   scope.
+renameAgdaTypeVar :: IR.TypeVarDecl -> Converter Agda.Name
+renameAgdaTypeVar (IR.TypeVarDecl srcSpan name) =
+  Agda.unqualify <$> renameAndDefineAgdaTypeVar srcSpan name
```
just95, line 51: This seems to have a similar intend as `convertTypeVarDecls` from `FreeC.Backend.Coq.Converter.Arg`.

src/lib/FreeC/Backend/Agda/Converter/Type.hs
```diff
@@ -0,0 +1,87 @@
...
+--   polytype cases are not applicable for our IR.
+--   Note: If polytypes are added in the future (e.g. for type classes) add
+--   their type translation here.
+--
+--   > -- not applicable (polytypes)
+--   > (∀α:κ.σ)’ = (α : κ’) → σ’ -- RankNTypes translate to pi types
+--   > (σ ↦ τ)’ = σ’ → τ’        -- functions handling polytypes e.g.
+--   >                           -- result from type class dictionary translation
+--   > -- all other cases (monotypes):
+--   > τ’ = m τ
```
just95: Strictly speaking, there are some polytypes already in the form of `IR.TypeSchema`, they are just handled somewhere else (i.e., in type signatures). It's true that higher rank polymorphism is not supported and would have to be implemented here, but there are also no plans to add support for the `RankNTypes` language extension at the moment. Thus, I wouldn't mention monotypes and polytypes at all. They just confuse the reader.

```suggestion
--   > τ’ = Free τ*
```

Instead, the comment should specify what is done by the function. For example, in this case the function `converts a type from IR to Agda by lifting it into the @Free@ monad`. If you want you can add that `this corresponds to the dagger translation for monotypes as describer by Abel et al.`.

src/lib/FreeC/Backend/Agda/Converter/Type.hs
```diff
@@ -0,0 +1,87 @@
...
+dagger :: IR.Type -> Converter Agda.Expr
+dagger = fmap free . star
+
+-- | The star translation for monotypes as described by Abel et al.
+--
+-- > (σ τ)* = σ* τ*
+-- > (σ → τ)* = σ’ → τ’
```
just95: `σ` is often used for substitutions in the documentation. If multiple meta variables of the same type occur, they usually have indices. Thus, I'd write `τ₁` and `τ₂`.

src/lib/FreeC/Backend/Agda/Converter/Type.hs
```diff
@@ -0,0 +1,87 @@
...
+-- > C* = Ĉ @S @P
+-- > α* = α
```
just95:
```suggestion
-- > C* = Ĉ Shape Pos
-- > α* = α̂
```

just95 (changes requested):

src/lib/FreeC/Backend/Agda/Base.hs
just95: I'm not sure whether it is actually safe to leave all type parameters hidden. In Coq, we need visible type applications because of "vanishing type arguments". In `Nil = Nil`, for example, Coq cannot infer the type arguments of `Nil`. I guess that we will have the same problem in Agda. Thus, I think that we should still apply all type arguments visibly. However, we can add `genericApply` later. For now, `applyFreeArgs` should suffice.

src/lib/FreeC/Backend/Agda/Converter/Type.hs
```diff
@@ -0,0 +1,87 @@
...
+
+-- | Lifts a single type.
+convertType :: IR.Type -> Converter Agda.Expr
+convertType = dagger
+
+-- | Functions not defined in terms lambdas (@f x = …@ not @f = \x -> …@) aren't
+--   evaluated until fully applied, i.e. cannot be bottom. Therefore interleaving
+--   monadic layers isn't needed.
+--
+--   > (τ₁ → τ₂ → … → τₙ)' = τ₁' → Free τ₂' → … → Free τₙ'
```
just95:
```suggestion
--   > τ₁ -> … -> τₙ → τ ↦ τ₁' →… → τₙ' → τ'
```

src/lib/FreeC/Backend/Agda/Converter/Type.hs
```diff
@@ -0,0 +1,87 @@
...
+convertFunctionType :: [IR.Type] -> Converter Agda.Expr
+convertFunctionType types = foldr1 Agda.fun <$> mapM dagger types
+
+-- | Constructors aren't evaluated until fully applied. Furthermore the return
+--   type of a constructor for a data type D has to be D and therefore cannot be
+--   lifted.
+--
+--   > (τ₁ → τ₂ → … → τₙ → D)' = τ₁' → Free τ₂' → … → τₙ' → D
```
just95:
```suggestion
--   > τ₁ -> … -> τₙ -> τ ↦ τ₁' → … → τₙ' → τ*
```

ichistmeinname (approved):

src/lib/FreeC/Backend/Agda/Converter/Type.hs
```diff
@@ -0,0 +1,80 @@
...
+renameAgdaTypeVar :: IR.TypeVarDecl -> Converter Agda.Name
+renameAgdaTypeVar (IR.TypeVarDecl srcSpan name) =
+  Agda.unqualify <$> renameAndDefineAgdaTypeVar srcSpan name
+
+-------------------------------------------------------------------------------
+-- Translations                                                              --
+-------------------------------------------------------------------------------
+
+-- | Converts a type from IR to Agda by lifting it into the @Free@ monad.
+--   This corresponds to the dagger translation for monotypes as described by
```
just95: Ideally, our Haddock comments start with a single sentence summary. Thus, I'd move the additional details in its own paragraph.

```suggestion
--
--   This corresponds to the dagger translation for monotypes as described by
```

src/lib/FreeC/Backend/Agda/Converter/Type.hs
```diff
@@ -0,0 +1,80 @@
...
+-- > C* = Ĉ Shape Position
+-- > α* = α̂
+star :: IR.Type -> Converter Agda.Expr
+star (IR.TypeVar s name) = Agda.Ident
+  <$> lookupAgdaIdentOrFail s IR.TypeScope (IR.UnQual (IR.Ident name))
+star (IR.TypeCon s name) =
+  applyFreeArgs <$> lookupAgdaIdentOrFail s IR.TypeScope name
+star (IR.TypeApp  _ l r) = Agda.app <$> star l <*> star r
+-- At the moment this case simplifies to
+-- @Agda.func <$> free (dagger l) <*> free (dagger r)@.
```
just95: You probably meant `star` instead of `dagger` and `fmap free` instead of `free`. However, I think that this comment can be removed entirely.

```suggestion
```

src/lib/FreeC/Backend/Agda/Converter/Type.hs
```diff
@@ -0,0 +1,80 @@
...
+
+-------------------------------------------------------------------------------
+-- Translations                                                              --
+-------------------------------------------------------------------------------
+
+-- | Converts a type from IR to Agda by lifting it into the @Free@ monad.
+--   This corresponds to the dagger translation for monotypes as described by 
+--   Abel et al.
+--
+--   > τ' = Free τ
```
just95: This would mean that the type is only wrapped with `Free`. But it is also lifted recursively using `*`.

```suggestion
--   > τ' = Free τ*
```

src/lib/FreeC/Backend/Agda/Converter/Type.hs
```diff
@@ -0,0 +1,80 @@
...
+
+module FreeC.Backend.Agda.Converter.Type
+  ( convertType
+  , convertFunctionType
+  , convertConstructorType
+  , renameAgdaTypeVar
+  )
+where
+
+import           Control.Monad                  ( mapM )
```
just95: `mapM` is part of the `Prelude`.

src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
```diff
@@ -0,0 +1,52 @@
...
+--   type signature and the definition (TODO).
+convertFuncDecl :: IR.FuncDecl -> Converter [Agda.Declaration]
+convertFuncDecl decl = localEnv $ sequence [convertSignature decl]
+
+-- | Converts the type signature of the given function to an Agda type
+--   declaration.
+convertSignature :: IR.FuncDecl -> Converter Agda.Declaration
+convertSignature (IR.FuncDecl _ ident tVars args retType _) =
+  Agda.funcSig <$> lookupValueIdent ident <*> convertFunc_ tVars types
+  where types = (IR.varPatType `map` args) `snoc` retType
```
just95: `map` is usually written in prefix notation.

src/lib/FreeC/Backend/Agda/Converter/Type.hs
```diff
@@ -0,0 +1,80 @@
...
+
+-- | Converts a type from IR to Agda by lifting it into the @Free@ monad.
+--   This corresponds to the dagger translation for monotypes as described by 
+--   Abel et al.
+--
+--   > τ' = Free τ
+dagger :: IR.Type -> Converter Agda.Expr
+dagger = fmap free . star
+
+-- | The star translation for monotypes as described by Abel et al.
```
just95: Like the comment for `dagger` starts with "converts", the comment for `star` should start with a third-person verb, too. Noun phrases should be used in the documentation of data types and variable bindings but not for functions. This forces one to describe what functions do and what variables are. See also our [Haddock Styleguide](https://github.com/FreeProving/guidelines/blob/master/CONTRIBUTING.md#haddock-styleguide).

just95 (changes requested):

src/lib/FreeC/Backend/Agda/Converter/Type.hs
JonasHoefer: resolved in #58 ([`FreeC.Backend.Agda.Converter.Arg`](https://github.com/FreeProving/free-compiler/blob/issue-58/src/lib/FreeC/Backend/Agda/Converter/Arg.hs))

src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
```diff
@@ -0,0 +1,52 @@
...
+lookupValueIdent (IR.DeclIdent srcSpan name) =
+  Agda.unqualify <$> lookupAgdaIdentOrFail srcSpan IR.ValueScope name
+
+-- | Converts a fully applied function.
+convertFunc_ :: [IR.TypeVarDecl] -> [Maybe IR.Type] -> Converter Agda.Expr
+convertFunc_ tVars ts = Agda.pi <$> tVars' <*> convertFunctionType
+  (map fromJust ts) -- handled in #19
+  where tVars' = addFreeArgs <$> mapM renameAgdaTypeVar tVars
+
+
```
just95:
```suggestion
```

src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
```diff
@@ -0,0 +1,52 @@
...
+                                                , renameAgdaTypeVar
+                                                )
+import qualified FreeC.Backend.Agda.Syntax     as Agda
+import           FreeC.Environment.LookupOrFail ( lookupAgdaIdentOrFail )
+import qualified FreeC.IR.Syntax               as IR
+import           FreeC.Monad.Converter          ( Converter
+                                                , localEnv
+                                                )
+
+
```
just95:
```suggestion
```

src/lib/FreeC/Backend/Agda/Converter/Module.hs
```diff
@@ -0,0 +1,29 @@
...
+
+import           Data.List.Extra                ( splitOn )
+
+import           FreeC.Backend.Agda.Converter.FuncDecl
+import qualified FreeC.Backend.Agda.Syntax     as Agda
+import qualified FreeC.IR.Syntax               as IR
+import           FreeC.Monad.Converter
+import           FreeC.Pipeline
+
+
```
just95:
```suggestion
```

just95 (approved):

PR #60, opened 2020-06-10
Add check for kinds of type expressions #22

### LINKED ISSUE #22: Add check for kinds of type expressions
## Description

The compiler should report an error if a type constructor is applied to the wrong number of arguments or a type variable is applied to any type arguments.

## Motivation

The compiler currently only supports type variables of [kind](https://wiki.haskell.org/Kind) `*`. If type variables of other kinds were supported, we would have to implement a mechanism to infer the kinds of type variables. Therefore, it is currently not allowed to write a type like `m a` for example. However, the compiler does not check whether a type variable is actually never applied at the moment. In consequence, generated Coq code may contain type errors if the input program contains such a type variable application. For a user it may be difficult to understand the error reported by Coq. Thus, a check should be implemented into the Free Compiler that produces a better error message.

A consequence of the restriction above is that every type constructor has a kind of the form `* -> κ`, i.e., the arguments of all type constructors must be types not type constructors themselves. Therefore, a type expression must contain fully applied type constructors only since it could neither be used as a type nor passed as a type argument otherwise. Whether this requirement is met by the input program should be checked by the compiler as well.

## Tasks

- [x] Add a compiler [`Pass`](https://freeproving.github.io/free-compiler/docs/master/FreeC-Pass.html#t:Pass) to the compiler [`pipeline`](https://freeproving.github.io/free-compiler/docs/master/FreeC-Pipeline.html#v:pipeline) that performs the check. Perform at least the following checks and report a fatal error if any check fails.
    - [x] Type variables do not occur on the left-hand side of type applications.
    - [x] Every `n`-ary type constructor is applied to exactly `n` type arguments.
 - [x] Add documentation and test cases for your compiler `Pass`.

stu204146-kiel:
Add a new pass that checks whether type expressions do not have type variables as well as function types on the left-hand side of type expressions.
Additionally, this pass checks whether all type constructors are applied to the correct number of arguments.

Closes #22

### DIFF
--- a/free-compiler.cabal
+++ b/free-compiler.cabal
@@ -170,6 +170,7 @@ library freec-internal
                       , FreeC.Pass.ExportPass
                       , FreeC.Pass.ImplicitPreludePass
                       , FreeC.Pass.ImportPass
+                      , FreeC.Pass.KindInferencePass
                       , FreeC.Pass.PartialityAnalysisPass
                       , FreeC.Pass.TypeInferencePass
                       , FreeC.Pass.TypeSignaturePass
--- a/free-compiler.cabal
+++ b/free-compiler.cabal
@@ -170,7 +170,7 @@ library freec-internal
                       , FreeC.Pass.ExportPass
                       , FreeC.Pass.ImplicitPreludePass
                       , FreeC.Pass.ImportPass
-                      , FreeC.Pass.KindInferencePass
+                      , FreeC.Pass.KindCheckPass
                       , FreeC.Pass.PartialityAnalysisPass
                       , FreeC.Pass.TypeInferencePass
                       , FreeC.Pass.TypeSignaturePass
--- a/free-compiler.cabal
+++ b/free-compiler.cabal
@@ -170,6 +170,7 @@ library freec-internal
                       , FreeC.Pass.ExportPass
                       , FreeC.Pass.ImplicitPreludePass
                       , FreeC.Pass.ImportPass
+                      , FreeC.Pass.KindCheckPass
                       , FreeC.Pass.PartialityAnalysisPass
                       , FreeC.Pass.TypeInferencePass
                       , FreeC.Pass.TypeSignaturePass
@@ -234,6 +235,7 @@ test-suite freec-unit-tests
                     , FreeC.Monad.Class.Testable
                     , FreeC.Monad.ReporterTests
                     , FreeC.Pass.EtaConversionPassTests
+                    , FreeC.Pass.KindCheckPassTests
                     , FreeC.Pass.PartialityAnalysisPassTests
                     , FreeC.Pass.ResolverPassTests
                     , FreeC.Pass.TypeInferencePassTests
--- a/src/lib/FreeC/Pass/KindCheckPass.hs
+++ b/src/lib/FreeC/Pass/KindCheckPass.hs
@@ -139,7 +139,7 @@ checkAlt (IR.Alt _ _ varPats rhs) = do
 
 -- | Checks whether the type annotation of a variable pattern is kind-correct.
 checkVarPat :: IR.VarPat -> Converter ()
-checkVarPat (IR.VarPat _ _ typ) = mapM_ checkType typ
+checkVarPat (IR.VarPat _ _ typ _) = mapM_ checkType typ
 
 -- | Checks if all type constructors have the correct number of arguments
 checkType :: IR.Type -> Converter ()
--- a/src/lib/FreeC/Pass/KindCheckPass.hs
+++ b/src/lib/FreeC/Pass/KindCheckPass.hs
@@ -51,8 +51,8 @@
 --   All type constructors in type expressions are applied to the correct
 --   number of arguments and there are no type variables or function types on
 --   the left-hand side of type applications.
-module FreeC.Pass.KindInferencePass
-  ( kindInferencePass
+module FreeC.Pass.KindCheckPass
+  ( kindCheckPass
   )
 where
 
@@ -66,11 +66,11 @@ import           FreeC.Monad.Reporter
 import           FreeC.Pass
 import           FreeC.Pretty                   ( showPretty )
 
--- | A compiler pass that checks wheter there are type applications with type
+-- | A compiler pass that checks whether there are type applications with type
 --   variables or function types on the left-hand side in the module and checks
 --   whether all type construcors in the module are fully applied.
-kindInferencePass :: Pass IR.Module
-kindInferencePass m@(IR.Module _ _ _ typeDecls typeSigs _ funcDecls) = do
+kindCheckPass :: Pass IR.Module
+kindCheckPass m@(IR.Module _ _ _ typeDecls typeSigs _ funcDecls) = do
   mapM_ checkTypeDecl typeDecls
   mapM_ checkTypeSig  typeSigs
   mapM_ checkFuncDecl funcDecls
--- a/src/lib/FreeC/Pass/KindCheckPass.hs
+++ b/src/lib/FreeC/Pass/KindCheckPass.hs
@@ -1,4 +1,3 @@
-
 -- | This module contains a compiler pass that checks if all type constructors
 --   in type expressions are fully applied and type variables are not applied.
 --
@@ -18,7 +17,7 @@
 --   >
 --   > data MyList a = MyNil | MyCons a (MyList a)
 --   >
---   > -- Invalid because 'MyList' is applied to 2 arguments
+--   > -- Invalid because 'MyList' is applied to 2 arguments.
 --   > fail1 :: MyList (MyList a a) -> Integer
 --   > fail1 x = 42
 --   >
@@ -81,7 +80,7 @@ kindCheckPass m@(IR.Module _ _ _ typeDecls typeSigs _ funcDecls) = do
 -- | Checks whether all type expressions in a type declaration are correct.
 checkTypeDecl :: IR.TypeDecl -> Converter ()
 checkTypeDecl (IR.DataDecl    _ _ _ conDecls) = mapM_ checkConDecl conDecls
-checkTypeDecl (IR.TypeSynDecl _ _ _ typ     ) = checkType typ
+checkTypeDecl (IR.TypeSynDecl _ _ _ typeExpr) = checkType typeExpr
 
 -- | Checks whether the arguments of a constructor declaration are kind-correct.
 checkConDecl :: IR.ConDecl -> Converter ()
--- a/src/lib/FreeC/Pass/KindCheckPass.hs
+++ b/src/lib/FreeC/Pass/KindCheckPass.hs
@@ -1,0 +1,11 @@
+module FreeC.Pass.KindCheckPass where
+
+import qualified FreeC.IR.Syntax               as IR
+import           FreeC.Monad.Converter
+import           FreeC.Pass
+
+kindCheckPass :: Pass IR.Module
+kindCheckPass = return
+
+checkType :: IR.Type -> Converter ()
+checkType _ = return ()
--- a/src/lib/FreeC/Pass/KindCheckPass.hs
+++ b/src/lib/FreeC/Pass/KindCheckPass.hs
@@ -18,16 +18,16 @@
 --   >
 --   > data MyList a = MyNil | MyCons a (MyList a)
 --   >
---   > -- Invalid because MyList is applied to 2 arguments
+--   > -- Invalid because 'MyList' is applied to 2 arguments
 --   > fail1 :: MyList (MyList a a) -> Integer
 --   > fail1 x = 42
 --   >
---   > -- Invalid because type variable a is on the right hand side of a types
---   > -- application
+--   > -- Invalid because type variable 'a' is on the right hand side of a type
+--   > -- application.
 --   > fail2 :: a a -> Integer
 --   > fail2 x = 42
 --   >
---   > -- Fails because MyList is applied to no arguments
+--   > -- Fails because 'MyList' is applied to no arguments.
 --   > fail3 :: MyList -> Integer
 --   > fail3 x = 42
 --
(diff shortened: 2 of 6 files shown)


src/lib/FreeC/Pass/KindInferencePass.hs
```diff
@@ -0,0 +1,211 @@
+
+-- | This module contains a compiler pass that checks if all type constructors
+--   in type expressions are fully applied and type variables are not applied.
+--
+--   For simplification, all type variables have kind @*@. Therefore, all type
+--   constructors must be fully applied, so partial application of type
+--   constructors is not allowed. This pass also checks that type variables and
+--   function types do not occur on the left hand side of a type application.
+--   If a type expression is found that does not match this rules, a fatal
+--   error is thrown. For simplification, a type expression that satisfies all
```
just95:
```suggestion
--   error is reported. For simplification, a type expression that satisfies all
```

src/lib/FreeC/Pass/KindInferencePass.hs
```diff
@@ -0,0 +1,211 @@
...
+--   of these rules is called kind-correct.
+--
+--   = Example
+--
+--   The following module contains three invalid definitions
```
just95:
```suggestion
--   The following module contains three invalid definitions.
```

You could maybe also add a short explanation for each declaration why it is invalid.

src/lib/FreeC/Pass/KindInferencePass.hs
```diff
@@ -0,0 +1,211 @@
...
+--   >
+--   > fail3 :: MyList -> Integer
+--   > fail3 x = 42
+--
+--   = Specification
+--
+--   == Preconditions
+--
+--   The environment must contain all type declarations that are used in the
+--   module
```
just95: The environment does not contain declarations but entries for declarations.

```suggestion
--   The environment must contain entries for all type constructors that are
--   used in the module.
```

End every sentence with a period in the comments and error messages below as well.

src/lib/FreeC/Pass/KindInferencePass.hs
```diff
@@ -0,0 +1,211 @@
...
+--
+--   == Translations
+--
+--   The AST is not changed. It is checked if all type constructor applications
+--   have the right number of arguments. It is also checked whether there are
+--   type variables or function types on the left-hand side of type
+--   applications. In these cases, a fatal error is thrown. Otherwise, the
```
just95:
```suggestion
--   applications. In these cases, a fatal error is reported. Otherwise, the
```

src/lib/FreeC/Pass/KindInferencePass.hs
```diff
@@ -0,0 +1,211 @@
...
+-- |
+checkMaybeTypeSchema :: Maybe IR.TypeSchema -> Converter ()
+checkMaybeTypeSchema Nothing  = return ()
+checkMaybeTypeSchema (Just x) = checkTypeSchema x
+
+-- | Checks whether an optional type expression is kind correct in case the
+--   type expression is present
+checkMaybeType :: Maybe IR.Type -> Converter ()
+checkMaybeType Nothing  = return ()
+checkMaybeType (Just x) = checkType x
```
just95: `checkMaybeType` and `checkMaybeTypeSchema` are `mapM_ checkType` and `mapM_ checkTypeSchema`, respectively. You don't have to add a new function if it just applies `mapM_`.

src/lib/FreeC/Pass/KindInferencePass.hs
```diff
@@ -0,0 +1,211 @@
...
+checkVarPat :: IR.VarPat -> Converter ()
+checkVarPat (IR.VarPat _ _ typ) = checkMaybeType typ
+
+-- | Checks whether a type is kind-correct
+checkType :: IR.Type -> Converter ()
+checkType = checkCorrectArities
+
+-- | Checks if all type constructors have the correct number of arguments
+checkCorrectArities :: IR.Type -> Converter ()
+checkCorrectArities = checkCorrectArities' 0
```
just95: A single function would suffice here.

```suggestion
-- | Checks whether a type is kind-correct.
checkType :: IR.Type -> Converter ()
checkType = checkType' 0
```

src/lib/FreeC/Pass/KindInferencePass.hs
```diff
@@ -0,0 +1,211 @@
...
+
+-- | Helper function for @checkCorrectArities@
+checkCorrectArities' :: Int -> IR.Type -> Converter ()
+checkCorrectArities' depth (IR.TypeVar srcSpan varId) =
```
just95: Please add a brief explanation to the comment what the purpose of the additional argument is.

src/lib/FreeC/Pass/KindInferencePass.hs
```diff
@@ -0,0 +1,211 @@
...
+  arity <- inEnv $ fromMaybe (-1) . lookupArity IR.TypeScope ident
+  when (arity /= depth)
+    $  reportFatal
+    $  Message srcSpan Error
+    $  "Type constructor "
+    ++ showPretty ident
+    ++ " is applied to wrong number of arguments: expected "
+    ++ show arity
+    ++ " but was "
+    ++ show depth
```
just95:
```suggestion
    ++ " is applied to wrong number of arguments. Expected "
    ++ show arity
    ++ " but was "
    ++ show depth
    ++ "."
```

free-compiler.cabal
```diff
@@ -170,6 +170,7 @@ library freec-internal
                       , FreeC.Pass.ExportPass
                       , FreeC.Pass.ImplicitPreludePass
                       , FreeC.Pass.ImportPass
+                      , FreeC.Pass.KindInferencePass
```
just95: Since the `KindInferencePass` does not actually infer any kinds, but checks whether every type actually has kind `*`, I'd call it `KindCheckPass` instead.

just95 (changes requested):

src/lib/FreeC/Pass/KindInferencePass.hs
```diff
@@ -0,0 +1,211 @@
...
+import           Data.Maybe                     ( fromMaybe )
+
+import           FreeC.Environment              ( lookupArity )
+import qualified FreeC.IR.Syntax               as IR
+import           FreeC.Monad.Converter
+import           FreeC.Monad.Reporter
+import           FreeC.Pass
+import           FreeC.Pretty                   ( showPretty )
+
+-- | A compiler pass that checks wheter there are type applications with type
```
Daniel-Teut:
```suggestion
-- | A compiler pass that checks whether there are type applications with type
```

src/lib/FreeC/Pass/KindInferencePass.hs
```diff
@@ -0,0 +1,211 @@
...
+checkMaybeTypeSchema (Just x) = checkTypeSchema x
+
+-- | Checks whether an optional type expression is kind correct in case the
+--   type expression is present
+checkMaybeType :: Maybe IR.Type -> Converter ()
+checkMaybeType Nothing  = return ()
+checkMaybeType (Just x) = checkType x
+
+-- | Checks whether all type declarations have kind-correct type expressions
+checkTypeDecls :: [IR.TypeDecl] -> Converter ()
```
Daniel-Teut: As with the `checkMaybe` functions I don't think it is necessary to add an extra function for a list of elements, if all it does, is applying `_mapM`.

Daniel-Teut (changes requested):

src/lib/FreeC/Pass/KindCheckPass.hs
```diff
@@ -0,0 +1,180 @@
...
+--
+--   = Example
+--
+--   The following module contains three invalid definitions.
+--
+--   > module Failures where
+--   >
+--   > data MyList a = MyNil | MyCons a (MyList a)
+--   >
+--   > -- Invalid because MyList is applied to 2 arguments
```
just95: Comments should be full sentences. End them with a period (also check the other comments in this file). 

```suggestion
--   > -- Invalid because 'MyList' is applied to 2 arguments.
```

Furthermore, I'd use Haddock notation in the examples. In this particular case the reader can understand what `MyList` is even without markup. However, in the comment for `fail2` it is not entirely clear whether `a` is the variable name or a misplaced article.

src/lib/FreeC/Pass/KindCheckPass.hs
```diff
@@ -0,0 +1,180 @@
...
+  arity <- inEnv $ fromMaybe (-1) . lookupArity IR.TypeScope ident
+  when (arity /= depth)
+    $  reportFatal
+    $  Message srcSpan Error
+    $  "Type constructor "
+    ++ showPretty ident
+    ++ " is applied to wrong number of arguments: Expected "
+    ++ show arity
+    ++ " but was "
+    ++ show depth
```
just95, line 168: Try to end error messages with a period.

src/lib/FreeC/Pass/KindCheckPass.hs
```diff
@@ -0,0 +1,180 @@
...
+checkTypeSchema :: IR.TypeSchema -> Converter ()
+checkTypeSchema (IR.TypeSchema _ _ typ) = checkType typ
+
+-- | Checks whether a function declaration has kind-correct type annotations
+checkFuncDecl :: IR.FuncDecl -> Converter ()
+checkFuncDecl (IR.FuncDecl _ _ _ varPats retType rhs) = do
+  mapM_ checkVarPat varPats
+  mapM_ checkType   retType
+  checkExpr rhs
+  return ()
```
just95: `()` is returned by `checkExpr` already,

```suggestion
```

src/lib/FreeC/Pass/KindCheckPass.hs
```diff
@@ -0,0 +1,180 @@
...
+checkExpr (IR.Var _ _ typeSchema      ) = mapM_ checkTypeSchema typeSchema
+checkExpr (IR.App _ lhs rhs typeSchema) = do
+  checkExpr lhs
+  checkExpr rhs
+  mapM_ checkTypeSchema typeSchema
+checkExpr (IR.TypeAppExpr _ lhs rhs typeSchema) = do
+  checkExpr lhs
+  checkType rhs
+  mapM_ checkTypeSchema typeSchema
+checkExpr (IR.If _ cond thenExp elseExp typeSchema) = do
```
just95: Try to abbreviate consistently. Expression are abbreviated `expr` instead of `exp` in `IR`.

```suggestion
checkExpr (IR.If _ cond thenExpr elseExpr typeSchema) = do
```

In some other places you wrote `typ` to abbreviate types. Unfortunately, the identifier `type` cannot be used. Thus, `type` is "abbreviated" `typeExpr` by most compiler passes.

just95 (changes requested):

Daniel-Teut (changes requested):
Looks good to me, but you should merge in the unit tests in the `issue-22-tests` branch.

Daniel-Teut (approved):

src/lib/FreeC/Pass/KindCheckPass.hs
```diff
@@ -0,0 +1,182 @@
...
+--
+--   = Example
+--
+--   The following module contains three invalid definitions.
+--
+--   > module Failures where
+--   >
+--   > data MyList a = MyNil | MyCons a (MyList a)
+--   >
+--   > -- Invalid because 'MyList' is applied to 2 arguments
```
just95:
```suggestion
--   > -- Invalid because 'MyList' is applied to 2 arguments.
```

src/test/FreeC/Pass/KindCheckPassTests.hs
```diff
@@ -0,0 +1,175 @@
+module FreeC.Pass.KindCheckPassTests where
```
just95, line 3:
```suggestion
-- | This module contains tests for "FreeC.Pass.KindCheckPass".

module FreeC.Pass.KindCheckPassTests where
```

src/lib/FreeC/Pass/KindCheckPass.hs
```diff
@@ -0,0 +1,182 @@
+
```
just95:
```suggestion
```

src/lib/FreeC/Pass/KindCheckPass.hs
```diff
@@ -0,0 +1,182 @@
...
+kindCheckPass m@(IR.Module _ _ _ typeDecls typeSigs _ funcDecls) = do
+  mapM_ checkTypeDecl typeDecls
+  mapM_ checkTypeSig  typeSigs
+  mapM_ checkFuncDecl funcDecls
+  return m
+
+-- | Checks whether all type expressions in a type declaration are correct.
+checkTypeDecl :: IR.TypeDecl -> Converter ()
+checkTypeDecl (IR.DataDecl    _ _ _ conDecls) = mapM_ checkConDecl conDecls
+checkTypeDecl (IR.TypeSynDecl _ _ _ typ     ) = checkType typ
```
just95:
```suggestion
checkTypeDecl (IR.TypeSynDecl _ _ _ typeExpr) = checkType typeExpr
```

just95 (changes requested):

src/test/FreeC/Pass/KindCheckPassTests.hs
```diff
@@ -26,26 +26,26 @@ testValidTypes = context "valid types" $ do
...
+      _ <- defineTestTypeCon "Int" 0 []
       checkType input
   it "should accept fully applied type synonyms" $ do
     input <- expectParseTestType "State Int Int"
     shouldSucceed $ do
       _ <- defineTestTypeSyn "State" ["s", "a"] "s -> (,) s a"
-      _ <- defineTestTypeCon "Int" 0
-      _ <- defineTestTypeCon "(,)" 2
+      _ <- defineTestTypeCon "Int" 0 []
+      _ <- defineTestTypeCon "(,)" 2 []
```
ichistmeinname, line 48: If the last argument resembles the constructor names, it shouldn't be empty for a pair, right?

src/test/FreeC/Pass/KindCheckPassTests.hs
```diff
@@ -107,28 +107,28 @@ testNotValidTypes = context "not valid types" $ do
...
       checkType input
   it "should not accept overapplied type constructors" $ do
     input <- expectParseTestType "State Int Int Int"
     shouldFail $ do
-      _ <- defineTestTypeCon "State" 2
-      _ <- defineTestTypeCon "Int" 0
+      _ <- defineTestTypeCon "State" 2 []
```
ichistmeinname, line 123: The same question applies here: should we have constructor names here as well?

ichistmeinname (changes requested):

src/test/FreeC/Pass/KindCheckPassTests.hs
just95: In this test case we do not care whether the data type has constructors or not. It is just important that there is a binary type constructor (which was called `(,)` in this case but must not be confused for the pair data type from Haskell). If the names of the constructors were provided, additional environment entries for the constructors would also have to be specified.

ichistmeinname: So it's a bit of a hack to use this function in such a way, because, originally, it was supposed to be used differently? ; )

just95: I wouldn't call it a hack. Since the environment in test cases should only contain the information that is necessary for the test, the code below implies that it is important that the binary data type has a data constructor. 

```haskell
_ <- defineTestTypeCon "(,)" 2 ["(,)"]
_ <- defineTestCon "(,)" 2 "forall a b. a -> b -> (,) a b"
```

But that is not the case. The additional `defineTestCon` just distracts from the `defineTestTypeSyn`. Thus, I think it is right to specify no constructors. I understand that the use of well known names such as `(,)` may be confusing at first. However, names such as `Foo` and `Bar` don't communicate the arity of the type constructors at all, but the arity (especially of `State`) is very important for the test cases.

just95 (approved):

ichistmeinname (approved):

PR #61, opened 2020-06-12
Implement translation of non-recursive datatype declarations #58

### LINKED ISSUE #58: Implement translation of non-recursive datatype declarations (with smart constructors)
Implement the translation of non-recursive datatype declarations. Add smart constructors for the `Free` lifted version using `pattern` declarations.

JonasHoefer:
Closing #58


src/lib/FreeC/Backend/Agda/Converter/TypeDecl.hs
```diff
@@ -0,0 +1,101 @@
...
+-- | Converts a single constructor of a (non-recursive) data type.
+convertConstructor :: IR.Type -> IR.ConDecl -> Converter Agda.Declaration
+convertConstructor retType (IR.ConDecl _ ident argTs) =
+  Agda.funcSig
+    <$> lookupValueIdent ident
+    <*> convertConstructorType argTs retType
+
+-- | Converts a list of constructors to a list of smart constructors.
+convertSmartConstrs :: [IR.ConDecl] -> Converter [Agda.Declaration]
+convertSmartConstrs = mapM convertSmartConstr
```
just95: I am not a huge fan of abbreviations. But if we abbreviate names, we should do so consistently. Since `ConstructorDeclaration` is abbreviated `ConDecl` in IR, I would rename `convertConstructor` to `convertConDecl` and `convertSmartConstrs` to `generateSmartConDecl` (I prefer the prefix `generate` instead of `convert` if the conversion function generates a syntactic construct that does not exist in Haskell). I also prefer long names over single letter abbreviations (e.g., `argTypes` and `typeVars` instead of `argTs` and `tVars`).

src/lib/FreeC/Backend/Agda/Syntax.hs
```diff
@@ -140,7 +188,6 @@ hiddenArg :: Name -> NamedArg Binder
 hiddenArg n =
   Arg hiddenArgInfo $ Named Nothing $ Binder Nothing $ mkBoundName_ n
 
--- | Argument meta data marking them as hidden.
```
just95: Why did you remove the comment? Also functions that are not exported and functions in `where` blocks should be documented even if the comments don't make it into the documentation generated by Haddock.

src/lib/FreeC/Environment/Fresh.hs
```diff
@@ -9,11 +9,14 @@ module FreeC.Environment.Fresh where
 import           Data.List                      ( elemIndex )
 import qualified Data.Map.Strict               as Map
 
+import qualified FreeC.Backend.Agda.Syntax     as Agda
 import qualified FreeC.Backend.Coq.Syntax      as Coq
 import           FreeC.Environment
 import           FreeC.Environment.Renamer
 import           FreeC.IR.SrcSpan
 import qualified FreeC.IR.Syntax               as IR
+import qualified FreeC.IR.SrcSpan              as IR
```
just95: `FreeC.IR.SrcSpan` does not have to be imported `qualified` since you have imported it unqualified above already. The module also does not have to be imported qualified in general. All modules that must be imported qualified are listed in `.hlint.yaml`. All other modules should be imported unqualified if possible.

src/lib/FreeC/Backend/Agda/Converter/TypeDecl.hs
```diff
@@ -0,0 +1,101 @@
...
+--
+--   We use Agda's @pattern@ declarations for smart constructors to simplify
+--   pattern matching in proofs.
+convertSmartConstr :: IR.ConDecl -> Converter Agda.Declaration
+convertSmartConstr (IR.ConDecl _ ident argTs) = do
+  smartName <- lookupSmartIdent ident
+  patternDecl smartName (repeat "x" `zip` argTs) $ \vars -> do
+    normalName <- Agda.IdentP . Agda.qname' <$> lookupValueIdent ident
+    let pureVal = foldl Agda.appP normalName vars
+    pure (Agda.IdentP (Agda.qname' Agda.Base.pure) `Agda.appP` pureVal)
```
just95: Try to use `return` instead of `pure` (especially if you are in a `do` block). I would use `pure` only if there is no `Monad` instance.

src/lib/FreeC/Backend/Agda/Converter/TypeDecl.hs
```diff
@@ -0,0 +1,101 @@
...
+  -> [(String, IR.Type)]
+  -- ^ Types and preferred names for the bound variables.
+  -> ([Agda.Pattern] -> Converter Agda.Pattern)
+  -- ^ Continuation for creating the definition using the new variables.
+  -> Converter Agda.Declaration
+patternDecl name vars k = localEnv $ do
+  names <- mapM (uncurry freshAgdaVar) vars
+  let decls = map (Agda.Arg Agda.defaultArgInfo . Agda.unqualify) names
+  let varPatterns = map Agda.IdentP names
+  Agda.patternSyn name decls <$> k varPatterns
```
just95, line 119: I would inline this into `convertSmartConstr`. It is used nowhere else and the continutation makes the code difficult to comprehend.

src/lib/FreeC/Environment/Fresh.hs
```diff
@@ -114,3 +117,13 @@ freshCoqQualid :: String -> Converter Coq.Qualid
 freshCoqQualid prefix = do
   ident <- freshHaskellIdent prefix
   inEnv $ renameQualid ident
+
+-- | Generates a new Agda identifer based on the given name.
+--
+--   Note: Type is redundant with next merge from master and the introduction
+--   of @Fresh@
```
just95: Better use the `TODO` keyword in such comments instead of `Note`. I periodically search for `TODO` comments and resolve them.

src/lib/FreeC/Backend/Agda/Converter/Arg.hs
```diff
@@ -0,0 +1,48 @@
+-- | This module contains functions for generating Agda function, constructor
+--   and type arguments from our intermediate representation.
+
+module FreeC.Backend.Agda.Converter.Arg
+  ( convertTypeVarDecl
+  , lookupIdent
+  , lookupValueIdent
+  , lookupTypeIdent
+  , lookupSmartIdent
```
just95: I don't think the `lookup` functions belong into this module. Furthermore, the name `lookupIdent` isn't ideal. It can easily be confused for [`FreeC.Environment.lookupIdent`](https://freeproving.github.io/free-compiler/docs/master/FreeC-Environment.html#v:lookupIdent). Consider to inline the definitions of these functions instead of just moving them to their own module. That would make much more explicit what's actually happening. If that causes expressions to become unreadable, consider to fall back to `do` notation and assign names to intermediate results.

src/lib/FreeC/Backend/Agda/Converter/Free.hs
```diff
@@ -6,9 +6,12 @@ module FreeC.Backend.Agda.Converter.Free
   , free
   , applyFreeArgs
   , addFreeArgs
+  , freeDataDecl
```
just95: Since `freeDataDecl` is not of general purpose, it should be part of `FreeC.Backend.Agda.Converter.TypeDecl`.

src/lib/FreeC/Backend/Agda/Converter/TypeDecl.hs
```diff
@@ -0,0 +1,101 @@
...
+                                                , lookupValueIdent
+                                                , lookupTypeIdent
+                                                , lookupSmartIdent
+                                                )
+import           FreeC.Environment.Fresh        ( freshAgdaVar )
+import           FreeC.Monad.Converter          ( Converter
+                                                , localEnv
+                                                )
+
+
```
just95:
```suggestion

```

just95 (changes requested):

src/lib/FreeC/Backend/Agda/Syntax.hs
```diff
@@ -158,6 +158,15 @@ ident = Ident . qname' . name
 set :: Expr
 set = ident "Set"
 
+-- | Creates a @data@ declaration with the given name, binding the list of type
+--   variables and defining the list of constructors.
+--
+--   > name [(α₁ : eₙ), …, {αₙ : eₙ}] [C₁, …, Cₘ]
+--   >   ↧
+--   > data name (α₁ : eₙ) … {αₙ : eₙ} : Set where
```
just95: To me it is not obvious what `(α₁ : eₙ) … {αₙ : eₙ}` should mean exactly. I guess that you want to express that the binders can be explicit or implicit. But the notation suggests that all but the last binder are explicit. I would just introduce meta variables `b₁ … bₙ` to express that the binders can be arbitrary.

Instead of `name` I'd suggest to use `D` since that is the meta variable we usually use for the names of data type declarations in the remaining documentation.

just95 (changes requested):

src/lib/FreeC/Backend/Agda/Syntax.hs
JonasHoefer: I probably cut instead of copied it, while writing the comment for the non hidden version.

src/lib/FreeC/Backend/Agda/Converter/TypeDecl.hs
JonasHoefer: In the first version of the code the two functions were combined in a single one. The result was quite large and also hard to comprehend. I think it's logical to split the function in the generation of the syntactic construct (a `pattern` declaration binding variables) and the actual code generation (the smart constructor itself). 
Abstracting the `pattern` declaration as a function taking a continuation also has the advantage of forcing the caller to generate the code in a `localEnv` and mirroring the actual generated scope in Agda.

src/lib/FreeC/Backend/Agda/Converter/Free.hs
```diff
@@ -32,7 +33,22 @@ applyFreeArgs qname = foldl1
...
+-- | Adds the reserved names for the free args @Shape@ and @Pos@ to a list of
 --   names.
 addFreeArgs :: [Agda.Name] -> [Agda.Name]
 addFreeArgs ts = Agda.Base.shape : Agda.Base.position : ts
+
+-- | Identifier for @Shape@.
+--
+--   > Shape
+shape :: Agda.Expr
+shape = Agda.Ident $ Agda.qname' $ Agda.Base.shape
```
ichistmeinname: If you cannot decide between explicit parentheses and `$`, please always go for parentheses instead! (In line 32 you use parentheses, for example.)

src/lib/FreeC/Backend/Agda/Converter/TypeDecl.hs
```diff
@@ -0,0 +1,115 @@
...
+-- | Converts a single constructor of a (non-recursive) data type.
+convertConstructor :: IR.Type -> IR.ConDecl -> Converter Agda.Declaration
+convertConstructor retType (IR.ConDecl _ ident argTypes) =
+  Agda.funcSig
+    <$> lookupValueIdent ident
+    <*> convertConstructorType argTypes retType
+
+-- | Converts a list of constructors to a list of smart constructors.
+generateSmartConDecls :: [IR.ConDecl] -> Converter [Agda.Declaration]
+generateSmartConDecls = mapM generateSmartConDecl
```
ichistmeinname: I think it's okay to use `mapM generateSmartConDecl` in the one place you actually need it instead of defining this extra function.

ichistmeinname (changes requested):

src/lib/FreeC/Backend/Agda/Converter/Free.hs
JonasHoefer: I usually use `$` when applying two or more functions to the same argument to avoid writing nested parentheses. I think `f (g (h x))` is harder to read then `f $ g $ h x`. I can of course change it.

just95 (approved):
The code looks good to me. However, we should probably start to add some test cases for the Agda back end.

src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
```diff
@@ -31,22 +31,17 @@ convertFuncDecl decl = localEnv $ sequence [convertSignature decl]
...
-  Agda.unqualify <$> lookupAgdaIdentOrFail srcSpan IR.ValueScope name
+convertSignature (IR.FuncDecl _ (IR.DeclIdent srcSpan name) tVars args retType _)
+  = Agda.funcSig <$> ident <*> convertFunc_ tVars types retType
+ where
+  types = map IR.varPatType args
+  ident = lookupUnQualAgdaIdentOrFail srcSpan IR.ValueScope name
 
 -- | Converts a fully applied function.
 convertFunc_
   :: [IR.TypeVarDecl] -> [Maybe IR.Type] -> Maybe IR.Type -> Converter Agda.Expr
```
ichistmeinname: Maybe that's just me not knowing the code well enough, but was such a naming scheme used before? It's first time I stumble upon a suffix `_`.

src/lib/FreeC/Backend/Agda/Converter/TypeDecl.hs
```diff
@@ -0,0 +1,110 @@
...
+                                                ( convertConstructorType )
+import           FreeC.Backend.Agda.Converter.Arg
+                                                ( convertTypeVarDecl )
+import           FreeC.Environment.Fresh        ( freshAgdaVar )
+import           FreeC.Environment.LookupOrFail
+import           FreeC.Monad.Converter          ( Converter
+                                                , localEnv
+                                                )
+
+-- | Converts a data or type synonym declaration.
```
ichistmeinname, line 26: Maybe the comment should mention, that the current state does not handle type synonyms (even if it's only temporary due to vertical prototyping)?

src/lib/FreeC/Backend/Agda/Converter/TypeDecl.hs
```diff
@@ -0,0 +1,110 @@
...
+
+-- | Converts a data declaration.
+convertDataDecl
+  :: IR.DeclIdent
+  -> [IR.TypeVarDecl]
+  -> [IR.ConDecl]
+  -> Converter Agda.Declaration
+convertDataDecl ident@(IR.DeclIdent srcSpan name) typeVars constrs =
+  localEnv
+    $   freeDataDecl
```
ichistmeinname: As I'm not quite sure what happens here, the comment should explain what the function does a bit more?

src/lib/FreeC/Backend/Agda/Converter/TypeDecl.hs
```diff
@@ -0,0 +1,110 @@
...
+
+-- | Creates a new pattern declaration binding variables with the given names
+--   and types.
+patternDecl
+  :: Agda.Name
+  -- ^ Name of the declaration.
+  -> [(String, IR.Type)]
+  -- ^ Types and preferred names for the bound variables.
+  -> ([Agda.Pattern] -> Converter Agda.Pattern)
+  -- ^ Continuation for creating the definition using the new variables.
```
ichistmeinname, line 113: I guess that you need to use a new line for the comments here due to the 80 chars rule? ; )

ichistmeinname (changes requested):

src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
JonasHoefer: The Agda code base uses this naming structure sometimes for  specialized versions of the same function (e.g. `mkBinder_` and `mkBinder`, `mkBoundName_` and `mkBoundName`). I use this notation for some smart constructors in `FreeC.Backend.Agda.Syntax`.
In this case the name is probably just a result of renaming and moving functions arround. I will change it.

src/lib/FreeC/Backend/Agda/Converter/TypeDecl.hs
JonasHoefer: That's right. This should conform to [haddocks comment style](https://haskell-haddock.readthedocs.io/en/latest/markup.html#constructors-and-record-fields). `-- ^` always refers to the argument before the comment. If only one `-- ^` is found after a declaration it refers to the whole function/constructor instead of the return type/last argument.
Writing the comments over two lines reduced readability and made them blend together.

ichistmeinname (approved):
Good work!

PR #64, opened 2020-06-14
Fix compilation failures for `undefined` and `error`

### LINKED ISSUE #62: Compilation fails if `undefined` is applied
### Description

The compiler produces an error when it tries to translate a program where `undefined` is applied to something. 

### Steps to Reproduce

Attempt to compile the following Haskell program.

`f x = undefined x`

#### Expected behavior

The function should be translated to 

```
Definition f (Shape : Type) (Pos : Shape -> Type) (P : Partial Shape Pos) {t0 t1
   : Type} (x : Free Shape Pos t0)
   : Free Shape Pos t1 :=
  @undefined Shape Pos P (Free Shape Pos t0 -> Free Shape Pos t1) >>= (fun f_0 => f_0 x).
```

#### Actual behavior

The compiler suffers an internal error:

`Only type arguments of functions and constructors can be applied visibly.`

However, translating
```
f x = g x
g = undefined
```
works correctly and produces 

```
Definition f (Shape : Type) (Pos : Shape -> Type) (P : Partial Shape Pos) {t0 t1
   : Type} (x : Free Shape Pos t0)
   : Free Shape Pos t1 :=
  @g Shape Pos P (Free Shape Pos t0 -> Free Shape Pos t1) >>= (fun f_0 => f_0 x).
```
for `f`. 

### Possible Causes 

The issue most likely lies in the pattern matching in the function `convertExpr'` in the module `FreeC.Backend.Coq.Converter.Expr`.
The function does not seem to consider that `undefined` could be applied, so we eventually land in a catch-all case that reports the error seen above. 

### Versions

 - Git commit: `da173dd843d843adab0def78d20a626cdadcbb29`
 - `ghc --version`:  `8.6.5`
 - `cabal --version`: `3.2.0.0`
 - `coqc --version`: `8.11.0`
 - OS: Ubuntu 19.10

stu204767:
As described in #62, the Free Compiler suffers an internal error if `undefined` is applied to anything, because the compiler doesn't expect the type of `undefined` to be a function. The same behavior occurs with `error`. This pull request fixes both bugs and adds some test cases for them. Closes #62.

### DIFF
--- a/src/lib/FreeC/Backend/Coq/Converter/Expr.hs
+++ b/src/lib/FreeC/Backend/Coq/Converter/Expr.hs
@@ -41,7 +41,7 @@ convertExpr' :: IR.Expr -> [IR.Type] -> [IR.Expr] -> Converter Coq.Term
 
 -- Constructors.
 --
--- Partially applied constructors are not evaluated in Haskell and therefor
+-- Partially applied constructors are not evaluated in Haskell and therefore
 -- cannot be @⊥@. The translated type of a constructor @C : τ₀ -> … -> τₙ@ is
 -- @c : τ₀' -> … -> τₙ*@ instead of @m(τ₀' -> m(τ₁' -> m(… -> τₙ')))@.
 --
--- a/src/lib/FreeC/Backend/Coq/Converter/Expr.hs
+++ b/src/lib/FreeC/Backend/Coq/Converter/Expr.hs
@@ -216,7 +216,7 @@ convertExpr' (IR.ErrorExpr srcSpan msg _) typeArgs args = do
   args'     <- mapM convertExpr args
   let partialArg = [Coq.Qualid (fst Coq.Base.partialArg)]
       callee     = genericApply Coq.Base.partialError partialArg typeArgs' []
-  generateApplyN (1 :: Int)
+  generateApplyN 1
                  callee
                  (Coq.InScope (Coq.string msg) Coq.Base.stringScope : args')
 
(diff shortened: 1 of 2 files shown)


MarvinLira (approved):

src/lib/FreeC/Backend/Coq/Converter/Expr.hs
```diff
@@ -210,14 +212,13 @@ convertExpr' (IR.ErrorExpr srcSpan msg _) typeArgs [] = do
...
-  return
-    (genericApply Coq.Base.partialError
-                  [partialArg]
-                  typeArgs'
-                  [Coq.InScope (Coq.string msg) Coq.Base.stringScope]
-    )
+  args'     <- mapM convertExpr args
+  let partialArg = [Coq.Qualid (fst Coq.Base.partialArg)]
+      callee     = genericApply Coq.Base.partialError partialArg typeArgs' []
+  generateApplyN (1 :: Int)
```
just95: Is the type annotation really needed?

```suggestion
  generateApplyN 1
```

just95 (changes requested):

src/lib/FreeC/Backend/Coq/Converter/Expr.hs
stu204767: No, but not type-annotating an Integer gave me a warning once while I tried different solutions for this issue. But this type annotation isn't needed here because the type annotation of `generateApplyN` specifies the first argument to be of type `Int`.

just95 (approved):

PR #67, opened 2020-06-16
Test cases for #22

Daniel-Teut:
Adds unit tests for the `kindCheckPass` in #22.

### DIFF
--- a/src/test/FreeC/Pass/KindCheckPassTests.hs
+++ b/src/test/FreeC/Pass/KindCheckPassTests.hs
@@ -1,0 +1,43 @@
+module FreeC.Pass.KindCheckPassTests where
+
+import           Test.Hspec
+
+import           FreeC.Monad.Class.Testable
+import           FreeC.Pass.KindCheckPass
+import           FreeC.Test.Environment
+import           FreeC.Test.Parser
+
+testKindCheckPass :: Spec
+testKindCheckPass = describe "FreeC.Pass.KindCheckPassTests" $ do
+  testValidTypes
+  testNotValidTypes
+
+testValidTypes :: Spec
+testValidTypes = context "valid types" $ do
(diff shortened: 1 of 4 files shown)


src/test/FreeC/Pass/KindCheckPassTests.hs
```diff
@@ -0,0 +1,182 @@
...
+      "should not accept type variable applications in type annotated case expression variables"
+    $ do
+        input <- expectParseTestModule
+          ["module M where", "f x = case x of {C (y :: m b) -> y};"]
+        shouldFail $ do
+          _ <- defineTestTypeVar "m"
+          _ <- defineTestTypeVar "a"
+          _ <- defineTestTypeVar "b"
+          _ <- defineTestVar "x"
+          _ <- defineTestVar "y"
```
ichistmeinname: When you have a lot of variables to introduce, it might be still readable enough when you map over all variables you like to introduce?

```suggestion
          _ <- mapM defineTestTypeVar ["m", "a", "b"]
          _ <- mapM defineTestVar ["x","y"]
```

ichistmeinname (approved):

stu204146-kiel (changes requested):
I would add comments to the test file

ichistmeinname:
> I would add comments to the test file

Good point, but as far as I haven't skimmed the other test files too quickly, the current status quo is that there are not comments associated to the tests.
I suggest that, instead, @Daniel-Teut opens a general issue with an _enhancement_ label to start adding comments to out test cases. I can imagine that the comments then reflect the corresponding Haskell-code that is tested.

That said, if you do not have other remarks, @stu204146-kiel, can you accept this PR, such that I can merge it into your branch, so we can merge your original ticket today as well?

stu204146-kiel:
I have no further remarks so the PR can be merged.

PR #69, opened 2020-06-17
 Implement adding size annotations to type expressions #59

### LINKED ISSUE #59: Implement adding size annotations to type expressions

JonasHoefer:
Adds `Size` annotations to data declarations and function signatures. To find recursive functions and data types `FreeC.IR.DependencyGraph` is used. Therefore the Agda conversion functions now work with `DependencyComponent`s. Closes #59


src/lib/FreeC/Backend/Agda/Analysis/RecursiveDataType.hs
```diff
@@ -0,0 +1,26 @@
+-- | This module contains functions for analysing data declarations. The goal is
+--   to decide if a data type is recursive.
+module FreeC.Backend.Agda.Analysis.RecursiveDataType
+  ( isRecursiveDataType
+  )
+where
```
just95: There is already a dependency analysis that also detects recursive data type declarations in [`FreeC.IR.DependencyGraph`](https://freeproving.github.io/free-compiler/docs/master/FreeC-IR-DependencyGraph.html). The dependency analysis will be moved entirely to passes in the future (see #71 and #72). For now, you should use [`groupTypeDecls`](https://freeproving.github.io/free-compiler/docs/master/FreeC-IR-DependencyGraph.html#v:groupTypeDecls) to find strongly connected components (i.e., [`DependencyComponent`s](https://freeproving.github.io/free-compiler/docs/master/FreeC-IR-DependencyGraph.html#t:DependencyComponent)) of the type dependency graph in `convertModule'`.

src/lib/FreeC/Backend/Agda/Base.hs
```diff
@@ -62,6 +65,22 @@ shape = Agda.name "Shape"
 position :: Agda.Name
 position = Agda.name "Pos"
 
+-------------------------------------------------------------------------------
+-- Sized Types                                                               --
+-------------------------------------------------------------------------------
+
+-- | Size type.
```
just95: The first sentence of the documentation of constants should be non phrases.

 ```suggestion
-- | The name of Agda's @Size@ data type.
```

File: src/lib/FreeC/Backend/Agda/Base.hs [hunk 3]
```diff
@@ -62,6 +65,22 @@ shape = Agda.name "Shape"
...
+--
+--   > Size
+size :: Agda.Name
+size = Agda.name "Size"
+
+-- | Agda function for larger size.
+--
+-- > e ↦ ↑ e
```
just95: The comment `e ↦ ↑ e` suggests that `up` maps Agda expressions `e` to expressions that are applied to `↑`. However, `up` is only the name of the function `↑`.

src/lib/FreeC/Backend/Agda/Base.hs
```diff
@@ -11,6 +11,9 @@ module FreeC.Backend.Agda.Base
   , pure
   , shape
   , position
+    -- * sized types
+  , size
+  , up
```
just95, line 16: I'm not too sure whether the names of sized types should be in this module since they are not part of our Agda base library.

src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
```diff
@@ -31,17 +35,22 @@ convertFuncDecl decl = localEnv $ sequence [convertSignature decl]
...
 convertSignature :: IR.FuncDecl -> Converter Agda.Declaration
-convertSignature (IR.FuncDecl _ (IR.DeclIdent srcSpan name) tVars args retType _)
-  = Agda.funcSig <$> ident <*> convertFunc tVars types retType
- where
-  types = map IR.varPatType args
-  ident = lookupUnQualAgdaIdentOrFail srcSpan IR.ValueScope name
+convertSignature (IR.FuncDecl _ (IR.DeclIdent s name) tVars args rType _) = do
+  let types = map IR.varPatType args
+  let ident = lookupUnQualAgdaIdentOrFail s IR.ValueScope name
+  decArg <- inEnv $ lookupDecArgIndex name
```
just95: The environment contains only the decreasing argument index of functions whose decreasing argument has been annotated using pragmas. The information from the environment is used by [`identifyDecArgs`](https://github.com/FreeProving/free-compiler/blob/93344610284e5d9998fe8199e80576e13b552a64/src/lib/FreeC/Backend/Coq/Analysis/DecreasingArguments.hs#L266). You would have to reuse that function here (or add a comment that `lookupDecArgIndex` is used only temporarily).

Regarding the renaming of arguments (e.g., `retType → rType`), I actually prefer the long names. In any case, you should try to use variable names consistently (at least within the same module). In `convertSignature` you call the return type `rType` but `returnType` in `convertFunc`. Sometimes you abbreviate `type` (as in `tVars`), but sometimes you don't (as in `rType`). Personally, I try to add as much information into variable names as possible. For example, I'd rename `tVars` to `typeArgs` because they are not just some type variables, but the type arguments of the function. Also try to stick to the naming conventions that have been established by the reaming source code. For example, variables of type `SrcSpan` are always called something like `srcSpan`. This term should almost never be abbreviated.

src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
```diff
@@ -31,17 +35,22 @@ convertFuncDecl decl = localEnv $ sequence [convertSignature decl]
...
+  Agda.funcSig <$> ident <*> convertFunc decArg tVars types rType
```
just95: I think the following would be easier to understand.

```haskell
ident <- lookupUnQualAgdaIdentOrFail s IR.ValueScope name
…
Agda.funcSig ident <$> …
```

src/lib/FreeC/Backend/Agda/Converter/Size.hs
```diff
@@ -0,0 +1,21 @@
...
+where
+
+import qualified FreeC.Backend.Agda.Syntax     as Agda
+import qualified FreeC.Backend.Agda.Base       as Agda.Base
+
+-- | Hidden Size argument.
+--
+--   > {Size}
+size :: Agda.Expr
+size = Agda.Ident $ Agda.qname' Agda.Base.size
```
just95: Isn't this just the type expression `Size`? The hidden argument `{Size}` should be `Agda.hiddenArg_ size`.

src/lib/FreeC/Backend/Agda/Converter/TypeDecl.hs
```diff
@@ -42,16 +49,21 @@ convertTypeDecl (IR.DataDecl _ ident tVars constrs) =
...
-  -> [IR.TypeVarDecl]
-  -> [IR.ConDecl]
+  :: IR.DeclIdent     -- ^ The name of the data type
+  -> [IR.TypeVarDecl] -- ^ Type parameters for the data type.
+  -> [IR.ConDecl]     -- ^ The constructors for this data type.
+  -> Bool             -- ^ Is this a recursive data declaration?
   -> Converter Agda.Declaration
-convertDataDecl ident@(IR.DeclIdent srcSpan name) typeVars constrs =
+convertDataDecl ident@(IR.DeclIdent srcSpan name) typeVars constrs isRec =
   localEnv -- The data declaration opens a new scope binding @S@, @P and α̂ᵢ.
```
just95:
```suggestion
  localEnv -- The data declaration opens a new scope binding @S@, @P@ and @α̂ᵢ@.
```

src/lib/FreeC/Backend/Agda/Converter/TypeDecl.hs
```diff
@@ -61,19 +73,22 @@ convertDataDecl ident@(IR.DeclIdent srcSpan name) typeVars constrs =
 convertConDecl
   :: IR.DeclIdent     -- ^ The identifier of the data type.
   -> [IR.TypeVarDecl] -- ^ The type parameters declared by the data type.
+  -> Bool             -- ^ Is this a recursive data declaration
```
just95:
```suggestion
  -> Bool             -- ^ Whether this is a recursive data declaration.
```

src/lib/FreeC/Backend/Agda/Converter/Type.hs
```diff
@@ -66,10 +96,36 @@ dagger = fmap free . star
...
+star _ (IR.FuncType _ l r) = Agda.fun <$> dagger' l <*> dagger' r
+
+dagger' :: IR.Type -> Converter Agda.Expr
+dagger' = dagger Nothing
+
+star' :: IR.Type -> Converter Agda.Expr
+star' = star Nothing
+
+-- | Is the given type a type application, which applies arguments to the given
+--   constructor?
```
just95: Start comments of functions with third person verbs.

```suggestion
-- | Tests whether the given type is a type application, which applies
--   arguments to the given constructor?
```

src/lib/FreeC/Backend/Agda/Converter/Type.hs
```diff
@@ -66,10 +96,36 @@ dagger = fmap free . star
...
-star (IR.TypeApp  _ l r) = Agda.app <$> star l <*> star r
-star (IR.FuncType _ l r) = Agda.fun <$> dagger l <*> dagger r
+star _ (IR.TypeApp  _ l r) = Agda.app <$> star' l <*> star' r
+star _ (IR.FuncType _ l r) = Agda.fun <$> dagger' l <*> dagger' r
+
+dagger' :: IR.Type -> Converter Agda.Expr
+dagger' = dagger Nothing
+
+star' :: IR.Type -> Converter Agda.Expr
+star' = star Nothing
```
just95: Add documentation for `pi`, `dagger'`, `star'` and the additional parameter of `dagger` and `star`. I'd also suggest to find better names for these functions. For example, `dagger'` could just be called `convertType` and `dagger` something like `convertSizedType`.

src/lib/FreeC/Backend/Agda/Syntax.hs
```diff
@@ -189,8 +193,11 @@ fun l             = Fun NoRange (defaultArg l)
 --
 --   > pi [α₁, …, αₙ] expr ↦ ∀ {α₁} … {αₙ} → expr
 pi :: [Name] -> Expr -> Expr
-pi decls =
-  Pi [TBind NoRange (hiddenArg <$> decls) (Underscore NoRange Nothing)]
+pi decls (Pi binder expr) = Pi
+  (TBind NoRange (hiddenArg <$> decls) (Underscore NoRange Nothing) : binder)
+  expr
+pi decls expr =
+  Pi [TBind NoRange (hiddenArg <$> decls) (Underscore NoRange Nothing)] expr
```
just95: Use `<$>` for computations and `map` for data.

```suggestion
pi decls (Pi binder expr) = Pi
  (TBind NoRange (map hiddenArg decls) (Underscore NoRange Nothing) : binder)
  expr
pi decls expr =
  Pi [TBind NoRange (map hiddenArg decls) (Underscore NoRange Nothing)] expr
```

This can probably also be rewritten such that there is less code duplication.

src/lib/FreeC/Backend/Agda/Converter/Type.hs
```diff
@@ -66,10 +96,36 @@ dagger = fmap free . star
...
+--   constructor?
+--
+--   > appliesTo (((C x₁) …) xₙ) C ↦ True
+appliesTo :: IR.Type -> IR.QName -> Bool
+appliesTo (IR.TypeApp _ l _    ) name = l `appliesTo` name
+appliesTo (IR.TypeCon _ conName) name = conName == name
+appliesTo _                      _    = False
+
+-------------------------------------------------------------------------------
+-- specialized syntax                                                        --
```
just95: Capitalize the first word of headings as well as all other words except for articles, conjunctions and prepositions.

```suggestion
-- Specialized Syntax                                                        --
```

src/lib/FreeC/Backend/Agda/Converter/TypeDecl.hs
```diff
@@ -6,17 +6,24 @@ module FreeC.Backend.Agda.Converter.TypeDecl
   )
 where
 
+import           Prelude                 hiding ( pi )
```
just95: As far as I can tell, this import isn't needed.

src/lib/FreeC/Backend/Agda/Converter/TypeDecl.hs
```diff
@@ -61,19 +73,22 @@ convertDataDecl ident@(IR.DeclIdent srcSpan name) typeVars constrs =
...
+    NoSrcSpan
+    (IR.TypeCon srcSpan ident)
+    (map IR.typeVarDeclToType typeVars)
 
 -- | Converts a single constructor of a (non-recursive) data type.
-convertConstructor :: IR.Type -> IR.ConDecl -> Converter Agda.Declaration
-convertConstructor retType (IR.ConDecl _ (IR.DeclIdent srcSpan name) argTypes)
+convertConstructor
+  :: Maybe IR.QName -> IR.Type -> IR.ConDecl -> Converter Agda.Declaration
+convertConstructor ident retType (IR.ConDecl _ (IR.DeclIdent srcSpan name) argTypes)
```
just95: This should be named something with `ConDecl` as well (`convertConDecl` should probably be named `convertConDecls` and this function `convertConDecl`). Add documentation for the new argument. Is this really only used for non-recursive data types?

just95 (changes requested):

src/lib/FreeC/Backend/Agda/Base.hs
JonasHoefer: At the moment the names are the same as in the [Agda stdlib](http://agda.github.io/agda-stdlib/Agda.Builtin.Size.html). Therefore if we plan to use the standard library (e.g. by reexporting the relevant parts via our base lib) we have to reserve the names somewhere. We could create a second module storing names from the stdlib. If we wanted to we could also remove the std lib dependency and define own names (preferably the same) for the builtin sized types constructs in our base library.

just95: Reexporting the names from the base library sounds like a good plan :+1:

src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
JonasHoefer: Yes, my plan was to use `lookupDecArg` only temporarily to test the conversion code. Since i have to use the analyis modules anyway I will change this module to use `identifyDecArgs`. 

I shortened the names in this function, because brittany did not enforce the 80 column limit. The code after formatting with the long names is (line 2 is 87 characters long!):
```haskell
convertSignature :: IR.FuncDecl -> Converter Agda.Declaration
convertSignature (IR.FuncDecl _ (IR.DeclIdent srcSpan name) typeVars args returnType _)
  = do
    let types = map IR.varPatType args
    let ident = lookupUnQualAgdaIdentOrFail srcSpan IR.ValueScope name
    decArg <- inEnv $ lookupDecArgIndex name
    Agda.funcSig <$> ident <*> convertFunc decArg typeVars types returnType
```

just95: That Brittany does not break long left-hand sides annoys me as well. I would still go for the long names, though. Meaningful variable names improve the comprehensibility of code more than long lines hurt readability. One way to avoid long lines is to use selector functions of the `IR`. I've now [added some more notes](https://github.com/FreeProving/guidelines/commit/2aeb86fe57557ebb07f24fb61ac63db16c49c7bf) about the naming conventions to the Haskell Styleguide and clarified that it is okay if left-hand sides exceed the line length limit.

src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
```diff
@@ -2,46 +2,67 @@
...
- where
-  types = map IR.varPatType args
-  ident = lookupUnQualAgdaIdentOrFail srcSpan IR.ValueScope name
+convertSignature :: IR.FuncDecl -> Maybe Int -> Converter Agda.Declaration
+convertSignature (IR.FuncDecl _ declIdent typeVars args returnType _) decArg =
+  do
+    let IR.DeclIdent srcSpan name = declIdent
+    let types                     = map IR.varPatType args
+    ident <- lookupUnQualAgdaIdentOrFail srcSpan IR.ValueScope name
+    Agda.funcSig ident <$> convertFunc decArg typeVars types returnType
```
just95:
```suggestion
    let argTypes                  = map IR.varPatType args
    ident <- lookupUnQualAgdaIdentOrFail srcSpan IR.ValueScope name
    Agda.funcSig ident <$> convertFunc decArg typeVars argTypes returnType
```

src/lib/FreeC/Backend/Agda/Converter/Type.hs
```diff
@@ -66,10 +98,47 @@ dagger = fmap free . star
...
+convertSizedType' _ (IR.TypeVar s name) = Agda.Ident
+  <$> lookupAgdaIdentOrFail s IR.TypeScope (IR.UnQual $ IR.Ident name)
+convertSizedType' _ (IR.TypeCon s name) =
   applyFreeArgs <$> lookupAgdaIdentOrFail s IR.TypeScope name
-star (IR.TypeApp  _ l r) = Agda.app <$> star l <*> star r
-star (IR.FuncType _ l r) = Agda.fun <$> dagger l <*> dagger r
+convertSizedType' _ (IR.TypeApp _ l r) =
+  Agda.app <$> convertType' l <*> convertType' r
+convertSizedType' _ (IR.FuncType _ l r) =
+  Agda.fun <$> convertType l <*> convertType r
```
just95: If I understand this code correctly a size annotation `i` is added to `τ` by generating `τ* i`. However in `convertTypeDecl`, the size is a hidden argument. Thus, I think `τ* {i}` should be generated. I'm also not sure whether it is always correct to put the size annotation at the outermost position. For example if `τ` is a type variable, it cannot be applied to a size. It may also be necessary to add size annotations to the arguments of type constructor applications and to function types recursively.

src/lib/FreeC/Backend/Agda/Syntax.hs
```diff
@@ -150,6 +151,9 @@ paren = Paren NoRange
 ident :: String -> Expr
 ident = Ident . qname' . name
 
+hiddenArg_ :: Expr -> Expr
+hiddenArg_ = HiddenArg NoRange . unnamed
```
just95, line 158: Add documentation for this function.

just95 (changes requested):

src/lib/FreeC/Backend/Agda/Converter/Type.hs
JonasHoefer: The `Maybe Agda.Expr` is the already hidden type variable. [`convertRecFuncType`](https://github.com/FreeProving/free-compiler/blob/84cf653e30f6a080ab849721c3acca597b36cb04/src/lib/FreeC/Backend/Agda/Converter/Type.hs#L45) and [`convertRecConType`](https://github.com/FreeProving/free-compiler/blob/84cf653e30f6a080ab849721c3acca597b36cb04/src/lib/FreeC/Backend/Agda/Converter/Type.hs#L68) call the transformations with the already hidden argument. Since all type annotations are hidden it seems reasonable to move the `Agda.hiddenArg_` call to `convertSizedType'`.
`convertSizedType` and `convertSizedType'` are always called from inside this module. I therefore assumed that `τ` is a type constructor if a size variable is passed. Applying the size variable to other nodes in the AST is of course possible (the first version of my code took a variable of type `Maybe (Agda.QName, Agda.Expr)` and searched for type constructors with the given name to apply the variable to; describing where to put an annotation shouldn't be a problem).
There are of course cases where the decreasing argument is inside another type. The following Agda code does type check and pass termination checks. 
```agda
data Pairₜ (S : Set) (P : S → Set) (A B : Set) : Set where
  pair : Free S P A → Free S P B → Pairₜ S P A B

pattern Pair x y = pure (pair x y)

weirdMap : ∀ {S P A B C i} → Free S P (Free S P A → Free S P B) 
                           → Free S P (Pairₜ S P (List S P A {i}) C) 
                           → Free S P (Pairₜ S P (List S P B) C)
weirdMap mf acs = acs >>= λ where
  (pair as mc) → as >>= λ where
    nil           → Pair [] mc
    (cons mx mxs) → weirdMap mf (Pair mxs mc) >>= λ where
      (pair mbs mc) → mf >>= λ where
        f → Pair (f mx ∷ mbs) mc
```
The issue with examples like this one is of course finding the decreasing argument (because it is passed to a constructor before the recursive call) and then linking it back to a part of the type.
The problem I see with applying the same size annotations to more then type constructor is constraining the type of the function. Assume `f` take an argument of type `List A i × List B i` but decreases just on the first component of the pair. The recursive call would occur on a variable of type `List A j × List B i` with `j < i` which does not type check, because `i` and `j` are clearly not equal.

just95 (approved):

nbun (approved):

PR #87, opened 2020-06-23
Prevent name conflicts between imported modules

### LINKED ISSUE #20: Name conflict between imported modules
### Description

The compiler usually renames identifiers during the translation to Coq. When the same identifier is used in two different modules, they are not renamed. Importing both modules can lead to a name conflict in Coq.

### Steps to Reproduce

1. Consider two Haskell modules

    ```haskell
    module A where
    data Foo = Bar

    module B where
    data Bar = Foo
    ```

   and compile them using the following command.

    ```bash
    freec -o generated A.hs B.hs 
    ```

    The identifiers `Foo` and `Bar` are not renamed by `freec` because there are no name conflicts when the individual modules are defined in Coq.

2. Now consider the following Haskell module that imports the two other modules.

    ```haskell
    module C where

    import A
    import B

    mkFoo :: Foo
    mkFoo = Bar

    mkBar :: Bar
    mkBar = Foo
    ```

    In Haskell there are no name conflicts because the namespaces for types and values are separated.

3. Compile the importing module as follows.

     ```bash
     freec -o generated C.hs 
     ```

#### Expected behavior

There should be no name conflicts in the generated Coq code for `C.hs`.

#### Actual behavior

The following Coq code is generated for `mkFoo`.

```coq
Definition mkFoo (Shape : Type) (Pos : Shape -> Type)
   : Free Shape Pos (Foo Shape Pos) :=
  Bar Shape Pos.
```

In Coq there is no distinction between the namespaces of values and types. Thus, the identifiers `Foo` and `Bar` are ambigious. Since `B` is imported after `A`, `Foo` and `Bar` refer to `B.Foo` and `B.Bar` respectively. However, in Haskell they refer to `A.Foo` and `A.Bar`.

### Versions

 - `freec --version`: `The Free Compiler, version 0.1.0.0 (v0.1.0.0-549-g8665b41, ghc 8.6, linux, x86_64)`
 - `ghc --version`: `8.6.5`
 - `cabal --version`: `2.4.1.0`
 - `coqc --version`: `8.11.0`
 - OS: Ubuntu 19.10

### Tasks

In order to fix this bug and avoid similar bugs in the future, all identifiers should be explicitly qualified in Coq.

 - [x] Try to find more examples where name conflicts can occur in Coq due to imports.
 - [ ] Add (failing) test cases for the examples.
 - [x] Adapt the translation of `import` declarations such that external identifiers must be qualified in Coq.
 - [ ] Qualify Coq identifiers of [environment entries](https://freeproving.github.io/free-compiler/docs/master/FreeC-Environment-Entry.html) when the [module interface](https://freeproving.github.io/free-compiler/docs/master/FreeC-Environment-ModuleInterface.html) is imported or exported.

MajaRet:
### Issue

Closes #20 

### Description of the Change

As described in the issue, if there were two modules that used the same identifiers and then those two modules were imported by a third module, there was a conflict between names as imported identifiers were not renamed or qualified.

If this pull request is merged, any time a module is exported, the Coq identifier in the module interface will be qualified with the name of the module. 
This way, such name conflicts are avoided.
The change occurs in the compiler's `ExportPass`. 

Additionally, the translation of `Import` declarations was changed so that external names can no longer be referred to by their short names in Coq, but have to be qualified. This was done by changing the translation of an import declaration (except those importing from the `Base` library) from the `From Generated Require Export [Module]` sentence to `From Generated Require [Module]`. 

Names defined in the `Base` library do not have to be qualified.

### Verification Process
I tested the implementation by translating settings like the one described in #20, with different constellations of clashing names. 
I tested that chained imports were translated correctly.
I tested that the change worked in the presence of renaming (for example, by having a data type declaration `data Foo = Foo`).
I also tested that type synonyms could still be handled correctly by translating the example given in the documentation of the compiler's `Import Pass`:

```haskell
module A where 
data Foo = Foo

module B where 
import A
type Bar a = Foo -> a

module C where 
import B
baz :: Bar ()
baz x = ()
```
All examples produced compilable Coq programs.

Dedicated test cases should be added as soon as possible.

### Additional Notes

If this pull request is merged, all external identifiers are required to be qualified with their original module name in the generated Coq code.
As a consequence, Coq programs importing the generated modules (e.g. proofs of the generated code) must now use qualified names if they refer to modules imported into the generated module. 

E.g. `ListFunctorProofs` in `examples` imports the module `ListFunctor` and refers to `map` and `id` in its proofs, which are imported from the modules `Data.List` and `Data.Function` in `ListFunctor`. These references must now be qualified as `Data.List.map` and `Data.Function.id` in `ListFunctorProofs`.

### DIFF
--- a/src/lib/FreeC/Backend/Coq/Syntax.hs
+++ b/src/lib/FreeC/Backend/Coq/Syntax.hs
@@ -37,6 +37,7 @@ module FreeC.Backend.Coq.Syntax
     -- * Imports
   , requireImportFrom
   , requireExportFrom
+  , requireFrom
   )
 where
 
@@ -223,3 +224,8 @@ requireImportFrom library modules = ModuleSentence
 requireExportFrom :: ModuleIdent -> [ModuleIdent] -> Sentence
 requireExportFrom library modules = ModuleSentence
   (Require (Just library) (Just Export) (NonEmpty.fromList modules))
+
+-- | Creates a @From ... Require ...@ sentence.
+requireFrom :: ModuleIdent -> [ModuleIdent] -> Sentence
+requireFrom library modules = ModuleSentence
+  (Require (Just library) Nothing (NonEmpty.fromList modules))
--- a/src/lib/FreeC/Backend/Coq/Syntax.hs
+++ b/src/lib/FreeC/Backend/Coq/Syntax.hs
@@ -227,5 +227,5 @@ requireExportFrom library modules = ModuleSentence
 
 -- | Creates a @From ... Require ...@ sentence.
 requireFrom :: ModuleIdent -> [ModuleIdent] -> Sentence
-requireFrom library modules = ModuleSentence
-  (Require (Just library) Nothing (NonEmpty.fromList modules))
+requireFrom library modules =
+  ModuleSentence (Require (Just library) Nothing (NonEmpty.fromList modules))
--- a/src/lib/FreeC/Backend/Coq/Syntax.hs
+++ b/src/lib/FreeC/Backend/Coq/Syntax.hs
@@ -215,17 +215,17 @@ disj t1 t2 = app (Qualid (bare "op_\\/__")) [t1, t2]
 -- Imports                                                                   --
 -------------------------------------------------------------------------------
 
--- | Creates a @From ... Require Import ...@ sentence.
+-- | Creates a @From … Require Import …@ sentence.
 requireImportFrom :: ModuleIdent -> [ModuleIdent] -> Sentence
 requireImportFrom library modules = ModuleSentence
   (Require (Just library) (Just Import) (NonEmpty.fromList modules))
 
--- | Creates a @From ... Require Export ...@ sentence.
+-- | Creates a @From … Require Export …@ sentence.
 requireExportFrom :: ModuleIdent -> [ModuleIdent] -> Sentence
 requireExportFrom library modules = ModuleSentence
   (Require (Just library) (Just Export) (NonEmpty.fromList modules))
 
--- | Creates a @From ... Require ...@ sentence.
+-- | Creates a @From … Require …@ sentence.
 requireFrom :: ModuleIdent -> [ModuleIdent] -> Sentence
 requireFrom library modules =
   ModuleSentence (Require (Just library) Nothing (NonEmpty.fromList modules))
--- a/src/lib/FreeC/Pass/ExportPass.hs
+++ b/src/lib/FreeC/Pass/ExportPass.hs
@@ -101,13 +101,13 @@ exportInterface modName = do
                      }
     )
  where
-   -- | Tests whether to export the given entry.
-   --
-   --   Only top-level entries that are declared in the module are exported.
-   --   Since the original names of entries are qualified with the name of the
-   --   module they are declared in, we can tell whether an entry is declared
-   --   in the exported module by comparing the prefix of its original name
-   --   with the module name.
+  -- | Tests whether to export the given entry.
+  --
+  --   Only top-level entries that are declared in the module are exported.
+  --   Since the original names of entries are qualified with the name of the
+  --   module they are declared in, we can tell whether an entry is declared
+  --   in the exported module by comparing the prefix of its original name
+  --   with the module name.
   isExported :: EnvEntry -> Bool
   isExported entry = case entryName entry of
     IR.Qual modName' _ -> modName' == modName
--- a/src/lib/FreeC/Pass/ExportPass.hs
+++ b/src/lib/FreeC/Pass/ExportPass.hs
@@ -8,8 +8,8 @@
 --   can be expanded properly.
 --   Additionally, all Coq identifiers of exported entries are qualified with
 --   their original module name in the module interface.
---   This prevents name conflicts when several modules that use the same identifier
---   are imported.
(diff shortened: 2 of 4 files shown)


src/lib/FreeC/Pass/ExportPass.hs
```diff
@@ -62,6 +66,9 @@ where
 
 import qualified Data.Map.Strict               as Map
 import qualified Data.Set                      as Set
+import qualified Data.Text                     as Text
+
+import           Language.Coq.Gallina
```
just95: Use the `FreeC.Backend.Coq.Syntax` wrapper module instead of `language-coq` directly.

```suggestion
import qualified FreeC.Backend.Coq.Syntax      as Coq
```

src/lib/FreeC/Pass/ExportPass.hs
```diff
@@ -103,3 +113,23 @@ exportInterface modName = do
...
+                 , entrySmartIdent = qualifyIdent (entrySmartIdent entry)
+                 }
+      else entry { entryIdent = qualifyIdent (entryIdent entry) }
+    | otherwise = entry
+
+
+  -- Qualifies an unqualified Coq identifier with the module name.
+  -- A qualified Coq identifer is not modified.
+  qualifyIdent :: Qualid -> Qualid
+  qualifyIdent (         Bare coqName ) = Qualified (Text.pack modName) coqName
```
just95: There is a smart constructor [`Coq.ident`](https://github.com/FreeProving/free-compiler/blob/1ca3159363e039f60c2b5cedb10ff5b7938d7f90/src/lib/FreeC/Backend/Coq/Syntax.hs#L68) (which is just an alias for `Text.pack`) for [`Coq.Ident`s and `Coq.ModuleIdent`s](https://github.com/FreeProving/language-coq/blob/c36521c742d0c5a41775c8caf4c514e45d8430c4/src/lib/Language/Coq/Gallina.hs#L90) (which both are just an alias for `Text`). Another alternative would be to add a smart constructor `Coq.qualified :: String -> String -> Coq.Qualid` analogous to [`Coq.bare :: String -> Coq.Qualid`](https://github.com/FreeProving/free-compiler/blob/1ca3159363e039f60c2b5cedb10ff5b7938d7f90/src/lib/FreeC/Backend/Coq/Syntax.hs#L72).

just95 (changes requested):

src/lib/FreeC/Pass/ExportPass.hs
```diff
@@ -103,3 +113,23 @@ exportInterface modName = do
...
+  qualifyExportedIdent :: EnvEntry -> EnvEntry
+  qualifyExportedIdent entry
+    | isExported entry = if isConEntry entry
+      then entry { entryIdent      = qualifyIdent (entryIdent entry)
+                 , entrySmartIdent = qualifyIdent (entrySmartIdent entry)
+                 }
+      else entry { entryIdent = qualifyIdent (entryIdent entry) }
+    | otherwise = entry
+
+
```
JonasHoefer: Try to avoid redundant/multiple newlines.
```suggestion

```

src/lib/FreeC/Pass/ExportPass.hs
```diff
@@ -103,3 +113,23 @@ exportInterface modName = do
   isExported entry = case entryName entry of
     IR.Qual modName' _ -> modName' == modName
     IR.UnQual _        -> False
+
+  -- Qualifies the Coq identifier in the module interface with the module name
+  -- when an entry of the module is exported.
+  -- This ensures that any module that imports the entry uses its qualified
+  -- name, which prevents name conflicts between imported modules.
```
JonasHoefer: Use Haddock comment also for local functions ([styleguide](https://github.com/FreeProving/guidelines/blob/main/CONTRIBUTING.md#haddock-styleguide)).

src/lib/FreeC/Backend/Coq/Syntax.hs
```diff
@@ -223,3 +224,8 @@ requireImportFrom library modules = ModuleSentence
 requireExportFrom :: ModuleIdent -> [ModuleIdent] -> Sentence
 requireExportFrom library modules = ModuleSentence
   (Require (Just library) (Just Export) (NonEmpty.fromList modules))
+
+-- | Creates a @From ... Require ...@ sentence.
```
JonasHoefer: To symbolize missing code we usually use `…` instead of `...` to explicitly communicate that the ellipses are not part of the languages.
```suggestion
-- | Creates a @From … Require …@ sentence.
```

src/lib/FreeC/Pass/ExportPass.hs
MajaRet: A smart constructor would be nicer, but considering `coqName` is already a `Coq.Ident` and not a string, it would be necessary to convert it to a string first in order to have it converted back to a `Coq.Ident` by the smart constructor... so I think it might be better to just use `Coq.ident` here.

JonasHoefer (changes requested):
Overall the code looks good to me. Just some minor suggestions based on the [styleguide](https://github.com/FreeProving/guidelines/blob/main/CONTRIBUTING.md).

just95 (approved):

JonasHoefer (approved):

PR #88, opened 2020-06-23
 Specification for lifted IR  #81

### LINKED ISSUE #81: Specification for lifted IR
To implement the lifted IR we have to decide, which responsibilities of the back ends should be moved to the earlier compiler stage. Based in these we can decide which language constructs the IR has to be able to represent. This issue should be used to document the responsibilities of the lifted IR.

JonasHoefer:
### Issue

Closes #81

### Description of the Change

Adds data types for expressions and types in the lifted IR.

### DIFF
--- a/src/lib/FreeC/LiftedIR/Effect.hs
+++ b/src/lib/FreeC/LiftedIR/Effect.hs
@@ -1,0 +1,9 @@
+-- | This module contains a data type for effect constraints.
+
+module FreeC.LiftedIR.Effect where
+
+-- | An effect constraint.
+--
+--   These effects corresponds to type classes constraining the container used by
+--   the @Free@ monad. Currently just partiality is supported.
+data Effect = Partiality deriving (Show, Eq)
--- a/src/lib/FreeC/LiftedIR/Syntax/Expr.hs
+++ b/src/lib/FreeC/LiftedIR/Syntax/Expr.hs
@@ -1,0 +1,116 @@
+-- | This module contains the definition of expressions for our lifted
+--   intermediate language.
+
+module FreeC.LiftedIR.Syntax.Expr where
+
+import           FreeC.IR.SrcSpan               ( SrcSpan )
+
+import           FreeC.LiftedIR.Effect          ( Effect )
+import           FreeC.LiftedIR.Syntax.Name
+import           FreeC.LiftedIR.Syntax.Type
+
+-- | An expression.
+data Expr
+  = -- | A constructor.
+    Con { exprSrcSpan :: SrcSpan
+        , exprConName :: ConName
+        , exprType    :: Type
+        }
+
+  | -- | A function or local variable.
+    Var { exprSrcSpan :: SrcSpan
+        , exprVarName :: VarName
+        , exprType    :: Type
+        }
+
+  | -- | Function or constructor application.
+    App { exprSrcSpan     :: SrcSpan
+        , exprAppFunc     :: Expr
+        , exprAppTypeArgs :: [Type]   -- ^ Visible type applications.
+        , exprEffects     :: [Effect] -- ^ Effect set.
+        , exprAppArgs     :: [Expr]   -- ^ Applied arguments.
+        , exprType        :: Type
+        }
+
+  | -- | @if@ expression.
+    If { exprSrcSpan :: SrcSpan
+       , ifExprCond  :: Expr
+       , ifExprThen  :: Expr
+       , ifExprElse  :: Expr
+       , exprType    :: Type
+       }
+
+  | -- | @case@ expression.
+    Case { exprSrcSpan       :: SrcSpan
+         , caseExprScrutinee :: Expr
+         , caseExprAlts      :: [Alt]
+         , exprType          :: Type
+         }
+
+  | -- | Error term @undefined@.
+    Undefined { exprSrcSpan :: SrcSpan
+              , exprType    :: Type
+              }
+
+  | -- | Error term @error "<message>"@.
+    ErrorExpr { exprSrcSpan  :: SrcSpan
+              , errorExprMsg :: String
+              , exprType     :: Type
+              }
+
+  | -- | An integer literal.
+    IntLiteral { exprSrcSpan     :: SrcSpan
+               , intLiteralValue :: Integer
+               , exprType        :: Type
+               }
+
+  | -- | A lambda abstraction.
+    Lambda { exprSrcSpan    :: SrcSpan
+           , lambdaExprArgs :: [VarPat]
+           , lambdaEprRhs   :: Expr
+           , exprType       :: Type
+           }
+
+  | -- | The bind operator for the free monad.
+    Bind { exprSrcSpan  :: SrcSpan
+         , exprBindArg  :: Expr -- ^ The left-hand side argument of @>>=@.
+         , exprBindCont :: Expr -- ^ The right-hand side argument of @>>=@.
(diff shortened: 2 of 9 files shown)


src/lib/FreeC/LiftedIR/Effect.hs
```diff
@@ -0,0 +1,9 @@
+-- | This module contains a data type for effect constraints.
+
+module FreeC.LiftedIR.Effect where
+
+-- | An effect constraint.
+--
+--   These effects corresponds to type classes constraining the container used by
+--   the @Free@ monad. Currently just partiality is supported.
+data Effect = Partiality deriving (Show, Eq)
```
nbun, line 9: I think we should try to link the effect constructors of the lifted IR to the effect defined here. Otherwise there is no guarantee that a program contains only the effects listed in its effect set.

src/lib/FreeC/LiftedIR/Syntax/Expr.hs
```diff
@@ -0,0 +1,116 @@
...
+
+  | -- | A lambda abstraction.
+    Lambda { exprSrcSpan    :: SrcSpan
+           , lambdaExprArgs :: [VarPat]
+           , lambdaEprRhs   :: Expr
+           , exprType       :: Type
+           }
+
+  | -- | The bind operator for the free monad.
+    Bind { exprSrcSpan  :: SrcSpan
```
nbun, line 81: What about `pure`?

src/lib/FreeC/LiftedIR/Syntax/Expr.hs
```diff
@@ -0,0 +1,116 @@
...
+               , exprType        :: Type
+               }
+
+  | -- | A lambda abstraction.
+    Lambda { exprSrcSpan    :: SrcSpan
+           , lambdaExprArgs :: [VarPat]
+           , lambdaEprRhs   :: Expr
+           , exprType       :: Type
+           }
+
```
just95, line 73: I just remembered that `pure` is not only needed in the declaration of smart constructors but also around integer literals, lambda expression and (most importantly) strict arguments such as decreasing arguments that are used in positions where lifted values are expected. Thus, we still need a syntactic construct for `pure` :sweat_smile: 

```suggestion

  | -- | The @pure@ constructor of the @Free@ monad.
    Pure { exprSrcSpan  :: SrcSpan
         , exprPureArg  :: Expr -- ^ The value that is lifted into the @Free@ monad.
         , exprType     :: Type
         }

```

just95 (changes requested):

src/lib/FreeC/LiftedIR/Syntax/Expr.hs
```diff
@@ -0,0 +1,116 @@
...
+    Var { exprSrcSpan :: SrcSpan
+        , exprVarName :: VarName
+        , exprType    :: Type
+        }
+
+  | -- | Function or constructor application.
+    App { exprSrcSpan     :: SrcSpan
+        , exprAppFunc     :: Expr
+        , exprAppTypeArgs :: [Type]   -- ^ Visible type applications.
+        , exprEffects     :: [Effect] -- ^ Effect set.
```
nbun, line 30: Is the effect set of an application supposed to be the union of the effects of its arguments? Is there a benefit in having the effect set information here instead of as part of the `Free` type constructor?

nbun (changes requested):

src/lib/FreeC/LiftedIR/Syntax/Expr.hs
just95: See https://github.com/FreeProving/free-compiler/pull/88#discussion_r444691255

src/lib/FreeC/LiftedIR/Syntax/Expr.hs
just95: The `exprEffects` field models the type class instance arguments that have to be passed to effectful functions in Coq and Agda. For example, `head` is defined as follows in Coq.

```haskell
Definition head (Shape : Type) (Pos : Shape -> Type) (P : Partial Shape Pos) {a : Type} … .
```

The `Partial` instance is needed such that `head` can use the `undefined` and `error` functions on its right-hand side.
If `head` is applied, the `Partial` instance `P` has to be passed explicitly.

```haskell
@head Shape Pos P a xs
```

The `exprEffects` corresponds exactly to the argument `P` in this case (comparred to the current implementation in Coq, the order of `exprAppTypeArgs` and `exprEffects` has been swapped in lifted IR such that the effects such `State s` could depend on type arguments in the future).
If there were multiple effects, multiple type class instances would have to be passed. `exprEffects` must be a list instead of a set since the order in which the effects have to be passed around is fixed by the function definition.

In this definition, the effect set of a function declaration is the union of all effect sets of functions that are used on the right-hand side. The effect set of local variables is always empty. For example, even if `xs` contains computations with further effects, they do not belong to `head`'s effect set. An application like `head [1 ? 2]` in Curry, for example, requires the non-determinism effect to be part of the effect set of the caller. However, `head` does not have to know that `Shape` and `Pos` provide a mechanism for non-determinism. It is only interested in the partiality effect that is provided by the same `Shape` and `Pos` in addition to non-determinism.

just95 (approved):

src/lib/FreeC/LiftedIR/Syntax/Expr.hs
nbun: > In this definition, the effect set of a function declaration is the union of all effect sets of functions that are used on the right-hand side. 

Since the effect set is not part of the function's signature, is it stored in the environment or recomputed each time?

just95: It is stored in the environment. There is an [`entryIsPartial`](https://github.com/FreeProving/free-compiler/blob/1ca3159363e039f60c2b5cedb10ff5b7938d7f90/src/lib/FreeC/Environment/Entry.hs#L94) flag for `FuncEntry`s which is set by the [`partialityAnalysisPass`](https://github.com/FreeProving/free-compiler/blob/1ca3159363e039f60c2b5cedb10ff5b7938d7f90/src/lib/FreeC/Pass/PartialityAnalysisPass.hs). In the future the flag would be replaced by a list of `Effect`s and the pass would be generalized to arbitrary effects.

src/lib/FreeC/LiftedIR/Syntax/Expr.hs
```diff
@@ -0,0 +1,122 @@
...
+  { conPatSrcSpan :: SrcSpan
+  , conPatName    :: ConName
+  }
+ deriving (Eq, Show)
+
+-- | A variable pattern used as an argument to a function, lambda abstraction
+--   or constructor pattern.
+--
+--   The variable pattern can optionally have a type signature
+--   or be annotated by a @!@.
```
Daniel-Teut: I assume the strictness annotations are not needed in the lifted IR? If so, the mention in the documentation should probably be removed.

Daniel-Teut (changes requested):

Daniel-Teut (approved):

src/lib/FreeC/LiftedIR/Effect.hs
nbun: Although a guarantee like this would be nice to have, it is probably too involved to implement, especially considering that it is basically an elaborate test of the effect set annotation algorithm.

nbun (approved):

PR #92, opened 2020-06-25
Merge tests for #20 into main

maclement:
Adds tests for the 'ExportPass' to check whether the bug described in #20 is fixed.

### DIFF
--- a/src/test/FreeC/Pass/ExportPassTests.hs
+++ b/src/test/FreeC/Pass/ExportPassTests.hs
@@ -23,21 +23,10 @@ import           FreeC.Pretty
 import           FreeC.Test.Environment
 import           FreeC.Test.Parser
 
--- | Looks up an exported entry in the given module interface.
---   also checks if the entry is defined in the given module.
---   Returns 'Nothing' if either the check or the lookup fails.
-lookupExportedEntry :: IR.Scope -> IR.QName -> ModuleInterface -> Maybe EnvEntry
-lookupExportedEntry scope qname moduleInterface =
-  if (scope, qname) `elem` Set.toList (interfaceExports moduleInterface)
-    then find ((qname ==) . entryName)
-              (Set.toList $ interfaceEntries moduleInterface)
-    else Nothing
-
 -- | Looks up an exported entry in the given module interface
 --   but does not check if the entry is defined in the given module.
-lookupExportedEntry'
-  :: IR.Scope -> IR.QName -> ModuleInterface -> Maybe EnvEntry
-lookupExportedEntry' scope qname moduleInterface = find
+lookupExportedEntry :: IR.Scope -> IR.QName -> ModuleInterface -> Maybe EnvEntry
+lookupExportedEntry scope qname moduleInterface = find
   (((scope, qname) ==) . entryScopedName)
   (Set.toList $ interfaceEntries moduleInterface)
 
@@ -68,9 +57,10 @@ testExportPass = describe "FreeC.Pass.ExportPass" $ do
         _       <- defineTestCon "A.Bar" 0 "A.Foo"
         _       <- exportPass input
         mOutput <- inEnv $ lookupAvailableModule "A"
-        let foo = fromJust $ lookupExportedEntry IR.TypeScope
-                                                 (Qual "A" (Ident "Foo"))
-                                                 (fromJust mOutput)
+        let foo = fromJust $ lookupExportedEntry
+              IR.TypeScope
+              (IR.Qual "A" (IR.Ident "Foo"))
+              (fromJust mOutput)
         entryIdent foo `shouldBeQualifiedWith` "A"
     it "qualifies Coq identifiers of constructor entries" $ do
       input <- expectParseTestModule ["module A where", "data A.Bar = A.Foo"]
@@ -79,9 +69,10 @@ testExportPass = describe "FreeC.Pass.ExportPass" $ do
         _       <- defineTestCon "A.Foo" 0 "A.Bar"
         _       <- exportPass input
         mOutput <- inEnv $ lookupAvailableModule "A"
-        let foo = fromJust $ lookupExportedEntry IR.ValueScope
-                                                 (Qual "A" (Ident "Foo"))
-                                                 (fromJust mOutput)
+        let foo = fromJust $ lookupExportedEntry
+              IR.ValueScope
+              (IR.Qual "A" (IR.Ident "Foo"))
+              (fromJust mOutput)
         entryIdent foo `shouldBeQualifiedWith` "A"
     it "qualifies Coq identifiers of type synonym entries" $ do
       input <- expectParseTestModule
@@ -92,9 +83,10 @@ testExportPass = describe "FreeC.Pass.ExportPass" $ do
         _       <- defineTestTypeSyn "A.Foo" [] "A.Bar"
         _       <- exportPass input
         mOutput <- inEnv $ lookupAvailableModule "A"
-        let foo = fromJust $ lookupExportedEntry IR.TypeScope
-                                                 (Qual "A" (Ident "Foo"))
-                                                 (fromJust mOutput)
+        let foo = fromJust $ lookupExportedEntry
+              IR.TypeScope
+              (IR.Qual "A" (IR.Ident "Foo"))
+              (fromJust mOutput)
         entryIdent foo `shouldBeQualifiedWith` "A"
     it "qualifies Coq identifiers of function declaration entries" $ do
       input <- expectParseTestModule
@@ -105,9 +97,10 @@ testExportPass = describe "FreeC.Pass.ExportPass" $ do
         _       <- defineTestFunc "A.mkFoo" 0 "A.Foo"
         _       <- exportPass input
         mOutput <- inEnv $ lookupAvailableModule "A"
-        let mkFoo = fromJust $ lookupExportedEntry IR.ValueScope
-                                                   (Qual "A" (Ident "mkFoo"))
-                                                   (fromJust mOutput)
+        let mkFoo = fromJust $ lookupExportedEntry
+              IR.ValueScope
+              (IR.Qual "A" (IR.Ident "mkFoo"))
+              (fromJust mOutput)
         entryIdent mkFoo `shouldBeQualifiedWith` "A"
     it "does not override qualification of Coq identifiers for entries" $ do
       _     <- expectParseTestModule ["module A where", "data A.Foo = A.Foo"]
@@ -120,5 +113,7 @@ testExportPass = describe "FreeC.Pass.ExportPass" $ do
         modifyEnv $ modifyEntryIdent IR.TypeScope name qualid
         _       <- exportPass input
         mOutput <- inEnv $ lookupAvailableModule "B"
-        let foo = fromJust $ lookupExportedEntry' name (fromJust mOutput)
+        let
+          foo =
+            fromJust $ lookupExportedEntry IR.TypeScope name (fromJust mOutput)
         entryIdent foo `shouldBeQualifiedWith` "A"
--- a/src/test/FreeC/Pass/ExportPassTests.hs
+++ b/src/test/FreeC/Pass/ExportPassTests.hs
@@ -7,9 +7,8 @@ where
 
 import           Test.Hspec
 
-import           Data.Maybe                     ( fromJust
-                                                , listToMaybe
-                                                )
+import           Data.List                      ( find )
+import           Data.Maybe                     ( fromJust )
 import qualified Data.Set                      as Set
 import qualified Data.Text                     as Text
 
--- a/src/test/FreeC/Pass/ExportPassTests.hs
+++ b/src/test/FreeC/Pass/ExportPassTests.hs
@@ -116,11 +116,8 @@ testExportPass = describe "FreeC.Pass.ExportPass" $ do
         _    <- defineTestTypeCon "A.Foo" 0 ["A.Foo"]
         _    <- defineTestCon "A.Foo" 0 "A.Foo"
         name <- parseTestQName "A.Foo"
-        let qualifier = Coq.ident "A"
-            textName  = Coq.ident "Foo"
-        modifyEnv $ modifyEntryIdent IR.TypeScope
-                                     name
-                                     (Coq.Qualified qualifier textName)
+        let qualid = Coq.Qualified (Coq.ident "A") (Coq.ident "Foo")
+        modifyEnv $ modifyEntryIdent IR.TypeScope name qualid
         _       <- exportPass input
         mOutput <- inEnv $ lookupAvailableModule "B"
         let foo = fromJust $ lookupExportedEntry' name (fromJust mOutput)
--- a/src/test/FreeC/Pass/ExportPassTests.hs
+++ b/src/test/FreeC/Pass/ExportPassTests.hs
@@ -121,6 +121,5 @@ testExportPass = describe "FreeC.Pass.ExportPass" $ do
                                      (Coq.Qualified qualifier textName)
         _       <- exportPass input
         mOutput <- inEnv $ lookupAvailableModule "B"
-        let foo = fromJust $ lookupExportedEntry' (Qual "A" (Ident "Foo"))
-                                                  (fromJust mOutput)
+        let foo = fromJust $ lookupExportedEntry' name (fromJust mOutput)
         entryIdent foo `shouldBeQualifiedWith` "A"
--- a/src/test/FreeC/Pass/ExportPassTests.hs
+++ b/src/test/FreeC/Pass/ExportPassTests.hs
@@ -25,14 +25,17 @@ import           FreeC.Pretty
 import           FreeC.Test.Environment
 import           FreeC.Test.Parser
 
-
+-- | looks up an exported entry in the given module interface
 lookupExportedEntry :: IR.Scope -> IR.QName -> ModuleInterface -> Maybe EnvEntry
 lookupExportedEntry scope qname moduleInterface =
-  if (scope, qname) `elem` (Set.toList $ interfaceExports moduleInterface)
-    then listToMaybe $ filter ((qname ==) . entryName)
-                              (Set.toList $ interfaceEntries moduleInterface)
+  if (scope, qname) `elem` Set.toList (interfaceExports moduleInterface)
+    then find ((qname ==) . entryName)
+              (Set.toList $ interfaceEntries moduleInterface)
     else Nothing
 
+-- | checks if the given 'Coq.Qualid' is qualified and compares the qualifier
+--   to the given module name.
+--   The check succeeds if the qualifier and module name are the same.
 shouldBeQualifiedWith :: Coq.Qualid -> String -> Converter Expectation
 shouldBeQualifiedWith qualid modNameStr = do
   case qualid of
--- a/src/test/FreeC/Pass/ExportPassTests.hs
+++ b/src/test/FreeC/Pass/ExportPassTests.hs
@@ -10,7 +10,6 @@ import           Test.Hspec
 import           Data.List                      ( find )
 import           Data.Maybe                     ( fromJust )
 import qualified Data.Set                      as Set
-import qualified Data.Text                     as Text
 
 import qualified FreeC.Backend.Coq.Syntax      as Coq
 import           FreeC.Environment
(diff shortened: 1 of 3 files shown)


src/test/FreeC/PipelineTests.hs
```diff
@@ -10,11 +10,13 @@ import           Test.Hspec
 
 import           FreeC.Pass.CompletePatternPassTests
 import           FreeC.Pass.EtaConversionPassTests
+import           FreeC.Pass.ExportPassTests
 import           FreeC.Pass.KindCheckPassTests
 import           FreeC.Pass.PartialityAnalysisPassTests
 import           FreeC.Pass.ResolverPassTests
 import           FreeC.Pass.TypeInferencePassTests
 
+
```
MajaRet: There's an extra white line here.
```suggestion
```

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,94 @@
...
+        _       <- defineTestTypeCon "A.Bar" 0 ["Foo"]
+        _       <- defineTestCon "A.Foo" 0 "A.Bar"
+        _       <- exportPass input
+        mOutput <- inEnv $ lookupAvailableModule "A"
+        let
+          cEntries =
+            filter isConEntry $ Set.toList $ interfaceEntries $ fromJust mOutput
+          foo = head cEntries
+        entryIdent foo `shouldBeQualifiedWith` "A"
+    it "Typesynonyms are qualified when exported" $ do
```
MajaRet:
```suggestion
    it "qualifies the Coq identifiers of exported type synonyms" $ do
```

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,94 @@
...
+        $    prettyString "Expected qualifier"
+        <$$> line
+        <>   indent 2 (prettyString modName)
+        <>   line
+        <$$> prettyString "but it was not qualified"
+
+testExportPass :: SpecWith ()
+testExportPass = describe "FreeC.Pass.ExportPass" $ do
+  context "Exported entries are qualified when exported " $ do
+    it "Names of data types should be qualified when exported" $ do
```
just95: The description that is given to `it` should read like a sentence that describes what *it* (i.e., the `ExportPass`) is expected to do.

```suggestion
    it "qualifies Coq identifiers of data type entries" $ do
```

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,94 @@
...
+        mOutput <- inEnv $ lookupAvailableModule "A"
+        let dEntries =
+              filter isDataEntry $ Set.toList $ interfaceEntries $ fromJust
+                mOutput
+            foo = head dEntries
+        entryIdent foo `shouldBeQualifiedWith` "A"
+    it "Names of constructors are qualified when exported" $ do
+      input <- expectParseTestModule ["module A where", "data Bar = Foo"]
+      shouldSucceedWith $ do
+        _       <- defineTestTypeCon "A.Bar" 0 ["Foo"]
```
just95:
```suggestion
        _       <- defineTestTypeCon "A.Bar" 0 ["A.Foo"]
```

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,94 @@
...
+        _       <- defineTestCon "A.Foo" 0 "A.Bar"
+        _       <- exportPass input
+        mOutput <- inEnv $ lookupAvailableModule "A"
+        let
+          cEntries =
+            filter isConEntry $ Set.toList $ interfaceEntries $ fromJust mOutput
+          foo = head cEntries
```
just95: It could be useful to add a utility function with the following type.

```haskell
lookupExportedEntry :: IR.Scope -> IR.QName -> ModuleInterface -> EnvEntry
```

Such a function would make more explicit which entry you are interested in and reduces boilerplate code in the test cases.

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,94 @@
...
+        <$$> line
+        <>   indent 2 (prettyString modName)
+        <>   line
+        <$$> prettyString "but it was not qualified"
+
+testExportPass :: SpecWith ()
+testExportPass = describe "FreeC.Pass.ExportPass" $ do
+  context "Exported entries are qualified when exported " $ do
+    it "Names of data types should be qualified when exported" $ do
+      input <- expectParseTestModule ["module A where", "data Foo = Bar"]
```
just95: An input module of this form would not cause the environment to contain qualified names. Thus, the declarations should be qualified in this module, too.

```suggestion
      input <- expectParseTestModule ["module A where", "data A.Foo = A.Bar"]
```

In theory, it would also be possible to pass an empty module to the `exportPass`.

```suggestion
      input <- expectParseTestModule ["module A where"]
```

I'm not sure which option I like better. The nice thing about the latter option is that it sets the additional expectation that all information other than the module name is extracted from the environment. However, the former provides additional documentation and reflects the actual state of the module better.

just95 (changes requested):
I think it would be good to add another test case that ensures that entries imported from other modules (e.g. `B.Foo`) are not qualified with the new module.

src/test/FreeC/Pass/ExportPassTests.hs
MajaRet:
```suggestion
    it "qualifies the Coq identifiers of exported data types" $ do
```

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,94 @@
...
+        _       <- defineTestCon "A.Bar" 0 "A.Bar"
+        _       <- defineTestTypeSyn "A.Foo" [] "A.Bar"
+        _       <- exportPass input
+        mOutput <- inEnv $ lookupAvailableModule "A"
+        let tsEntries =
+              filter isTypeSynEntry $ Set.toList $ interfaceEntries $ fromJust
+                mOutput
+            foo = head tsEntries
+        entryIdent foo `shouldBeQualifiedWith` "A"
+    it "Function definitions are qualified when exported" $ do
```
MajaRet:
```suggestion
    it "qualifies the Coq identifiers of exported function definitions" $ do
```

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,94 @@
...
+        _       <- defineTestTypeCon "A.Foo" 0 ["A.Bar"]
+        _       <- defineTestCon "A.Bar" 0 "A.Foo"
+        _       <- exportPass input
+        mOutput <- inEnv $ lookupAvailableModule "A"
+        let dEntries =
+              filter isDataEntry $ Set.toList $ interfaceEntries $ fromJust
+                mOutput
+            foo = head dEntries
+        entryIdent foo `shouldBeQualifiedWith` "A"
+    it "Names of constructors are qualified when exported" $ do
```
MajaRet:
```suggestion
    it "qualifies the Coq identifiers of exported constructors" $ do
```

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,94 @@
...
+import           FreeC.Environment.ModuleInterface
+import           FreeC.Monad.Class.Testable
+import           FreeC.Monad.Converter
+import           FreeC.Pass.ExportPass
+import           FreeC.Pretty
+import           FreeC.Test.Environment
+import           FreeC.Test.Parser
+
+
+shouldBeQualifiedWith :: Coq.Qualid -> String -> Converter Expectation
```
MajaRet, line 35: I think there should also be Haddock documentation for the functions in this module.

MajaRet (changes requested):
Overall, I'm quite happy with the tests. But I do agree that it would be a good idea to add a test case that ensures that identifiers that should not be changed (mainly imported/already qualified names) are not changed.

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,117 @@
...
+        _       <- defineTestCon "A.Foo" 0 "A.Foo"
+        _       <- defineTestTypeCon "B.Foo" 0 ["B.Foo"]
+        _       <- defineTestCon "B.Foo" 0 "B.Foo"
+        _       <- exportPass input
+        _       <- exportPass input'
+        mOutput <- inEnv $ lookupAvailableModule "A"
+        let foo = fromJust $ lookupExportedEntry IR.TypeScope
+                                                 (Qual "A" (Ident "Foo"))
+                                                 (fromJust mOutput)
+        entryIdent foo `shouldBeQualifiedWith` "A"
```
just95: If you looked up `B.Foo` in the module interface of `A` (without doing the `interfaceExports` check in `lookupExportedEntry`), the Coq identifier would be qualified with `A` as well. The first application of the `exportPass` qualifies all entries in the environment with `A` including the entries which are defined in `B`. Thus, the entries of `B.Foo` would have to be added after the first `exportPass`.

```haskell
        _       <- defineTestTypeCon "A.Foo" 0 ["A.Foo"]
        _       <- defineTestCon "A.Foo" 0 "A.Foo"
        _       <- exportPass input
        _       <- defineTestTypeCon "B.Foo" 0 ["B.Foo"]
        _       <- defineTestCon "B.Foo" 0 "B.Foo"
        _       <- exportPass input'
``` 

However, you don't need `data B.Foo = B.Foo` in the first place. You are actually interested in a situation like `module B where import A`. The module interface for `B` should then contain an entry for `A.Foo` whose Coq identifier is still qualified with `A`.

The test described above is most likely to fail, though, because the first `exportPass` does not change the entries in the environment itself. Only the entries in the resulting module interface are qualified, entries in the environment remain unqualified. Thus, the second `exportPass` will not "see" the qualification and qualifies everything again.

I think there are two options to solve this issue.

 1. You could create the module interface for `A` in a [`moduleEnv`](https://github.com/FreeProving/free-compiler/blob/6864d3b5a33b1b1206b355a9b3b62618605d99db/src/lib/FreeC/Monad/Converter.hs#L190) and import the interface afterwards using [`importInterface`](https://github.com/FreeProving/free-compiler/blob/6864d3b5a33b1b1206b355a9b3b62618605d99db/src/lib/FreeC/Pass/ImportPass.hs#L91).
 2. You perform only one `exportPass` and add utility methods to create environment entries with qualified Coq identifiers (or qualify existing Coq identifiers).

src/test/FreeC/PipelineTests.hs
```diff
@@ -24,3 +25,4 @@ testPipeline = do
   testPartialityAnalysisPass
   testResolverPass
   testTypeInferencePass
+  testExportPass
```
just95: I'd put this after `testEtaConversionPass` to keep the same order as in the imports.

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,117 @@
...
+import           FreeC.Environment.ModuleInterface
+import           FreeC.IR.Syntax               as IR
+import           FreeC.Monad.Class.Testable
+import           FreeC.Monad.Converter
+import           FreeC.Pass.ExportPass
+import           FreeC.Pretty
+import           FreeC.Test.Environment
+import           FreeC.Test.Parser
+
+-- | looks up an exported entry in the given module interface
```
just95:
```suggestion
-- | Looks up an exported entry in the given module interface.
```

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,117 @@
...
+lookupExportedEntry :: IR.Scope -> IR.QName -> ModuleInterface -> Maybe EnvEntry
+lookupExportedEntry scope qname moduleInterface =
+  if (scope, qname) `elem` Set.toList (interfaceExports moduleInterface)
+    then find ((qname ==) . entryName)
+              (Set.toList $ interfaceEntries moduleInterface)
+    else Nothing
+
+-- | checks if the given 'Coq.Qualid' is qualified and compares the qualifier
```
just95:
```suggestion
-- | Checks if the given 'Coq.Qualid' is qualified and compares the qualifier
```

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,117 @@
...
+      return
+        $    expectationFailure
+        $    showPretty
+        $    prettyString "Expected qualifier"
+        <$$> line
+        <>   indent 2 (prettyString modNameStr)
+        <>   line
+        <$$> prettyString "but it was not qualified"
+
+testExportPass :: SpecWith ()
```
just95: We usually don't need `SpecWith`. There is a type synonym `Spec` for `SpecWith ()`.

```suggestion
-- | Test group for 'exportPass' tests.
testExportPass :: Spec
```

src/test/FreeC/Pass/ExportPassTests.hs
MajaRet: Comments are full sentences, so they should be capitalized and end with a period.
```suggestion
-- | Looks up an exported entry in the given module interface.
```

src/test/FreeC/Pass/ExportPassTests.hs
MajaRet:
```suggestion
-- | Checks if the given 'Coq.Qualid' is qualified and compares the qualifier
```

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,117 @@
...
+              (Set.toList $ interfaceEntries moduleInterface)
+    else Nothing
+
+-- | checks if the given 'Coq.Qualid' is qualified and compares the qualifier
+--   to the given module name.
+--   The check succeeds if the qualifier and module name are the same.
+shouldBeQualifiedWith :: Coq.Qualid -> String -> Converter Expectation
+shouldBeQualifiedWith qualid modNameStr = do
+  case qualid of
+    Coq.Qualified s _ -> return $ Text.unpack s `shouldBe` modNameStr
```
just95: `Text` has both a `Show` and an `Eq` instance. Thus you can avoid an import of `Text` by writing ``s `shouldBe` Coq.ident modNameStr`` instead.

src/test/FreeC/Pass/ExportPassTests.hs
MajaRet: This function should probably have a short Haddock comment, as well.
```suggestion
-- | Test group for 'exportPass' tests.
testExportPass :: SpecWith ()
```

MajaRet (changes requested):

just95 (changes requested):

src/test/FreeC/Pass/ExportPassTests.hs
just95: See https://github.com/FreeProving/free-compiler/pull/92#discussion_r446505713

src/test/FreeC/Pass/ExportPassTests.hs
just95: See <https://github.com/FreeProving/free-compiler/pull/92#discussion_r446505720>.

MajaRet (approved):

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,126 @@
...
+        name <- parseTestQName "A.Foo"
+        let qualifier = Coq.ident "A"
+            textName  = Coq.ident "Foo"
+        modifyEnv $ modifyEntryIdent IR.TypeScope
+                                     name
+                                     (Coq.Qualified qualifier textName)
+        _       <- exportPass input
+        mOutput <- inEnv $ lookupAvailableModule "B"
+        let foo = fromJust $ lookupExportedEntry' (Qual "A" (Ident "Foo"))
+                                                  (fromJust mOutput)
```
just95: `Qual "A" (Ident "Foo")` is the same as `name`.

```suggestion
        let foo = fromJust $ lookupExportedEntry' name (fromJust mOutput)
```

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,126 @@
...
+      input <- expectParseTestModule ["module B where", "import A"]
+      shouldSucceedWith $ do
+        _    <- defineTestTypeCon "A.Foo" 0 ["A.Foo"]
+        _    <- defineTestCon "A.Foo" 0 "A.Foo"
+        name <- parseTestQName "A.Foo"
+        let qualifier = Coq.ident "A"
+            textName  = Coq.ident "Foo"
+        modifyEnv $ modifyEntryIdent IR.TypeScope
+                                     name
+                                     (Coq.Qualified qualifier textName)
```
just95: You could save a lot of vertical space by defining the entrire `Coq.Qualid` in the `let` expression.

```suggestion
        let qualid = Coq.Qualified (Coq.ident "A") (Coq.ident "Foo")
        modifyEnv $ modifyEntryIdent IR.TypeScope name qualid
```

just95 (changes requested):

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,126 @@
...
+
+import           Data.List                      ( find )
+import           Data.Maybe                     ( fromJust )
+import qualified Data.Set                      as Set
+
+import qualified FreeC.Backend.Coq.Syntax      as Coq
+import           FreeC.Environment
+import           FreeC.Environment.Entry
+import           FreeC.Environment.ModuleInterface
+import           FreeC.IR.Syntax               as IR
```
just95: Oh, and I've just noticed that you were writing `Qual` and `Ident` instead of `IR.Qual` and `IR.Ident`. Always qualify imports of `Syntax` modules.

```suggestion
import qualified FreeC.IR.Syntax               as IR
```

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,126 @@
...
+  if (scope, qname) `elem` Set.toList (interfaceExports moduleInterface)
+    then find ((qname ==) . entryName)
+              (Set.toList $ interfaceEntries moduleInterface)
+    else Nothing
+
+-- | Looks up an exported entry in the given module interface
+--   but does not check if the entry is defined in the given module.
+lookupExportedEntry' :: IR.QName -> ModuleInterface -> Maybe EnvEntry
+lookupExportedEntry' qname moduleInterface =
+  find ((qname ==) . entryName) (Set.toList $ interfaceEntries moduleInterface)
```
just95: You should probably also take the scope into consideration to find the right entry.

```suggestion
lookupExportedEntry'
  :: IR.Scope -> IR.QName -> ModuleInterface -> Maybe EnvEntry
lookupExportedEntry' scope qname moduleInterface = find
  (((scope, qname) ==) . entryScopedName)
  (Set.toList $ interfaceEntries moduleInterface)
```

src/test/FreeC/Pass/ExportPassTests.hs
```diff
@@ -0,0 +1,126 @@
...
+
+-- | Looks up an exported entry in the given module interface.
+--   also checks if the entry is defined in the given module.
+--   Returns 'Nothing' if either the check or the lookup fails.
+lookupExportedEntry :: IR.Scope -> IR.QName -> ModuleInterface -> Maybe EnvEntry
+lookupExportedEntry scope qname moduleInterface =
+  if (scope, qname) `elem` Set.toList (interfaceExports moduleInterface)
+    then find ((qname ==) . entryName)
+              (Set.toList $ interfaceEntries moduleInterface)
+    else Nothing
```
just95: I think you can remove this function in favor of `lookupExportedEntry'`. The check whether the entry is in the `inferfaceExports` is not needed for any test case.

just95 (changes requested):

src/test/FreeC/Pass/ExportPassTests.hs
maclement: I did that in an earlier commit. I somehow overlooked that.

just95 (approved):

PR #94, opened 2020-06-26
Spell check block 9

JonasHoefer:


ichistmeinname (approved):

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithHelpers.hs
```diff
@@ -61,8 +61,8 @@ convertRecFuncDeclsWithHelpers' decls = do
   (helperDecls, mainDecls) <- mapAndUnzipM (uncurry transformRecFuncDecl)
                                            (zip decls decArgs)
   -- Convert helper and main functions.
-  -- The right hand side of the main functions are inlined into helper the
-  -- functions. Because inlining can produce fesh identifiers, we need to
+  -- The right hand sides of the main functions are inlined into the helper
```
just95:
```suggestion
  -- The right-hand sides of the main functions are inlined into the helper
```

There are probably a lot more unhyphenated occurrences of "right-hand" and "left-hand" side. Could you do a *find and replace* on all `.hs` files (also outside block 9)? : )

just95 (changes requested):

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithHelpers.hs
JonasHoefer: Will do.

just95: Thanks : )

just95 (approved):

PR #102, opened 2020-06-29
Add partial alternatives for subterm functions

### LINKED ISSUE #52: Add partial alternatives for subterm functions
## Description

Add functions with the following types to [`FreeC.IR.Subterm`](https://github.com/FreeProving/free-compiler/blob/main/src/lib/FreeC/IR/Subterm.hs) (names are temporary).

```haskell
selectSubterm' :: Subterm a => a -> Pos -> a
replaceSubterm' :: Subterm a => a -> Pos -> a -> a
replaceSubterms' :: Subterm a => a -> [(Pos, a)] -> a
```

## Motivation

The functions for working with subterms in [`FreeC.IR.Subterm`](https://github.com/FreeProving/free-compiler/blob/main/src/lib/FreeC/IR/Subterm.hs) like [`selectSubterm`](https://github.com/FreeProving/free-compiler/blob/1fbc1d584e113745ae6acbb1b35475d73ad398ed/src/lib/FreeC/IR/Subterm.hs#L211), and [`replaceSubterm`](https://github.com/FreeProving/free-compiler/blob/1fbc1d584e113745ae6acbb1b35475d73ad398ed/src/lib/FreeC/IR/Subterm.hs#L223) are lifted to the `Maybe` monad. They return `Nothing` if the given position is invalid. However, it is often known that these functions are applied to valid position only. This leads to a common pattern of applying `fromJust` to the result. For example [`findSubterms`](https://github.com/FreeProving/free-compiler/blob/1fbc1d584e113745ae6acbb1b35475d73ad398ed/src/lib/FreeC/IR/Subterm.hs#L260) operates on [`allPos`](https://github.com/FreeProving/free-compiler/blob/1fbc1d584e113745ae6acbb1b35475d73ad398ed/src/lib/FreeC/IR/Subterm.hs#L175)itions of a term and thus knows that it is safe to unwrap the value returned by `selectSubterm` using `fromJust`. [Other parts of the code](https://github.com/FreeProving/free-compiler/blob/main/src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithHelpers.hs#L151) operate on similar assumptions. It would be nice to have functions that behave like `fromJust . selectSubterm` and `fromJust . replaceSubterm`, respectively, but report a better error message (involving the position that was not found) if the assumption that the position is valid is broken.

## Additional context

For additional context follow [this discussion](https://github.com/FreeProving/free-compiler/pull/39#discussion_r435012987).

Daniel-Teut:
### Issue

Closes #52.

### Description of the Change

Adds partial versions of some subterm functions that throw an error instead of returning `Nothing`. Also replaces some function calls with the alternative version where appropriate.

### DIFF
--- a/src/lib/FreeC/IR/Subterm.hs
+++ b/src/lib/FreeC/IR/Subterm.hs
@@ -39,8 +39,7 @@ import           Data.Composition               ( (.:) )
 import           Data.List                      ( intersperse
                                                 , isPrefixOf
                                                 )
-import           Data.Maybe                     ( fromJust
-                                                , fromMaybe
+import           Data.Maybe                     ( fromMaybe
                                                 , listToMaybe
                                                 )
 import           Data.Map.Strict                ( Map )
@@ -287,13 +286,13 @@ replaceSubterms' = foldl (\term (pos, term') -> replaceSubterm' term pos term')
 --   satisfy the provided predicate.
 findSubtermPos :: Subterm a => (a -> Bool) -> a -> [Pos]
 findSubtermPos predicate term =
-  filter (predicate . fromJust . selectSubterm term) (allPos term)
+  filter (predicate . selectSubterm' term) (allPos term)
 
 -- | Gets a list of subterms of the given expression that satisfy the
 --   provided predicate.
 findSubterms :: Subterm a => (a -> Bool) -> a -> [a]
 findSubterms predicate term =
-  filter predicate (map (fromJust . selectSubterm term) (allPos term))
+  filter predicate (map (selectSubterm' term) (allPos term))
 
 -- | Gets the first subterm of the given expression that satisfies the
 --   provided predicate.
--- a/src/lib/FreeC/IR/Subterm.hs
+++ b/src/lib/FreeC/IR/Subterm.hs
@@ -20,8 +20,11 @@ module FreeC.IR.Subterm
   , rightOf
     -- * Subterms
   , selectSubterm
+  , selectSubterm'
   , replaceSubterm
+  , replaceSubterm'
   , replaceSubterms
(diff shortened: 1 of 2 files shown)


src/lib/FreeC/IR/Subterm.hs
```diff
@@ -66,12 +68,26 @@ nullary :: b -> [a] -> Maybe b
...
+    ++ showPretty term
+    ++ " does not exists."
+
 -------------------------------------------------------------------------------
 -- Direct children                                                           --
 -------------------------------------------------------------------------------
 
 -- | Type class for AST nodes with child nodes of the same type.
-class Subterm a where
+class Pretty a => Subterm a where
```
just95: The documentation for `Subterm` should explain why it is a subclass of `Pretty`.

src/lib/FreeC/IR/Subterm.hs
```diff
@@ -247,6 +274,10 @@ replaceSubterms term ((p, e) : pes) = do
   term' <- replaceSubterm term p e
   replaceSubterms term' pes
 
+-- | Like 'replaceSubterms' but throws an error if any of the subterms could
+--   not be replaced.
+replaceSubterms' :: Subterm a => a -> [(Pos, a)] -> a
+replaceSubterms' = foldl (\term (pos, term') -> replaceSubterm' term pos term')
```
just95, line 281: I like the solution using `foldl`. Could you rewrite `replaceSubterms` using `foldlM` as well? : )

just95 (changes requested):

JonasHoefer (approved):

just95 (approved):

Issue #72, opened 2020-06-17
Add support for heterogenous compiler passes


just95:
## Description

Compiler passes should be able to change the type of processed modules.

## Motivation

Currently a compiler `Pass` is a function of type `a -> Converter a`. The type variable `a` is usually instantiated with `IR.Module`. The idea is now to make passes functions of type `a -> Converter b` . We will refer to passes with different argument and return types as *heterogeneous compiler passes*.

One application of such heterogeneous passes is dependency analysis. Currently, dependency analysis has to be performed both in the passes and in the back end because the information about strongly connected components cannot be preserved in `IR.Module`s. In #71, modules will be parameterized over their content data type. This will allow us to write the `DependencyAnalysisPass` as a heterogeneous pass of type `IR.Module -> Converter (IR.ModuleOf [DependencyComponent IR.TopLevelDecl])`. With this new type, information from the compiler passes can be passed to the back end.

The same technique could be used to encode on type level the assumption that types are annotated after type inference in the future.

## Drawbacks

Since all compiler passes have the same type, they can currently be stored in a list (see [`FreeC.Pipeline`](https://github.com/FreeProving/free-compiler/blob/1b666387fe35fca42652471b6e46c8d3f1305c03/src/lib/FreeC/Pipeline.hs#L25)). This will not be possible in the presence of heterogeneous passes. Some kind of type aligned sequence would have to be used if we still want to represent passes as data. Thus, it is probably best to just combine passes using `(>=>)` for now.

JonasHoefer:
Just as a side note. The definition of a _heterogeneous compiler pass_ is an arrow in the Kleisli category of the `Converter` monad and therefore is equivalent to [`Kleisli`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Control-Arrow.html#v:Kleisli) for `Converter`. If the compiler pipeline becomes more complex (e.g. optional stages based on the choice of backend, splitting and rejoining phases, reusing outputs, ...) we could consider using the `Arrow` instance to avoid writing composition boiler plate code (and maybe write down the pipeline in a more declarative way).

PR #103, opened 2020-06-29
Issue 95

### LINKED ISSUE #95: Translate IR types to lifted IR types
- [x] Implement the IR to lifted IR translation for types. 
- [x] Use the resulting lifted IR AST to generate Agda type signatures using the existing mechanisms from the Agda back end.

JonasHoefer:
### Issue 

Closes #95

### Description of the Change

Adds IR to lifted IR translation for types and refactors the Agda back end to use this new intermediate stage.

### Alternate Designs

Parts of the logic (e.g. choosing which translation to use, finding decreasing arguments) must remain in the back end for the moment. Introducing `Declaration`s for the lifted IR and moving dependency analysis to an early stage in the pipeline (see #72) might allow us to move more logic from the back ends to the _IR to lifted IR_ translation.

### Verification Process

The old Agda back end tests were reused to test the new lifted IR functions.


src/lib/FreeC/Backend/Agda/Converter/Type.hs
```diff
@@ -1,134 +1,89 @@
...
--- | Tests whether the given type is a type application, which applies
---   arguments to the given constructor?
+-- | Translates a given type in lifted IR to Agda.
 --
---   > appliesTo (((C x₁) …) xₙ) C ↦ True
-appliesTo :: IR.Type -> IR.QName -> Bool
-appliesTo (IR.TypeApp _ l _    ) name = l `appliesTo` name
-appliesTo (IR.TypeCon _ conName) name = conName == name
-appliesTo _                      _    = False
+--   If a @Size@ variable is needed, but non is provided an error will be occur.
```
nbun:
```suggestion
--   If a @Size@ variable is needed, but none is provided, an error occurs.
```

src/lib/FreeC/LiftedIR/Converter/Type.hs
```diff
@@ -0,0 +1,94 @@
...
+  , convertType
+  , convertType'
+  )
+where
+
+import qualified FreeC.IR.Syntax               as IR
+import           FreeC.IR.SrcSpan               ( SrcSpan(NoSrcSpan) )
+import qualified FreeC.LiftedIR.Syntax         as LIR
+
+-- | Functions not defined in terms lambdas (@f x = …@ not @f = \x -> …@) aren't
```
nbun:
```suggestion
-- | Functions not defined in terms of lambdas (@f x = …@ not @f = \x -> …@) aren't
```

src/lib/FreeC/LiftedIR/Converter/Type.hs
```diff
@@ -0,0 +1,94 @@
...
+--   > τ₁ -> … -> τₙ -> ρ ↦ τ₁' -> … -> τₙ' -> ρ'
+convertFuncType :: [IR.Type] -> [LIR.Type]
+convertFuncType = map convertType
+
+convertRecFuncType :: Int -> [IR.Type] -> [LIR.Type]
+convertRecFuncType decIndex args =
+  let startArgs = map convertType $ take (decIndex - 1) args
+      decArg    = markOutermostDecreasing $ convertType $ args !! decIndex
+      endArgs   = map convertType $ drop (decIndex + 1) args
+  in  startArgs ++ (decArg : endArgs)
```
nbun:
```suggestion
  let convArgs = map convertType args
      (startArgs, decArg:endArgs) = splitAt (decIndex - 1) convArgs
  in  startArgs ++ (markOutermostDecreasing decArg : endArgs)
```

src/lib/FreeC/LiftedIR/Converter/Type.hs
```diff
@@ -0,0 +1,94 @@
...
+
+convertConArg :: IR.QName -> IR.Type -> LIR.Type
+convertConArg ident = markAllDec ident . convertType
+
+-------------------------------------------------------------------------------
+-- Translations                                                              --
+-------------------------------------------------------------------------------
+
+-- | Converts a type from IR to Lifted IR by lifting it into the @Free@ monad.
```
nbun:
```suggestion
-- | Converts a type from IR to lifted IR by lifting it into the @Free@ monad.
```

src/lib/FreeC/LiftedIR/Converter/Type.hs
```diff
@@ -0,0 +1,94 @@
...
+--
+--   This corresponds to the dagger translation for monotypes as described by
+--   Abel et al.
+--
+--   > τ' = Free τ*
+convertType :: IR.Type -> LIR.Type
+convertType = LIR.FreeTypeCon NoSrcSpan . convertType'
+
+-- | Lifts a type from IR to Lifted IR by renaming type variables and
```
nbun:
```suggestion
-- | Lifts a type from IR to lifted IR by renaming type variables and
```

src/lib/FreeC/LiftedIR/Converter/Type.hs
```diff
@@ -0,0 +1,94 @@
...
+  name
+  (markAllDec decName `map` ts)
+  (name == decName || dec)
+
+-- | Marks the outermost occurring type constructor as decreasing.
+--
+--   Note: Could be generalized to annotate a constructor based on a position.
+--   At the moment this isn't needed, because our termination checker doesn't
+--   cover cases, where the decreasing argument is just part of anther argument.
+--   e.g. Decreasing in the first element of a pair.
```
nbun:
```suggestion
--   At the moment this isn't needed because our termination checker doesn't
--   cover cases where the decreasing argument is part of another argument.
--   For example, a decreasing argument in the first element of a pair is not covered.
```

src/lib/FreeC/LiftedIR/Syntax/Type.hs
```diff
@@ -44,3 +44,23 @@ data Type
       , typeFreeArg :: Type
       }
  deriving (Eq, Show)
+
+-------------------------------------------------------------------------------
+-- Utility                                                                   --
+-------------------------------------------------------------------------------
+
+-- | Applies a type level expression lifted IR expression to a type constructor.
```
nbun: Usually you apply a constructor to a type and not the other way around, if that's what is meant here?

src/lib/FreeC/LiftedIR/Syntax/Type.hs
```diff
@@ -44,3 +44,23 @@ data Type
...
+typeApp :: Type -> Type -> Type
+typeApp (TypeCon srcSpan name ts dec) t = TypeCon srcSpan name (t : ts) dec
+typeApp _ _ = error "Cannot apply type to non type constructor!"
+
+-- | Smart constructor for lifted IR functions.
+func :: Type -> Type -> Type
+func = FuncType NoSrcSpan
+
+-- | Whether the type contains a decreasing argument.
```
nbun:
```suggestion
-- | Decides whether a type contains a decreasing argument.
```

nbun (changes requested):

src/lib/FreeC/LiftedIR/Syntax/Type.hs
```diff
@@ -44,3 +44,23 @@ data Type
       , typeFreeArg :: Type
       }
  deriving (Eq, Show)
+
+-------------------------------------------------------------------------------
+-- Utility                                                                   --
+-------------------------------------------------------------------------------
+
+-- | Applies a lifted IR type constructor to type level expression.
```
nbun:
```suggestion
-- | Applies a lifted IR type constructor to a type level expression.
```

nbun (approved):

src/test/FreeC/Backend/Agda/Converter/TypeDeclTests.hs
```diff
@@ -54,7 +54,7 @@ testConvertDataDecls =
       shouldConvertTypeDeclsTo
         (Recursive ["data List a = Nil | Cons a (List a)"])
         [ "data List (Shape : Set)(Pos : Shape → Set)(a : Set) : {Size} → Set where"
-        ++ "  nil : ∀ {i} → List Shape Pos a {↑ i}"
+        ++ "  nil : List Shape Pos a"
```
just95, line 57: In the `free-proving-agda-code` repository we wrote [`nil : ∀ {i} → List S P A {i}`](https://github.com/FreeProving/free-proving-agda-code/blob/8247549708d3a01a78d0e7afbb79f8f57e4a9bd2/src/Free/List.agda#L12).  Have you tested whether `{↑ i}` works as well?

src/lib/FreeC/LiftedIR/Converter/Type.hs
```diff
@@ -0,0 +1,105 @@
...
+-------------------------------------------------------------------------------
+
+-- | Converts a type from IR to lifted IR by lifting it into the @Free@ monad.
+--
+--   This corresponds to the dagger translation for monotypes as described by
+--   Abel et al.
+--
+--   > τ' = Free τ*
+convertType :: IR.Type -> LIR.Type
+convertType = LIR.FreeTypeCon NoSrcSpan . convertType'
```
just95: In order to distinguish functions that convert IR to lifted IR from the functions that convert lifted IR to the target language, I think that we should give them different names. Especially since the `convert` prefix suggests that a function is in the `Converter` moand, I would use that prefix in this case. I'd recommend the following naming scheme.

 - `lift*` for the translation from IR to lifted IR.
 - `convertLifted*` for the translation from lifted IR to the intermediate language.
 - `convert*` for the translation from the IR to the target language (i.e., `convertLifted* . lift*`).

This will also make the qualified import of the IR to lifted IR conversion functions unnecesarry.

src/lib/FreeC/LiftedIR/Syntax/Type.hs
```diff
@@ -44,3 +44,23 @@ data Type
...
  deriving (Eq, Show)
+
+-------------------------------------------------------------------------------
+-- Utility                                                                   --
+-------------------------------------------------------------------------------
+
+-- | Applies a lifted IR type constructor to a type level expression.
+typeApp :: Type -> Type -> Type
+typeApp (TypeCon srcSpan name ts dec) t = TypeCon srcSpan name (t : ts) dec
+typeApp _ _ = error "Cannot apply type to non type constructor!"
```
just95: The problem I've got with this function is that `typeApp (D t₁ … tₙ) t₀ = D t₀ t₁ … tₙ`. That's why `LIR.convertType'` reverses the argument list and why you have to apply `reverse` in `Agda.convertType`. The reversing of arguments is a potential source of errors and not documented anywhere. I think that `LIR.convertType'` should be implemented with an accumulator internally such that the order of arguments is preserved.

src/lib/FreeC/LiftedIR/Syntax/Type.hs
```diff
@@ -44,3 +44,23 @@ data Type
...
+
+-- | Smart constructor for lifted IR functions.
+func :: Type -> Type -> Type
+func = FuncType NoSrcSpan
```
just95: I don't like smart constructors that just drop the `SrcSpan` argument. They encourage the user to drop location information. In the IR there is for example the following smart constructor which does exactly what you use `LIR.func` for, but does force the user to think about relevant location information (or actively decide that there is no `SrcSpan` than can be inserted).

```haskell
-- | Creates a function type with the given argument and return types.
funcType :: SrcSpan -> [Type] -> Type -> Type
funcType srcSpan = flip (foldr (FuncType srcSpan))
```

Also note that smart constructor names should match the constructor names.

just95 (changes requested):

src/test/FreeC/Backend/Agda/Converter/TypeDeclTests.hs
JonasHoefer: Yes, i did some test with the old Ada backend. This new implementation drops the whole annotation if it isn't needed. I tested multiple data types from the `free-proving-agda-code` repository and have not found any problems.
> The old Agda back end annotated all constructor with a Size annotation,
> because it was convenient to do so. The new back end places annotations
> only if needed. This change does not effect the behavior in Agda.

src/lib/FreeC/LiftedIR/Converter/Type.hs
```diff
@@ -0,0 +1,105 @@
...
+
+-- | Lifts a type from IR to lifted IR by renaming type variables and
```
Daniel-Teut: As far as I can see `convertType'` does not actually rename any type variables or constructors.

Daniel-Teut (changes requested):

src/lib/FreeC/LiftedIR/Converter/Type.hs
```diff
@@ -0,0 +1,105 @@
...
+--   > (τ₁ τ₂)* = τ₁* τ₂*
+--   > (τ₁ -> τ₂)* = τ₁' -> τ₂'
+--   > C* = Ĉ Shape Position
+--   > α* = α̂
+liftType' :: IR.Type -> LIR.Type
+liftType' (IR.TypeCon srcSpan name) = LIR.TypeCon srcSpan name Nil False
+liftType' (IR.TypeVar srcSpan name) = LIR.TypeVar srcSpan name
+liftType' (IR.TypeApp _ l r       ) = liftType' l `LIR.typeApp` liftType' r
+liftType' (IR.FuncType srcSpan l r) =
+  LIR.FuncType srcSpan (liftType l) (liftType r)
```
just95: I don't think that using a `SnocList` for `typeConArgs` really solves the underlying problem that I've pointed out in my [previous comment](https://github.com/FreeProving/free-compiler/pull/103#discussion_r448538273). On the upside, `SnocList` makes explicit that the list is in reverse order. On the downside, the list is effectively still stored in the wrong order.

Consider the following grammar for the lifted IR, for example.

```haskell
Type ::= ConIdent { Type } | … 
```

Parsers and pretty printers for this grammar still have to reverse the arguments `{ Type }`. That's odd but may be justified if it simplified operations on the lifted IR. 

From the point of view of `convertLiftedType`, it does not matter whether the arguments are stored as a list or `SnocList`. Only `Foldable` and `Traversable` instances are needed. However, since those instances are not tail recursive in case of `SnocList`, performance would be better when `typeConArgs` were stored as a list. 

The only code that really benifits from a `SnocList` is `liftType'`. However, `liftType'` can easily be written in terms of a helper function that accumulates the arguments.

```suggestion
liftType' :: IR.Type -> LIR.Type
liftType' = flip liftTypeApp' []

-- | Like 'liftType'' but accumulates the type arguments in a second parameter.
--
--   > (C τ₁ … τₙ)* = C τ₁* … τₙ*
--   > (τ₁ -> τ₂)* = τ₁' -> τ₂'
--   > α* = α
liftTypeApp' :: IR.Type -> [IR.Type] -> LIR.Type
liftTypeApp' (IR.TypeCon srcSpan name) ts =
  LIR.TypeCon srcSpan name (map liftType' ts) False
liftTypeApp' (IR.TypeVar srcSpan name) [] = LIR.TypeVar srcSpan name
liftTypeApp' (IR.TypeApp _ l r       ) ts = liftTypeApp' l (r : ts)
liftTypeApp' (IR.FuncType srcSpan l r) [] =
  LIR.FuncType srcSpan (liftType l) (liftType r)
liftTypeApp' _ (_ : _) = error "liftTypeApp': Only type constructors can be applied!"
```

Compared to the `SnocList` approach, much less code has to be written (`typeApp` can even be deleted), the argument list does not have to be reversed in any of our use cases and the code in `convertLiftedType` gets slightly more performant.

src/lib/FreeC/Backend/Agda/Converter/FuncDecl.hs
```diff
@@ -13,15 +13,14 @@ import           FreeC.Backend.Agda.Converter.Arg
...
-                                                , convertRecFuncType
-                                                )
+                                                ( convertLiftedFuncType )
 import qualified FreeC.Backend.Agda.Syntax     as Agda
 import           FreeC.Backend.Coq.Analysis.DecreasingArguments
                                                 ( identifyDecArgs )
 import           FreeC.Environment.LookupOrFail
 import           FreeC.IR.DependencyGraph
 import qualified FreeC.IR.Syntax               as IR
+import qualified FreeC.LiftedIR.Converter.Type as LIR
```
just95: After renaming the lifting functions, it is not necessary anymore to qualify imports of `FreeC.LiftedIR.Converter.*`.

```suggestion
import           FreeC.LiftedIR.Converter.Type
```

just95 (changes requested):

Daniel-Teut (approved):

src/lib/FreeC/LiftedIR/Converter/Type.hs
JonasHoefer: My problem with this approach is approach is that it also doesn't solve the problem and just moves it. Storing the type parameters in the 'correct' order forces us to use `foldl` to build the AST in the backend, therefore just reversing the list again. Even worse, the IR already has the right structure! This approach just reverses the structure twice while passing it through the pipeline.
```
     ⋰                                        ⋰
    @                         :              $  
   / \                       / \            / \ 
  @   τ₂   =>  IR.TypeApp C τ₁  :    =>    $   τ₂  
 / \                           / \        / \   
C   τ₁                        τ₂  …      C   τ₁ 
```
Storing the list in 'reverse' order (or as a snoc list to just visually flip it) doesn't change the structure, allowing us to use the right recursion scheme to build the AST in the back end (`foldr` for `[LIR.Type]` or `foldl` for `SnocList LIR.Type`), building the AST without reversing the list. 

Since parsing and pretty printing are left to right operations we have to deal with the problem of reversing the list at some point. I think we should just solve the problem as early as possible to avoid influencing the main flow of data (i.e. reversing all type applications twice while compiling).

Dealing with left recursion while parsing shouldn't be a problem. We can either parse the structure right recursive and reverse it later (`reverse` or `foldl`) or use [`chainl`](https://hackage.haskell.org/package/parsec-3.1.14.0/docs/src/Text.Parsec.Combinator.html#chainl), which avoids unnecessary backtracking by parsing bottom up. Note that dealing with the problem while parsing avoids the `foldl` in the backend.

While pretty printing we of course have to deal with the problem again. One way of reducing the thunk buildup of `foldl` is to use [`foldl'`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-List.html#v:foldl-39-) instead.

---

From a performance point of view (which shouldn't really matter, because our list has like 5 items :smile:) it's just a trade of between reversing during IO or between pipeline stages. 

From an elegance/readability point of view I'm not quite sure which approach is better. I think I lean towards the simpler algorithms working on the expected structure. As a side note: I already started working on the conversion of expressions. The Coq backend uses the same technique of accumulating arguments which I reused, because in this case it simplifies the algorithm.

just95: > performance [...] shouldn't really matter, because our list has like 5 items

I agree. Performance is not really a priority for us. But why do we then care about the implicit reversals performed by `foldl` in the back end? The point I wanted to make with my performance argument is that the `SnocList` approach complicates the code but doesn't even improve performance.

> Since parsing and pretty printing are left to right operations we have to deal with the problem of reversing the list at some point.

I've included the parsing and pretty printing example to show that even though `liftType'` can be defined slightly more compactly with `SnocList`s, other code will become more complex. For parsing a list of arguments, we can simply use `Parsec.many` and `hsep . map pretty` for pretty printing. I guess that much more code is needed to parse or pretty print a `SnocList`.

> The Coq backend uses the same technique of accumulating arguments

That's only partially true. Yes, I don't use an accumulator argument in [`convertType'`](https://github.com/FreeProving/free-compiler/blob/35bbb90b0c5d776489dae00539b2b55acbfb0ae1/src/lib/FreeC/Backend/Coq/Converter/Type.hs#L44). However, unlike `Agda.typeApp`, [`Coq.app`](https://github.com/FreeProving/free-compiler/blob/35bbb90b0c5d776489dae00539b2b55acbfb0ae1/src/lib/FreeC/Backend/Coq/Syntax.hs#L92) adds the type argument to the back of the list and not to the front. If I recommended to use `(++)` instead of `(:)` in `Agda.typeApp`, I'd most likely have started a discussion about performance. That's one reason why I've suggested to use an accumulator argument. The other reason is that [`convertExpr'`](https://github.com/FreeProving/free-compiler/blob/35bbb90b0c5d776489dae00539b2b55acbfb0ae1/src/lib/FreeC/Backend/Coq/Converter/Expr.hs#L41) requires an accumulator argument. Therefore, adding an accumulator argument to `liftType'` improves consistency.

just95 (approved):

PR #105, opened 2020-06-30
Complete spell check of block 5

stu204146-kiel:
### Issue

Is linked to Branch Spell check block 6

### Description of the Change

All files in block 5 were spell checked

### DIFF
--- a/free-compiler.cabal
+++ b/free-compiler.cabal
@@ -158,9 +158,9 @@ library freec-internal
                       , FreeC.IR.Syntax.Pragma
                       , FreeC.IR.Syntax.Type
                       , FreeC.IR.Syntax.TypeDecl
-                      , FreeC.IR.Syntax.TypeSchema
+                      , FreeC.IR.Syntax.TypeScheme
                       , FreeC.IR.Syntax.TypeVarDecl
-                      , FreeC.IR.TypeSchema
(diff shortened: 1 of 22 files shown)


nbun (approved):
It seems to me that both "schemas" and "schemata" are correct plural forms of "schema". As long as the usage is consistent, either one should be fine.

just95 (changes requested):
I agree with @nbun that it is okay to write "schemata" instead of "schemas" as long as we do so consitently. However, since I wrote "schemas" everywhere else, this PR currently introduces inconsistencies. Thus, all other occurrence of "schemas" would have to be replaced as well. I took a look at the [original publication by Damas and Milner](https://web.cs.wpi.edu/~cs4536/c12/milner-damas_principal_types.pdf) to check which plural form they use. As it turns out, they do not call it a "schama" but a "scheme" (pl. "schemes"). Thus, I'd suggest to replace "schema" by "scheme" everywhere (including the names of the data type, constructor and variables).

nbun:
> Thus, I'd suggest to replace "schema" by "scheme" everywhere (including the names of the data type, constructor and variables).

@stu204146-kiel Have you considered this suggestion?

just95 (approved):

PR #108, opened 2020-07-03
Implement a rudimentary sharing operator in Coq

### LINKED ISSUE #77: Implement a first version of the share operator in Coq
## Description

A first version of the [`share` operator](https://bunkenburg.net/papers/ModelingCallTimeChoiceAsEffect.pdf) using explicit scope tags should be implemented in Coq.

The `share` operator associates multiple occurrences of a shared variable with the same identifier so that an effect handler can recognize that two occurrences of a variable (with a possibly effectful value) refer to the same result, and handle them accordingly.

Because `share` depends on an effect stack that, unlike in Haskell, needs to be fixed (and therefore effect-specific) in Coq, more work is needed to actually define the `share` operator correctly. As a consequence, this version of `share` is not intended to be functional yet.

## Motivation

In order to implement sharing in the presence of effects such as non-determinism, sharing must be modeled explicitly. 
The compiler should eventually be able to recognize shared variables and enforce that their values are shared correctly in the resulting Coq code in the presence of various effects.

MajaRet:
### Issue

Closes #77

### Description of the Change

A non-functional, effect-generic sharing operator (as well as a strict version of it) was implemented in Coq. 
The operator simply takes a `Free` value and wraps it in an additional `Free` layer via `pure`. 
The strict version evaluates the argument first via `>>=` and then wraps it in two layers of `Free` with two applications of `pure`. 

### Alternate Designs

Right now, the operator operates on `Free` values with arbitrary effects. To get a functional `share` operator, the effect stack will be required to include effects for state and scoped sharing.


nbun (changes requested):
The sharing operators should have names that reflect their semantics, that is, they should include cbn and cbv.

nbun (approved):

maclement (approved):

ichistmeinname (approved):

Issue #110, opened 2020-07-03
Investigate the possibility of implementing the type class :<: in Coq


MajaRet:
## Description

It should be investigated whether it is possible to define the [`:<:` type class (p. 26)](https://bunkenburg.net/papers/ModelingCallTimeChoiceAsEffect.pdf)  in Coq.

## Motivation

The Haskell version of the share operator uses a type class `:<:` that consists of a function `inj` to dynamically inject an effect into a combination of effects containing that effect. 
This way, we don't have to rely on the concrete structure of the effect stack when we define functions that require a certain effect  (e.g. the `share` operator, whose argument is required to contain the `Share` effect). 

## Alternatives
An alternative to using `:<:` would be to use a fixed effect stack so the placement of an effect within a combination of effects can be done explicitly. 
However, this will likely lead to a significant loss of flexibility.

## ToDo

- [x] Implement a type class `Injectable` in Coq that represents that a given effect is contained in another given effect (stack).
- [x] Define useful instances of the type class.
- [ ] Add tests to ensure that the correct instance is consistently chosen in situations where more than one instance fits.

PR #111, opened 2020-07-03
Implement Share and Choice effects

### LINKED ISSUE #109: Implement effect combinator and effects for sharing and non-determinism in Coq
## Description

The following effects should be implemented in Coq:
- A `Comb` effect representing the combination of two effects.
- A `Share` effect representing sharing using a state and explicit scoping tags.
- An `ND` effect representing a non-deterministic program.

## Motivation

In order to model sharing in Coq, we must be able to mark sharing scopes and assign IDs to shared expressions so that a handler can correctly identify shared expressions and handle them accordingly. For this, we need a `Share` effect that contains a state and tags to open and close scopes.

Because we need to combine the `Share` effect with the other effect(s) of the program, we also need a combinator `Comb` for effects. 

The `ND` effect models a non-deterministic program. Non-determinism is an effect where sharing (call-time choice) has to be modeled explicitly, so this effect is useful to test our implementation of sharing.

MajaRet:
### Issue

Closes #109 
Depends on #113

### Description of the Change

The effects `Share` for sharing using explicit scope tags, and `Choice` for non-deterministic programs have been implemented in Coq.

Smart constructors using the new `Injectable` type class have also been added to `Identity` and `Maybe`.

### DIFF
--- a/base/Free/Injectable.v
+++ b/base/Free/Injectable.v
@@ -62,10 +62,9 @@ Fixpoint embed {A : Type} {Shape : Type} {Pos : Shape -> Type} (Shape' : Type)
  | impure s pf => impure (injS s) (fun p => embed Shape' Pos' (pf (injP p)))
  end.
 
-(* TODO: Give Shape and Pos before Shape' and Pos' *)
-(* Partial instance for an embedded effect *)
-Instance PartialLifted (Shape' : Type) (Pos' : Shape' -> Type) 
-  (Shape : Type) (Pos : Shape -> Type) `{Injectable Shape Pos Shape' Pos'}
+(* Partial instance *)
+Instance PartialLifted (Shape : Type) (Pos : Shape -> Type) 
+  (Shape' : Type) (Pos' : Shape' -> Type) `{Injectable Shape Pos Shape' Pos'}
   `(Partial Shape Pos) : Partial Shape' Pos' := {
     undefined := fun {A : Type} => embed Shape' Pos' undefined;
     error     := fun {A : Type} (msg : string) => embed Shape' Pos' (error msg)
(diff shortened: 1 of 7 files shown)


nbun (changes requested):
Looks good to me, well done. For now I would prefer to keep the original `Id`, `Just` and `Nothing` constructors until we have the infrastructure for dynamically building the effect stack based on the input program.

nbun (approved):

base/Free/Injectable.v
```diff
@@ -0,0 +1,72 @@
...
+   `{Injectable F_Shape F_Pos H_Shape H_Pos}
+   : Injectable F_Shape F_Pos 
+   (Comb.Shape G_Shape H_Shape) (Comb.Pos G_Pos H_Pos) | 2 := {
+      injS := fun s => inr (injS s); 
+      injP s := fun p => injP p;
+      (*prjS := fun s => match s with
+         | inr s' => prjS s'
+         | _     => None
+         end;*)
+   }.
```
ichistmeinname: As these instances overlap (which we explicitly need to tackle in the Haskell code basis), I'd like to see some tests using `injS` and `injP` that show how to overcome the overlapping problem -- or to check if there is any problem at all.

Can you add the definition of tests to the corresponding ticket? We can discuss later, if we find another candidate, who might be interested in testing this (as part of the review process).

MajaRet: These instances have explicitly assigned priorities (1 for `Inject_comb` and 2 for `Inject_rec`), so if I understand Coq's way of inferring instances correctly, Coq should always use `Inject_comb` when the instances overlap. 
We will likely need to add explicit type annotations to values using these instances anyway because we need to control the order of the effect stack, which cannot really be done reliably by letting Coq infer the instance.

I agree that we should add tests that test whether that's always the case. I'll add it to ticket #110 .

maclement (approved):
Looks good to me. Well done!

PR #117, opened 2020-07-07
Implement cbneed (sharing) operator

### LINKED ISSUE #115: Add an operator that models call-by-need evaluation
## Description

An operator `cbneed` that models call-by-need evaluation (call-by-name evaluation + sharing) should be added to `Malias`. This is the 'sharing operator'. 

Unlike `cbv` and `cbn`, `cbneed` cannot operate on arbitrary `Free` values, but only those values that contain the `Share` effect. Therefore, its type signature will contain an appropriate constraint.

## Motivation

Currently, `Malias` contains operators that model call-by-value (strict) evaluation and call-by-name (non-strict) evaluation.
The `cbneed` operator will model non-strict evaluation in combination with the sharing of effectful values (call-by-need evaluation). 

## Dependencies

This issue does not depend on any other open issues.

MajaRet:
### Issue

Closes #115 

### Description of the Change

This commit adds an operator `cbneed` that models call-by-need evaluation, i.e. non-strict evaluation and sharing. It requires its argument to contain the `Share` effect.

The operator does not support deep sharing yet (sharing inside of the arguments of a shared expression). 
It should eventually be extended to support deep sharing, as well.

### Verification Process

For now, I simply observed the results of various functions that share non-deterministic values using `cbneed` (using a handler). More extensive testing, ideally not just with non-determinism but also other effects that interact with sharing (such as tracing), should definitely be done.

### DIFF
--- a/base/Free/Malias.v
+++ b/base/Free/Malias.v
@@ -5,24 +5,25 @@ From Base Require Import Free.
 From Base Require Export Free.Instance.Comb.
 From Base Require Export Free.Instance.Share.
 
-(** An operator to model call-by-value evaluation *)
+(* An operator to model call-by-value evaluation *)
 Definition cbv {A : Type} {Shape : Type} {Pos : Shape -> Type} (p : Free Shape Pos A) 
   : Free Shape Pos (Free Shape Pos A) :=
   p >>= fun x => pure (pure x).
 
-(** An operator to model call-by-name evaluation *)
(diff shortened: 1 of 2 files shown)


nbun (changes requested):
I think it would be a good idea to define a notation for `(>>)` in order to reduce the visual clutter.

nbun (approved):

jo-hanna1997 (approved):

PR #118, opened 2020-07-07
Implement effect handlers

### LINKED ISSUE #116: Implement effect handlers
## Description

Handlers for `Free` programs containing various effects should be implemented.

## Motivation

Programs translated with the Free Compiler produce `Free` values. The semantics of such a program are given by a handler that transforms such a `Free` value into an appropriate data structure (e.g. `option` for `Maybe` or a choice tree for `ND`). 

To test the implementation of sharing and its interaction with other effects, we should define all necessary handlers to fully evaluate an effectful program containing sharing. 

## Dependencies

This issue does not depend on any open issues.

## ToDo

- [x] Implement a handler for a program without effects (`Identity` effect handler)
- [x] Implement a handler for a non-deterministic program (`ND` effect handler)
- [x] Implement a handler for a program containing tracing (`Trace` effect handler)
- [x] Implement a handler to evaluate sharing in combination with non-determinism (`Share` + `ND` handler)
- [x] Implement a handler to evaluate sharing in combination with tracing (`Share` + `Trace` handler)
- [x] Write tests for the handlers.

MajaRet:
### Issue

Closes #116 

### Description of the Change

Several effect handlers were defined:
- A handler for an effect-free program (`Identity` handler)
- A handler for programs containing `Maybe`
- A handler for non-deterministic programs (`ND` handler)
- A handler for the tracing effect (`Trace` handler)
- A call-time-choice handler (`Share` + `ND` handler)
- A sharing handler for programs containing tracing (`Share` + `Trace` handler)

In addition, a test module for the handlers was also added.

### Drawbacks & Alternate Designs

At this point, we require a separate sharing handler for each relevant effect (non-determinism and tracing). For a possible alternate design, see issue #125. 
In addition, we cannot currently handle sharing for more than one effect at once. If we wanted to handle an effect that contains both sharing and tracing, we would need to define a dedicated handler.

### Verification Process

The test module `SharingHandlerTests` was defined and used to test that the implementation produces the expected results. However, we should attempt to actually prove relevant properties of the handlers in the future.


base/Free/Handler.v
```diff
@@ -0,0 +1,155 @@
...
+  | pure x => pure (Leaf x)
+  | impure (inl ND.sfail) _ => pure (Empty A)
+  | impure (inl (ND.schoice mid)) pf => 
+      runChoice (pf true) >>= fun l =>
+      runChoice (pf false) >>= fun r =>
+      pure (Branch mid l r)  
+  | impure (inr s) pf => impure s (fun p : Pos s => runChoice (pf p))
+  end.
+
+Definition ID : Type := nat * nat * nat.
```
ichistmeinname: This definition should stay closer to its usage (`runNDSharing`?)

base/Free/Handler.v
```diff
@@ -0,0 +1,155 @@
...
+(* Handler for a program containing tracing *)
+Fixpoint runTracing A {Shape : Type} {Pos : Shape -> Type} 
+  (fm : Free (Comb.Shape Trace.Shape Shape) (Comb.Pos Trace.Pos Pos) A) 
+  : Free Shape Pos (A * list (option ID * string)) :=
```
ichistmeinname: What's the `option ID` for?

base/Free/Handler.v
```diff
@@ -0,0 +1,155 @@
...
+Fixpoint runNDSharing A {Shape : Type} {Pos : Shape -> Type} 
+  (n : nat * nat)
+  (fs : Free (Comb.Shape Share.Shape (Comb.Shape ND.Shape Shape))
+  (Comb.Pos Share.Pos (Comb.Pos ND.Pos Pos)) A) 
+  : Free (Comb.Shape ND.Shape Shape) (Comb.Pos ND.Pos Pos) A :=
+  let fix nameChoices (next : nat) (scope : nat * nat) 
+  (scopes : list (nat * nat)) 
+  (fs : Free (Comb.Shape Share.Shape (Comb.Shape ND.Shape Shape)) 
+  (Comb.Pos Share.Pos (Comb.Pos ND.Pos Pos)) A)
+  : Free (Comb.Shape ND.Shape Shape) (Comb.Pos ND.Pos Pos) A  :=
```
ichistmeinname: The identation of the types are a bit hard to read.

base/Free/Handler.v
```diff
@@ -0,0 +1,155 @@
...
+            end
+  | impure (inr s) pf => impure s (fun p : Pos s => runTracing (pf p))
+  end.
+
+(* Helper function to construct a triple from a pair and a single value *)
+Definition tripl A B C (p : A * B) (c : C) : A * B * C :=
+  let '(a,b) := p in (a,b,c). 
+
+(* Handler for sharing combined with non-determinism (call-time choice) *)
+Fixpoint runNDSharing A {Shape : Type} {Pos : Shape -> Type}
```
ichistmeinname: Does the sharing handler need to be specific to the effect? It seems like there a lot of code parts that are duplicated.

ichistmeinname (changes requested):
Is there a concrete concern against defining the handlers together with their instances? In its current form, when I need the `run` handler, I get a lot of additional exported function that I won't need. As alternative to moving the handlers to their instance files, we can use modules to restrict the visibility of functions when import this file.

base/Free/Handler.v
MajaRet: I'm working on that : )
I'll also define synonyms (such as `SNDShare` for `Comb.Shape Share.Shape (Comb.Shape ND.Shape Shape)` ) for the types so they're easier to read.

MajaRet:
Yes, I think it makes sense to put the handlers together with the instances.

base/Free/Handler.v
MajaRet: That's how we can determine which messages are different occurrences of the same shared variable. `option ID * string` is actually how `Trace.Shape` is defined. 
`runChoice` does the same thing, but the `option ID` is hidden in the datatype `Tree` there.

I haven't really tested this handler yet, so it may still change.

base/Free/Handler.v
MajaRet: It is used in `runTracing`, though.

base/Free/Handler.v
ichistmeinname: Ah, I see. That's the corresponding part to the ids used in the non-determinism effect.

base/Free/Handler.v
ichistmeinname: That, I got only later ; )

base/Free/Handler.v
MajaRet: The handler needs to take effect-specific actions (e.g. construct a `Choice` with the current scope and counter when it sees an `schoice`, and it needs to use `nameChoices` recursively to do that).
I think Niels is working on a more general solution, but that uses the higher-order approach to sharing, not our explicit scoping tag approach.

So unfortunately, I'm not sure we can avoid the duplication right now.

EDIT: Niels says it is possible, but it would be a rather big change and we'd need to investigate how/whether it can be implemented in Coq (a Haskell implementation does currently exist).

ichistmeinname: Then let's just open a ticket for some other time!

MajaRet: On it! : )

base/Free/Handler.v
```diff
@@ -0,0 +1,155 @@
+(** Handlers for non-determinism, sharing of non-determnistic values and 
+    Maybe, as well as a handler for an effect-free program *)
```
nbun:
```suggestion
(** Handlers for non-determinism, partiality, tracing and sharing of
    non-determinism and tracing, as well as a handler for
    effect-free programs *)
```

base/Free/Util/Search.v
```diff
@@ -0,0 +1,68 @@
...
+Definition ID : Type := nat * nat * nat.
+
+Inductive Tree A :=
+| Empty  : Tree A
+| Leaf   : A -> Tree A
+| Branch : option ID -> Tree A -> Tree A -> Tree A.
+
+Inductive Decision := L | R.
+
+Definition beq_id (id1 id2 : ID) : bool :=
```
nbun: I know that this is probably my own code, but for the sake of consistency I would prefer camel case everywhere.

base/Free/Handler.v
```diff
@@ -0,0 +1,155 @@
...
+  | impure (inr s) pf => impure s (fun p : Pos s => runMaybe (pf p))
+  end.
+
+(* Handler for a non-deterministic program *)
+Fixpoint runChoice A {Shape : Type} {Pos : Shape -> Type} 
+  (fc : Free (Comb.Shape ND.Shape Shape) (Comb.Pos ND.Pos Pos) A) 
+  : Free Shape Pos (Tree A) :=
+  match fc with
+  | pure x => pure (Leaf x)
+  | impure (inl ND.sfail) _ => pure (Empty A)
```
nbun: Writing `impure s pf` and then pattern matching on `s` instead of the whole pattern each time does lead to problems with the termination checker? I am not sure if I remember correctly and if this is still the case.

nbun (changes requested):

base/Free/Malias.v
```diff
@@ -20,11 +20,11 @@ Definition cbn {A : Type} {Shape : Type} {Pos : Shape -> Type}
...
   `{Injectable Share.Shape Share.Pos Shape Pos} (p : Free Shape Pos A)
   : Free Shape Pos (Free Shape Pos A) :=
-  Get >>= fun '(i,j) =>
-  Put (i+1,j) >>
-  pure (BeginShare (i,j) >>
-      Put (i,j+1) >>
+  Get _ _ >>= fun '(i,j) =>
+  Put _ _ (i+1,j) >>
+  pure (BeginShare _ _ (i,j) >>
+      Put _ _ (i,j+1) >>
```
ichistmeinname: Maybe you should add notations for these smart constructors that infer their types?

base/Free/Instance/Maybe.v
```diff
@@ -16,24 +17,51 @@ Module Maybe.
...
+      : Free Shape' Pos' A 
+     := pure x.
+
+    Definition Nothing_inj {A : Type} 
+                           (Shape' : Type)
+                           (Pos' : Shape' -> Type) 
+                           `{Injectable Shape Pos Shape' Pos'} 
+       : Free Shape' Pos' A 
+      := impure (injS tt) (fun p : Pos' (injS tt) => 
+      (fun (x : Void) => match x with end) (injP p)).
```
ichistmeinname, line 34: A direct pattern match on `injP p` isn't possible, because Coq does not "know" its exact type, right?

ichistmeinname (changes requested):

nbun (approved):

base/Free/Instance/Maybe.v
MajaRet: That's right. Writing 
```Coq
(...) := impure (injS tt) (fun p : Pos' (injS tt) => 
      match injP p with end).
```
causes the error message
```
Found a matching with no clauses on a term unknown to have an empty inductive type.
```

base/Free/Malias.v
MajaRet: Now that I think about it, it could be dangerous to infer the types at all. For example, take
```Coq
Definition p := cbneed (pure true) >>= fun sx => sx.
``` 
That's a valid program right now, but its type is now fixed as `Free Share.Shape Share.Pos (Bool Shape Pos)`, so we cannot add any other effects (e.g. define something that contains `Choice p p`) or properly handle the program.

`cbneed` shouldn't really be used by the user (especially if we abstract sharing through a typeclass (#137) ), but it still seems like a better idea to make `Shape` and `Pos` explicit.

nbun (approved):

base/Free/Malias.v
ichistmeinname: I'm not sure why a more polymorphic type wouldn't work here? Something like `Share :<: sig` in Haskell-notation.

ichistmeinname (approved):

PR #122, opened 2020-07-08
Implement a tracing effect in Coq

### LINKED ISSUE #120: Implement a tracing effect in Coq
## Description

An effect modelling Haskell's `trace` functionality should be implemented in Coq.

## Motivation

Just like non-determinism, tracing is an effect that produces different results depending on the presence or absence of sharing. 

For our purposes, tracing is more relevant to the compiler than non-determinism since Haskell supports tracing (via the `trace` function) whereas it does not support non-deterministic operations. So while non-determinism can never actually occur in a Haskell program, we could translate Haskell's `trace` function to Coq using the tracing effect, and ensure that it behaves the same as the Haskell version by making use of sharing.

MajaRet:
### Issue

Closes #120

### Description of the Change

A tracing effect (as well as a `trace` operator) was implemented in Coq based on [this implementation (p.6)](https://www-ps.informatik.uni-kiel.de/~sad/haskell2019-preprint.pdf).

### DIFF
--- a/base/Free/Instance/Trace.v
+++ b/base/Free/Instance/Trace.v
@@ -1,7 +1,7 @@
 (** * Definition of the tracing effect in terms of the free monad. *)
 
 From Base Require Import Free.
-From Base Require Import Free.Instance.Share.
+From Base Require Import Free.Util.Sharing.
 From Base Require Import Free.Util.Void.
 Require Export Coq.Strings.String.
 
--- a/base/Free/Instance/Trace.v
+++ b/base/Free/Instance/Trace.v
@@ -1,13 +1,11 @@
 (** * Definition of the tracing effect in terms of the free monad. *)
 
 From Base Require Import Free.
+From Base Require Import Free.Util.Sharing.
 From Base Require Import Free.Util.Void.
 Require Export Coq.Strings.String.
 
 Module Trace.
-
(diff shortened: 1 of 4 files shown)


base/Free/Instance/Trace.v
```diff
@@ -0,0 +1,39 @@
...
+
+From Base Require Import Free.
+From Base Require Import Free.Util.Void.
+Require Export Coq.Strings.String.
+
+Module Trace.
+
+  (* Type synonym for a tracing id. This id is needed to identify
+     shared messages. *)
+  Definition ID : Type := nat * nat * nat.
```
ichistmeinname: Maybe we should add a dedicated module for all constructs and helper functions related to sharing. If I'm not mistaken, you use the same definition for `ND`.

ichistmeinname (changes requested):

base/Free/Instance/Trace.v
MajaRet: Good idea.

MarvinLira: I would also prefer to have the share-related definitions like `ID` in a single location instead of defining them in each file where we need it.

ichistmeinname (approved):

MarvinLira (approved):

PR #128, opened 2020-07-10
Add support for simple let bindings

### LINKED ISSUE #80: Add support for simple let bindings
## Description

Non-recusive `let` bindings of the following form should be added to the intermediate representation and Haskell front end.

```haskell
let { x₁ = e₁; …; xₙ = eₙ } in e
```

The declarations of `x₁` through `xₙ` should be non-recursive, but may depend on each other (i.e., `x₁, …, xₙ` can be free in `e₁, …, eₙ`) in arbitrary order (i.e., `xᵢ` can depend on any `xⱼ` with `i, j ∈ {1, …, n}` as long as there is no cyclic dependency).

## Motivation

The pattern matching compiler generates `let` expressions during guard elimination. For example, the following function declaration

```haskell
foo x | otherwise = 1
```

is transformed as follows.

```haskell
foo a0
  = let a2 = undefined
        a1
          = case a0 of
                a3 -> if otherwise then 1 else a2
      in a1
```

Since the Free Compiler does not support `let` expressions at the moment, guards cannot be used even if the pattern amcthing compiler is enabled. For now, we do not need local function declarations nor recursive `let` bindings. Such simple `let` expressions could be modeled using lambda abstractions with minimal changes to the Haskell front end and no changes to the intermediate representation. However, since `let` expressions will likely be needed to model sharing, I think we should add them to the intermediate representation.

maclement:
### Issue

  Closes #80

### Description of the Change

Add support for simple let bindings of the structure:
``` 
 let x1 = e1
      . . .
      xn = en
      in e
```
where `x1 .... xn` are variable pattern and `e1 ... en` are expressions.


src/lib/FreeC/IR/Subst.hs
```diff
@@ -159,6 +159,7 @@ instance ApplySubst IR.Expr IR.Expr where
       let (subst', args') = newRenameArgs subst args
           expr'           = applySubst subst' expr
       in  IR.Lambda srcSpan args' expr' exprType
+    applySubst' expr@(IR.Let _ _ _ _) = expr -- TODO needed ?
```
ichistmeinname: The case for `Let` should be analogous to `Lambda`.
Let-expression can be expressed as lambda expressions as follows (if the let-binding is not recursive).

```haskell
let x = f 1 True
in x + 42
```
is the same as
```haskell
(\x -> x + 42) (f 1 True)
```

maclement: Oh, i missed that. I will keep that in mind when implementing the other functions for let. Thanks!

maclement:
I marked this pull as "ready to review". However, there is still a bug that i cannot identify.
The scenario is as simple as it can get. 
`let b = True in b`
The occurring error is "Identifier not in scope: b"

Edit: The bug is found and (hopefully) fixed.

src/lib/FreeC/Backend/Coq/Converter/Expr.hs
```diff
@@ -230,6 +230,18 @@ convertExpr' (IR.Lambda _ args expr _) [] [] = localEnv $ do
...
+  let bindPatsExprs = map (\x -> (IR.bindVarPat x, IR.bindExpr x)) binds
+  expr' <- convertExpr expr -- wrong order ?
+  foldrM helper expr' bindPatsExprs
+   where
+    helper :: (IR.VarPat, IR.Expr) -> Coq.Term -> Converter Coq.Term
+    helper (p, e) inExprTerm = do
+      e'      <- convertExpr e
+      qualid  <- renameAndDefineVar (IR.varPatSrcSpan p) (IR.varPatIsStrict p) (IR.varPatIdent p) (IR.varPatType p)
+      varPatType' <- mapM convertType (IR.varPatType p)
+      return (Coq.Let qualid [] varPatType' e' inExprTerm)
```
just95: The line marked with `--wrong order ?` is probably causing the "Identifier not in scope" error. `convertExpr` has to look up `b` but `b` is inserted by `helper` into the environment. With `foldr` instead of `foldrM` it should be possible to fix the order.

```suggestion
convertExpr' (IR.Let _ binds expr _) [] [] = localEnv $ do
  let bindPatsExprs = map (\x -> (IR.bindVarPat x, IR.bindExpr x)) binds
  foldr helper (convertExpr expr) bindPatsExprs
   where
    helper :: (IR.VarPat, IR.Expr) -> Converter Coq.Term -> Converter Coq.Term
    helper (p, e) inExprConverter = do
      e'      <- convertExpr e
      qualid  <- renameAndDefineVar (IR.varPatSrcSpan p) (IR.varPatIsStrict p) (IR.varPatIdent p) (IR.varPatType p)
      varPatType' <- mapM convertType (IR.varPatType p)
      Coq.Let qualid [] varPatType' e' <$> inExprConverter
```

src/lib/FreeC/Backend/Coq/Analysis/ConstantArguments.hs
```diff
@@ -143,6 +143,8 @@ makeConstArgGraph decls = do
           checkExpr e1 [] && checkExpr e2 [] && checkExpr e3 []
         checkExpr (IR.Case _ expr alts _) _ =
           checkExpr expr [] && all (flip checkAlt []) alts
+        checkExpr (IR.Let _ binds expr _) _ =
+          checkExpr expr [] && all (flip checkBind []) binds
```
just95: Here you probably have to check whether any binding shadows `x` in `expr`.

src/lib/FreeC/Backend/Coq/Converter/Expr.hs
maclement: That is a really nice solution i did not think of. And it fixes the bug. Thank you!

src/lib/FreeC/Backend/Coq/Analysis/ConstantArguments.hs
```diff
@@ -170,6 +172,15 @@ makeConstArgGraph decls = do
...
           | otherwise              = checkExpr expr args
 
+        -- | Applies 'checkExpr' to the right hand sides of a bind inside a
+        --   @let@ expression.
+        --
+        --   If a variable pattern shadows @x_i@, @x_i@ is not unchanged.
+        checkBind :: IR.Bind -> [IR.Expr] -> Bool
+        checkBind (IR.Bind _ varPat expr) args
+          | x `shadowedBy` [varPat] = not (callsG expr)
+          | otherwise               = checkExpr expr args
```
just95: If another binding of the `let` expression shadows `x`, `expr` also don't have to be checked recursively (i.e., move the check to `checkExpr`).

src/lib/FreeC/Backend/Coq/Analysis/DecreasingArguments.hs
```diff
@@ -207,6 +207,12 @@ checkDecArgs decls knownDecArgIndecies decArgIndecies = all
       let smaller' = withoutArgs args smaller
       in  checkExpr decArg smaller' expr []
 
+    -- The right hand side of a @let@-binding shadow existing (structurally
+    -- smaller) variables.
+    checkExpr' (IR.Let _ binds expr _) _ =
+      let smaller' = withoutArgs (map IR.bindVarPat binds) smaller
+      in checkExpr decArg smaller' expr []
```
just95: The right-hand sides of `binds` could contain recursive calls as well. Thus, they have to be checked, too.

src/lib/FreeC/Backend/Coq/Converter/Expr.hs
```diff
@@ -230,6 +230,17 @@ convertExpr' (IR.Lambda _ args expr _) [] [] = localEnv $ do
...
+convertExpr' (IR.Let _ binds expr _) [] [] = localEnv $ do
+  let bindPatsExprs = map (\x -> (IR.bindVarPat x, IR.bindExpr x)) binds
+  foldr helper (convertExpr expr) bindPatsExprs
+   where
+    helper :: (IR.VarPat, IR.Expr) -> Converter Coq.Term -> Converter Coq.Term
+    helper (p, e) inExprConverter = do
+      e'      <- convertExpr e
+      qualid  <- renameAndDefineVar (IR.varPatSrcSpan p) (IR.varPatIsStrict p) (IR.varPatIdent p) (IR.varPatType p)
+      varPatType' <- mapM convertType (IR.varPatType p)
+      Coq.Let qualid [] varPatType' e' <$> inExprConverter
```
just95: A bettern name for `helper` would be `convertBind`. If the decomposition of the binding is moved to the helper function, the name is also reflected in the type.

```suggestion
-- Let bindings.
convertExpr' (IR.Let _ binds expr _) [] [] = localEnv
  $ foldr convertBind (convertExpr expr) binds
 where
  convertBind :: IR.Bind -> Converter Coq.Term -> Converter Coq.Term
  convertBind bind inExprConverter = do
    let p = IR.bindVarPat bind
        e = IR.bindExpr bind
    e'     <- convertExpr e
    qualid <- renameAndDefineVar (IR.varPatSrcSpan p)
                                 (IR.varPatIsStrict p)
                                 (IR.varPatIdent p)
                                 (IR.varPatType p)
    varPatType' <- mapM convertType (IR.varPatType p)
    Coq.Let qualid [] varPatType' e' <$> inExprConverter
```

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithSections.hs
```diff
@@ -409,6 +409,10 @@ removeConstArgsFromExpr constArgs rootExpr = do
     (expr', []) <- removeConstArgsFromExpr' expr
     alts'       <- mapM removeConstArgsFromAlt alts
     return (IR.Case srcSpan expr' alts' exprType, [])
+  removeConstArgsFromExpr' (IR.Let srcSpan binds expr exprType) = do
+    binds'      <- mapM removeConstArgsFromBind binds
+    (expr', []) <- removeConstArgsFromExpr' expr
+    return (IR.Let srcSpan binds' expr' exprType, [])
```
just95: `shadowVarPats` have to be called here instead of in `removeConstArgsFromBind` since bindings shadow variables in other bindings and the `in` expression as well.

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithSections.hs
```diff
@@ -568,6 +580,10 @@ removeConstTypeArgsFromExpr constTypeVars rootExpr = do
     shadowVarPats args $ do
       (expr', []) <- removeConstTypeArgsFromExpr' expr
       return (IR.Lambda srcSpan args expr' exprType, [])
+  removeConstTypeArgsFromExpr' (IR.Let srcSpan binds expr exprType) = do
+      binds'      <- mapM removeConstTypeArgsFromBind binds
+      (expr', []) <- removeConstTypeArgsFromExpr' expr
+      return (IR.Let srcSpan binds' expr' exprType, [])
```
just95, line 587: Same as with `removeConstArgsFromExpr'`.

src/lib/FreeC/Frontend/Haskell/Simplifier.hs
```diff
@@ -922,3 +926,22 @@ simplifyAlt (HSE.Alt _ _ rhss@(HSE.GuardedRhss _ _) _) =
   experimentallySupported "Guards" rhss
 simplifyAlt (HSE.Alt _ _ _ (Just binds)) =
   notSupported "Local declarations" binds
+
+-- | Simplifies a group of bindings inside a 'let' clause
+simplifyBinds :: HSE.Binds SrcSpan -> Simplifier [IR.Bind]
+simplifyBinds binds@(HSE.IPBinds _ _) =
+  notSupported "Binding groups for implicit parameters" binds
```
just95: Implicit parameters are not supported in general. Thus, we could simplify this error message as well as the error message for `HSE.IPVar`.

```suggestion
  notSupported "Implicit parameters" binds
```

src/lib/FreeC/Frontend/Haskell/Simplifier.hs
```diff
@@ -922,3 +926,22 @@ simplifyAlt (HSE.Alt _ _ rhss@(HSE.GuardedRhss _ _) _) =
...
+     simplifyBind (HSE.PatBind srcSpan varPat (HSE.UnGuardedRhs _ expr) Nothing) = do
+        varPat' <- simplifyVarPat varPat
+        expr'   <- simplifyExpr expr
+        return (IR.Bind srcSpan varPat' expr')
+     simplifyBind (HSE.PatBind _  _ rhss@(HSE.GuardedRhss _ _) _) =
+       experimentallySupported "Guards" rhss
+     simplifyBind (HSE.PatBind _ _ _ (Just binds)) =
+       notSupported "Local declarations" binds
+     simplifyBind decl =
+       expected "A pattern binding" decl
```
just95:
```suggestion
       expected "A variable pattern binding" decl
```

src/lib/FreeC/Frontend/IR/Parser.hs
```diff
@@ -506,9 +506,10 @@ exprParser = setExprType <$> lExprParser <*> Parsec.optionMaybe
 --   >         | "if" expr "then" expr "else" expr     (conditional)
 --   >         | "case" expr "of" alts                 (case expression)
 --   >         | fexpr                                 (function application)
+--   >         | "let" binds "in" expr                 (let expression)
```
just95: I think that this is easier to read if `fexpr` is the last alternative.

```suggestion
--   >         | "if" expr "then" expr "else" expr     (conditional)
--   >         | "case" expr "of" alts                 (case expression)
--   >         | "let" binds "in" expr                 (let expression)
--   >         | fexpr                                 (function application)
```

src/lib/FreeC/Frontend/IR/Parser.hs
```diff
@@ -506,9 +506,10 @@ exprParser = setExprType <$> lExprParser <*> Parsec.optionMaybe
...
 lExprParser :: Parser IR.Expr
 lExprParser =
-  lambdaExprParser <|> ifExprParser <|> caseExprParser <|> fExprParser
+  lambdaExprParser <|> ifExprParser <|> caseExprParser <|> fExprParser <|> letExprParser
```
just95: To keep this in sync with the comment.

```suggestion
  lambdaExprParser <|> ifExprParser <|> caseExprParser <|> letExprParser <|> fExprParser
```

src/lib/FreeC/Frontend/IR/Parser.hs
```diff
@@ -542,6 +543,15 @@ lExprParser =
       <*> altsParser
       <*> return Nothing
 
+  letExprParser :: Parser IR.Expr
```
just95, line 552:
```suggestion
  -- @lexpr ::= "let" binds "in" expr | …@
  letExprParser :: Parser IR.Expr
```

src/lib/FreeC/Frontend/IR/Token.hs
```diff
@@ -222,6 +222,8 @@ data Keyword
   | TYPE      -- ^ @"type"@
   | UNDEFINED -- ^ @"undefined"@
   | WHERE     -- ^ @"where"@
+  | LET       -- ^ @"let"@
+  | IN        -- ^ @"in"@
```
just95: The other constructors are sorted alphabetically (in the `Pretty` instance, too). Furthermore, the two new keywords are missing from the list of `keywords`.

src/lib/FreeC/IR/Inlining.hs
```diff
@@ -140,6 +140,10 @@ inlineExpr decls = inlineAndBind
     shadowVarPats varPats $ do
       expr' <- inlineAndBind expr
       return ([], [], IR.Lambda srcSpan varPats expr' exprType)
+  inlineExpr' (IR.Let srcSpan binds expr exprType) = do
+    binds' <- mapM inlineBind binds
+    expr'  <- inlineAndBind expr
+    return ([], [], IR.Let srcSpan binds' expr' exprType)
```
just95: Move `shadowVarPats` here.

src/lib/FreeC/IR/Reference.hs
```diff
@@ -224,6 +224,8 @@ instance HasRefs IR.Expr where
   refSet (IR.IntLiteral _ _ exprType) = refSet exprType
   refSet (IR.Lambda _ args expr exprType) =
     unions [refSet args, withoutArgs args (refSet expr), refSet exprType]
+  refSet (IR.Let _  binds expr  exprType) =
+    unions [refSet expr, refSet binds, refSet exprType]
```
just95: The variables bound by the `let` expression are not referenced by the `let` expression (i.e., they are not free).

```suggestion
  refSet (IR.Let _ binds expr exprType) = withoutArgs (map IR.bindVarPat binds)
    $ unions [refSet expr, refSet binds, refSet exprType]
```

src/lib/FreeC/IR/Reference.hs
```diff
@@ -240,6 +242,11 @@ instance HasRefs IR.ConPat where
 instance HasRefs IR.VarPat where
   refSet = refSet . IR.varPatType
 
+-- | Bindings refer to the used variable pattern and the types used in type
+-- signatures.
```
just95:
```suggestion
-- | Bindings refer to the types used in the variable pattern's type signature
--   as well as all references of the right-hand side.
--
--   If the right-hand side refers to the bound variable, the bound variable is
--   also part of the bindings 'refSet'. Bound variables are removed from the
--   references of a @let@ expression by the 'HasRefs' instance of 'IR.Expr'.
```

src/lib/FreeC/IR/Similar.hs
```diff
@@ -296,6 +296,13 @@ instance Similar IR.TypeSchema where
 --        > ———————————————————————————————————————————————————
 --        >  Γ ⊢ if e₁ then e₂ else e₃ ≈ if f₁ then f₂ else f₃
 --
+--    * Two @let@ expressions are similar if their binds and the expression are
+--      similar.
+--
+--        >  Γ ⊢ e ≈ e',  Γ ⊢ x₁ ≈ x'₁, …, Γ ⊢ xₙ ≈ x'ₙ  Γ ⊢ e₁ ≈ e'₁, …, Γ ⊢ eₙ ≈ e'ₙ
+--        > ——————————————————————————————————————————————————————————————————————————————
+--        >  Γ ⊢ let { x₁ = e₁; …; xₙ = eₙ } in e ≈ let { x₁' = e₁'; …; xₙ' = eₙ' } in e'
+--
```
just95: This rule does not allow bindings to be renamed. For example, `let { x = y ; y = z } in x` should be similar to `let { x' = y'; y' = z } in x'`. The following rule takes renaming of bindings into account.

```suggestion
--    * Two @let@ expressions are similar if their bindings and the expression
--      are similar under an extended 'Renaming' @Γ'@ that maps the variables
--      bound by the first @let@ expression to the variables bound by the
--      second @let@ expression.
--
--        >  Γ' ⊢ e ≈ f,    Γ' ⊢ b₁ ≈ d₁,    …,    Γ' ⊢ bₙ ≈ dₙ
--        > ——————————————————————————————————————————————————————
--        >  Γ ⊢ let { b₁; …; bₙ } in e ≈ let { d₁; …; dₙ } in f
--
--        where @Γ' = Γ ∪ { x₁ ↦ y₁, …, xₙ ↦ yₙ }@ and @x₁, …, xₙ@
--        and @y₁, …, yₙ@ denote the names of variables bound by
--        @b₁, …, bₙ@ and @d₁, …, dₙ@, respectively.
```

The behavior specified above, also has to be implemented.

src/lib/FreeC/IR/Similar.hs
```diff
@@ -477,3 +487,13 @@ instance Similar IR.ConDecl where
...
+
+-- | Two bindings are similar if their variable pattern and their expression are
+--   similar.
+--
+--   >  Γ ⊢ v ≈ v', …, Γ ⊢ e ≈ e'
+--   > ———————————————————————————————
+--   >   Γ ⊢  v = e ≈ v' = e'
+instance Similar IR.Bind where
+  similar' (IR.Bind _ vP1 e1) (IR.Bind _ vP2 e2) =
+    similar' vP1 vP2 .&&. similar' e1 e2
```
just95: The meta variables `p` and `q` are used for variable patterns with optional type annotation everywhere else.

```suggestion
-- | Two bindings are similar if their variable pattern and their expression are
--   similar.
--
--   >  Γ ⊢ p ≈ q, …, Γ ⊢ e ≈ f
--   > ——————————————————————————
--   >     Γ ⊢ p = e ≈ q = f
instance Similar IR.Bind where
  similar' (IR.Bind _ p e) (IR.Bind _ q f) =
    similar' p q .&&. similar' e f
```

src/lib/FreeC/IR/Subst.hs
```diff
@@ -159,6 +159,11 @@ instance ApplySubst IR.Expr IR.Expr where
       let (subst', args') = newRenameArgs subst args
           expr'           = applySubst subst' expr
       in  IR.Lambda srcSpan args' expr' exprType
+    applySubst' (IR.Let srcSpan binds expr exprType) =
+      let (subst', varpats') = newRenameArgs subst (map IR.bindVarPat binds)
+          binds'             = zipWith (\v b -> b {IR.bindVarPat = v}) varpats' binds
+          expr'              = applySubst subst' expr
+      in IR.Let srcSpan binds' expr' exprType
```
just95: The substitution has to be applied to the bindings as well.

src/lib/FreeC/IR/Subst.hs
```diff
@@ -250,6 +261,11 @@ instance ApplySubst IR.Type IR.VarPat where
     let maybeVarType' = applySubst subst maybeVarType
     in  IR.VarPat srcSpan varIdent maybeVarType' isStrict
 
+instance ApplySubst IR.Type IR.Bind where
```
just95, line 277: Add a documentation comment.

src/lib/FreeC/IR/Syntax/Expr.hs
```diff
@@ -81,6 +81,13 @@ data Expr
            , lambdaEprRhs   :: Expr
            , exprTypeSchema :: Maybe TypeSchema
            }
+
+  | -- | A let expression.
+    Let { exprSrcSpan :: SrcSpan
+        , binds :: [Bind]
+        , inExpr :: Expr
+        , exprTypeSchema :: Maybe TypeSchema
+        }
```
just95: Fields that are specific to other constructors of `Expr` have names that are prefixed with the constructor names.

```suggestion
    Let { exprSrcSpan    :: SrcSpan
        , letExprBinds   :: [Bind]
        , letExprIn      :: Expr
        , exprTypeSchema :: Maybe TypeSchema
        }
```

src/lib/FreeC/IR/Syntax/Expr.hs
```diff
@@ -244,10 +251,16 @@ prettyExprPred' 0 (Lambda _ args expr _) =
     <>  hsep (map pretty args)
     <+> prettyString "->"
     <+> prettyExprPred 0 expr
+prettyExprPred' 0 (Let _ bs e _) =
+  prettyString "let"
+    <+> hsep (map pretty bs)
+    <+> prettyString "in"
+    <+> prettyExprPred 0 e
```
just95: The `Pretty` instance should generate parsable IR code. Thus, there should be braces around the bindings.

```suggestion
prettyExprPred' 0 (Let _ bs e _) =
  prettyString "let"
    <+> braces
          (space <> prettySeparated (semi <> space) (map pretty bs) <> space)
    <+> prettyString "in"
    <+> prettyExprPred 0 e
```

Also update the comment above this code block to include `let` expressions.

src/lib/FreeC/IR/Syntax/Expr.hs
```diff
@@ -358,3 +371,23 @@ instance Pretty VarPat where
     parens (pretty varName <+> colon <> colon <+> pretty varType)
   pretty (VarPat _ varName (Just varType) True) =
     char '!' <> parens (pretty varName <+> colon <> colon <+> pretty varType)
+
+
+-------------------------------------------------------------------------------
```
just95:
```suggestion

-------------------------------------------------------------------------------
```

src/lib/FreeC/Pass/ResolverPass.hs
```diff
@@ -785,3 +791,9 @@ instance Resolvable IR.VarPat where
   resolve varPat = do
     varType' <- mapM resolve (IR.varPatType varPat)
     return varPat { IR.varPatType = varType' }
+
+instance Resolvable IR.Bind where
```
just95, line 803: Add a documentation comment.

src/lib/FreeC/Pass/TypeInferencePass.hs
```diff
@@ -666,6 +666,22 @@ annotateExprWith' (IR.Lambda srcSpan args expr _) resType =
     addTypeEquation srcSpan funcType resType
     return (IR.Lambda srcSpan args' expr' (makeExprType resType))
 
+
+annotateExprWith' (IR.Let srcSpan binds expr _) resType =
+  withLocalTypeAssumption $ do
+    retType  <- liftConverter freshTypeVar
+    binds'   <- mapM annotateBind binds
+    expr'    <- annotateExprWith expr retType
```
just95:
```suggestion
    exprType <- liftConverter freshTypeVar
    binds'   <- mapM annotateBind binds
    expr'    <- annotateExprWith expr exprType
```

src/lib/FreeC/Pass/TypeInferencePass.hs
```diff
@@ -666,6 +666,22 @@ annotateExprWith' (IR.Lambda srcSpan args expr _) resType =
...
+    return (IR.Let srcSpan binds' expr' (makeExprType resType))
+     where
+       annotateBind :: IR.Bind -> TypeInference IR.Bind
+       annotateBind (IR.Bind bindSrcSpan bindVarPat bindExpr) = do
+           bindVarPat' <- annotateVarPat bindVarPat
+           bindType    <- liftConverter freshTypeVar
+           bindExpr'   <- annotateExprWith bindExpr bindType
+           addTypeEquation srcSpan (fromJust $ IR.varPatType bindVarPat') bindType
+           return (IR.Bind bindSrcSpan bindVarPat' bindExpr')
+
```
just95: This implementation does not infer the type of polymorphic `let` bindings correctly. The following is allowed in Haskell for example.

```haskell
let l = length in l [1,2,3] + l [True,False]
```

I expect polymorphic `let` bindings to be needed for the translation of guards. However, to support them we would have to add type arguments to `let` bindings as shown below. And probably also have to change a lot of code in the `TypeInferencePass`.

```haskell
let l @a = length @a in l @Integer [1,2,3] + l @Bool [True,False]
```

Thus, I'd say that we should ignore this for now. However, there should be documentation and/or a ticket for the missing feature.

src/lib/FreeC/Pass/TypeInferencePass.hs
```diff
@@ -666,6 +666,22 @@ annotateExprWith' (IR.Lambda srcSpan args expr _) resType =
...
+annotateExprWith' (IR.Let srcSpan binds expr _) resType =
+  withLocalTypeAssumption $ do
+    retType  <- liftConverter freshTypeVar
+    binds'   <- mapM annotateBind binds
+    expr'    <- annotateExprWith expr retType
+    return (IR.Let srcSpan binds' expr' (makeExprType resType))
+     where
+       annotateBind :: IR.Bind -> TypeInference IR.Bind
+       annotateBind (IR.Bind bindSrcSpan bindVarPat bindExpr) = do
+           bindVarPat' <- annotateVarPat bindVarPat
```
just95: Since other bindings can refer to this binding but are entered by `annotateVarPat` into the type assumption, all variable patterns have to be annotated before the first binding.

src/lib/FreeC/Pass/TypeInferencePass.hs
```diff
@@ -666,6 +666,22 @@ annotateExprWith' (IR.Lambda srcSpan args expr _) resType =
...
+           bindType    <- liftConverter freshTypeVar
+           bindExpr'   <- annotateExprWith bindExpr bindType
+           addTypeEquation srcSpan (fromJust $ IR.varPatType bindVarPat') bindType
```
just95: I guess that we can recycle the type the variable pattern has been annotated with.

```suggestion
           let Just bindType = IR.varPatType bindVarPat'
           bindExpr' <- annotateExprWith bindExpr bindType
```

just95 (changes requested):

src/lib/FreeC/Pass/TypeInferencePass.hs
ichistmeinname: @macmaccamcam Can you open a ticket and resolve the conversation afterwards?

maclement: Yes, i was just about to do that.

src/lib/FreeC/Backend/Coq/Analysis/ConstantArguments.hs
```diff
@@ -143,8 +143,9 @@ makeConstArgGraph decls = do
           checkExpr e1 [] && checkExpr e2 [] && checkExpr e3 []
         checkExpr (IR.Case _ expr alts _) _ =
           checkExpr expr [] && all (flip checkAlt []) alts
-        checkExpr (IR.Let _ binds expr _) _ =
-          checkExpr expr [] && all (flip checkBind []) binds
+        checkExpr (IR.Let _ binds expr _) _
+          | x `shadowedBy` map IR.bindVarPat binds = not (callsG expr)
+          | otherwise                              = checkExpr expr []
```
just95: The right-hand sides of the bindings still need to be checked with `checkExpr` and `callsG`, respectively.

src/lib/FreeC/Backend/Coq/Analysis/DecreasingArguments.hs
```diff
@@ -232,6 +232,11 @@ checkDecArgs decls knownDecArgIndecies decArgIndecies = all
       let smaller' = withoutArgs varPats smaller
       in  checkExpr decArg smaller' expr []
 
+    checkBind :: IR.Bind -> Bool
+    checkBind (IR.Bind _ varPat expr) =
+      let smaller' = withoutArgs [varPat] smaller
```
just95: All bound variables are visible in all binders. Thus, you have to use the same `smaller'` as in the `IR.Let` rule of `checkExpr'`.

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithHelpers.hs
```diff
@@ -95,9 +95,9 @@ transformRecFuncDecl (IR.FuncDecl srcSpan declIdent typeArgs args maybeRetType e
     -- Generate main function declaration. The main function's right-hand side
     -- is constructed by replacing all @case@-expressions of the decreasing
     -- argument by an invocation of the corresponding recursive helper function.
-    let mainExpr = replaceSubterms' expr (zip caseExprsPos helperApps)
+    let maletExprIn = replaceSubterms' expr (zip caseExprsPos helperApps)
         mainDecl =
-          IR.FuncDecl srcSpan declIdent typeArgs args maybeRetType mainExpr
+          IR.FuncDecl srcSpan declIdent typeArgs args maybeRetType maletExprIn
```
just95: *find-and-replace* seems to have gone wrong :D

src/lib/FreeC/Frontend/IR/Token.hs
```diff
@@ -241,8 +242,10 @@ keywords = map
   , ELSE
   , ERROR
   , FORALL
+  , IN
   , IF
   , IMPORT
+  , LET
```
just95:
```suggestion
  , IF
  , IMPORT
  , IN
  , LET
```

src/lib/FreeC/Frontend/IR/Token.hs
```diff
@@ -258,13 +261,13 @@ instance Pretty Keyword where
   pretty ELSE      = prettyString "else"
   pretty ERROR     = prettyString "error"
   pretty FORALL    = prettyString "forall"
+  pretty IN        = prettyString "in"
   pretty IF        = prettyString "if"
   pretty IMPORT    = prettyString "import"
+  pretty LET       = prettyString "let"
```
just95:
```suggestion
  pretty IF        = prettyString "if"
  pretty IMPORT    = prettyString "import"
  pretty IN        = prettyString "in"
  pretty LET       = prettyString "let"
```

src/lib/FreeC/IR/Similar.hs
```diff
@@ -337,7 +343,9 @@ similarExpr (IR.Case _ e as _) (IR.Case _ f bs _) =
 similarExpr (IR.If _ e1 e2 e3 _) (IR.If _ f1 f2 f3 _) =
   similar' e1 f1 .&&. similar' e2 f2 .&&. similar' e3 f3
 similarExpr (IR.Let _ bs e _) (IR.Let _ cs f _) =
-  similar' e f .&&. similar' bs cs
+  let ns = map (IR.varPatQName . IR.bindVarPat) bs
+      ms = map (IR.varPatQName . IR.bindVarPat) cs
+  in similar' e f . extendRenaming IR.ValueScope ns ms .&&. similar' bs cs
```
just95: Since the `Similar` instance for `IR.Bind` extends the renaming by the variable bound by the binding only, this corresponds to

```
Γ ∪ { xᵢ ↦ yᵢ } ⊢ bᵢ ≈ dᵢ
```

instead of

```
Γ' ⊢ bᵢ ≈ dᵢ
```

If you remove `extendRenaming` from the `Similar` instance of `IR.Bind`, the following should work.

```suggestion
  in  (similar' e f .&&. similar' bs cs) . extendRenaming IR.ValueScope ns ms
```

src/lib/FreeC/IR/Subst.hs
```diff
@@ -161,7 +161,7 @@ instance ApplySubst IR.Expr IR.Expr where
       in  IR.Lambda srcSpan args' expr' exprType
     applySubst' (IR.Let srcSpan binds expr exprType) =
       let (subst', varpats') = newRenameArgs subst (map IR.bindVarPat binds)
-          binds'             = zipWith (\v b -> b {IR.bindVarPat = v}) varpats' binds
+          binds'             = zipWith (\v (IR.Bind s _ e) -> IR.Bind s v (applySubst subst' e)) varpats' binds
```
just95: An `ApplySubst IR.Expr IR.Bind` instance could help to shorten this line.

src/lib/FreeC/Pass/ResolverPass.hs
```diff
@@ -746,6 +746,7 @@ instance Resolvable IR.Expr where
     alts'      <- mapM resolve alts
     exprType'  <- mapM resolve exprType
     return (IR.Case srcSpan scrutinee' alts' exprType')
+  -- Shadow variable patterns and recursively
```
just95:
```suggestion

  -- Shadow variable patterns and resolve recursively.
```

Or even better: move this up to the rule for `IR.Lambda` and update the comment there.

just95 (changes requested):

src/lib/FreeC/Pass/TypeInferencePass.hs
just95: See #157.

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithHelpers.hs
maclement: Whoops. Good catch!

src/lib/FreeC/Backend/Coq/Analysis/ConstantArguments.hs
```diff
@@ -173,6 +175,8 @@ makeConstArgGraph decls = do
           | x `shadowedBy` varPats = not (callsG expr)
           | otherwise              = checkExpr expr args
 
+
+
         -- | Tests whethe the given variable is shadowed by the given
         --   variale patterns.
```
just95:
```suggestion

        -- | Tests whethe the given variable is shadowed by the given
        --   variale patterns.
```

src/lib/FreeC/IR/Subst.hs
```diff
@@ -179,6 +180,10 @@ instance ApplySubst IR.Expr IR.Alt where
         expr'              = applySubst subst' expr
     in  IR.Alt srcSpan conPat varPats' expr'
 
+instance ApplySubst IR.Expr IR.Bind where
```
just95, line 189:
```suggestion
-- | Applies the given expression substitution to an the right-hand side of a
--   @let@ binding.
--
--   The variable that is bound by the binding is not renamed by this instance.
--   The 'ApplySubst' instance for expressions renames all variables that are
--   bound by @let@ expressions.
instance ApplySubst IR.Expr IR.Bind where
```

just95 (approved):

Daniel-Teut (approved):

PR #114, opened 2020-07-06
Issue 106: IR -> LIR -> Agda translations

JonasHoefer:
### Issue

Closes #106 

### Description of the Change

- IR to LIR and LIR to Agda translations for expressions.
- translation of imports for Agda (for easier testing)

### Verification Process

I used the existing Agda tests and created new test cases (#66) for some new syntactic constructs.

### Additional Notes

~~Because the IR to LIR translation runs before the IR to Agda/Coq translation it's not possible to use `localEnv` and declare variables during it, because the back end has to do lookups for the renamed versions. As a consequence, adding variables to the environment must happen in the back end.~~

As a temporary solution we store the back end names in `LIR.Var` expressions, making lookups and `localEnv` calls in the back end obsolete. A better solution would be to generate globally unique names in the IR to LIR translation and rename in them in the back end. This would allow us to remove Agda and Coq names from the Converter monad and decouple the pipeline and the back ends.

Currently missing Agda back end features:
- mutual data types and recursion
- type synonyms

Issue #98, opened 2020-06-26
Add abstractions for front end and back end


just95:
## Description

Add data types that abstract from front ends and back ends. A front end basically maps [`SrcFile`s](https://github.com/FreeProving/free-compiler/blob/6864d3b5a33b1b1206b355a9b3b62618605d99db/src/lib/FreeC/IR/SrcSpan.hs#L36) to [`IR.Module`s](https://github.com/FreeProving/free-compiler/blob/6864d3b5a33b1b1206b355a9b3b62618605d99db/src/lib/FreeC/IR/Syntax/Module.hs#L18). A back end on the other hand maps `IR.Module`s to the source code in the target language (could be a `String` or also a `SrcFile` for example). Additionally, the data type should provide mechanisms for the generation of back end specific files such as `_CoqProject`. Select an appropriate monad and add additional information such as a name, description, file extension etc. to the data type as you see fit.

## Motivation

Since we want to add command line options to select different front ends and back ends, we have to abstract from the concrete front end or back end in `Main`. In this first step, we want to lay the ground work for such command line options. To test the abstractions, pseudo front and back ends should be added for the intermediate representation (IR) which just parse and pretty print, respectively. These pseudo front and back ends will also improve debugging of the Free Compiler by allowing us to control the IR that is plugged into the pipeline and seeing the IR after all passes have been applied.

## Tasks

Do the following on the `main` branch.

 - [x] Add modules `FreeC.Backend` and `FreeC.Frontend`.
 - [x] Add the data types described above to the corresponding modules.
 - [x] Add instance of the front end data type for the Haskell front end.
 - [x] Add instance of the back end data type for the Coq front end.
 - [x] Add a list or map of available back ends / front ends to the modules.

Do the following on the `dev-agda` branch.

 - [x] Merge the `main` branch (or the feature branch for this issue).
 - [x] Add an instance of the back end data type for the Agda back end.

Issue #97, opened 2020-06-26
Invoke passes from `Main` module


just95:
## Description

The invocation of the [compiler pipeline](https://github.com/FreeProving/free-compiler/blob/6864d3b5a33b1b1206b355a9b3b62618605d99db/src/lib/FreeC/Pipeline.hs#L27)  should be moved from [`convertModule`](https://github.com/FreeProving/free-compiler/blob/6864d3b5a33b1b1206b355a9b3b62618605d99db/src/lib/FreeC/Backend/Coq/Converter/Module.hs#L28) to the command line interface in [`Main.hs`](https://github.com/FreeProving/free-compiler/blob/6864d3b5a33b1b1206b355a9b3b62618605d99db/src/exe/Main.hs#L125).

## Motivation

The compiler `Pass`es are currently invoked in the Coq back end during the conversion of modules. This is mainly the case because both the passes and the conversion have to be performed in the same [`moduleEnv`](https://github.com/FreeProving/free-compiler/blob/6864d3b5a33b1b1206b355a9b3b62618605d99db/src/lib/FreeC/Monad/Converter.hs#L190). Since we want to add support for more back ends and the compiler pipeline must be executed before each back end, the `Pass`es should be executed before `convertModule` is invoked in `Main.hs`.

## Tasks

Do the following on the `main` branch.

 - [x] Call `runPipeline` in `Main.hs`.
 - [x] Rename `convertModule'` to `convertModule` in the [Coq back end](https://github.com/FreeProving/free-compiler/blob/6864d3b5a33b1b1206b355a9b3b62618605d99db/src/lib/FreeC/Backend/Coq/Converter/Module.hs#L28).

Do the following on the `dev-agda` branch.

 - [x] Merge the `main` branch (or the feature branch for this issue).
 - [x] Rename `convertModule'` to `convertModule` in the [Agda back end](https://github.com/FreeProving/free-compiler/blob/dev-agda/src/lib/FreeC/Backend/Agda/Converter/Module.hs).

PR #129, opened 2020-07-14
Issue 121: Expand Agda base library

### LINKED ISSUE #121: Expand Agda base library
The Agda base lib is still missing some data structures and functions. Add the following data structures and functions associated to them (as listed in the `prelude.toml`).
- [x] Integer
- [x] Pair 
- [x] List
- [x] Unit

JonasHoefer:
### Issue

Closes #121 Blocked by #114 

### Description of the Change

Adds the following 
- missing functions to the Agda base library.
- an Agda example using the new functions

### Verification Process

The example type checks.


base/agda/Base/Free/Integer.agda
```diff
@@ -0,0 +1,69 @@
...
+  negative = record
+    { Constraint = λ _ → ⊤ᵖ
+    ; fromNeg    = λ n → - (+ n)
+    }
+
+-- If it encounters a negative exponent the Haskell implementation of @(^)@ raises an exception using @errorWithoutStackTrace@.
+_^_ : ∀ {S P} → ⦃ Partial S P ⦄ → Free S P (ℤ S P) → Free S P (ℤ S P) → Free S P (ℤ S P)
+mx ^ pure (+ 0)         = pure 1
+mx ^ pure (+ (suc n))   = mx * (mx ^ (pure (+ n)))
+mx ^ pure (ℤᵖ.negsuc n) = error "*** Exception: Negative exponent"
```
just95: The `*** Exception:` prefix is not part of the error message but printed by GHC's runtime system.

```haskell
GHCi> error "Negative exponent"
*** Exception: Negative exponent
CallStack (from HasCallStack):
  error, called at <interactive>:1:1 in interactive:Ghci1
```

```suggestion
mx ^ pure (ℤᵖ.negsuc n) = error "Negative exponent"
```

base/agda/Base/Free/Integer.agda
```diff
@@ -0,0 +1,69 @@
+module Base.Free.Integer where
+
+open import Data.Nat                   using (zero; suc)
+open import Data.Integer               using (+_; -_) renaming (ℤ to ℤᵖ; _+_ to _+ᵖ_; _-_ to _-ᵖ_; _*_ to _*ᵖ_)
+open import Data.Integer.Properties    using (_≤?_; _<?_) renaming (_≟_ to _≟ᵖ_)
+open import Relation.Nullary.Decidable using (⌊_⌋)
+
+-- for literals
```
just95:
```suggestion
-- Imports for literals.
```

base/coq/Prelude/Integer.v
```diff
@@ -46,10 +47,12 @@ Section SecInteger.
...
   (* exponentiation *)
-  Definition powInteger (n1 : Free' Integer') (n2 : Free' Integer') : Free' Integer' :=
+  Definition powInteger (P : Partial Shape Pos) (n1 : Free' Integer') (n2 : Free' Integer') : Free' Integer' :=
     n1 >>= fun(n1' : Integer') =>
-      n2 >>= fun(n2' : Integer') =>
-        pure (Z.pow n1' n2').
+      if Z.ltb n1' 0
+        then error "Negative exponent"
+        else n2 >>= fun(n2' : Integer') =>
+          pure (Z.pow n1' n2').
```
just95: `n1` is the base and not the exponent. Thus, it has to be tested whether `n2'` is negative and not whether `n1'` is negative. Furthermore, the order of evaluation needs to be changed since the exponent is evaluated first by Haskell as the following example demonstrates.

```haskell
GHCi> error "a" ^ error "b"
*** Exception: b
CallStack (from HasCallStack):
  error, called at <interactive>:7:13 in interactive:Ghci3
```

Interestingly, the base does not seem to be evaluated if the exponent is `0`.

```haskell
GHCi> undefined ^ 0
1
```

Thus, I think that the operator needs to be implemented as follows.

```coq
  Definition powInteger (P : Partial Shape Pos) (n1 : Free' Integer') (n2 : Free' Integer') : Free' Integer' :=
    n2 >>= fun(n2' : Integer') =>
      if Z.ltb n2' 0
        then error "Negative exponent"
        else if Z.eqb n2' 0
               then pure 1%Z
               else n1 >>= fun(n1' : Integer') =>
                 pure (Z.pow n1' n2').
```

---

I actually never considered the order of evaluation when I've implemented the operators originally. Thus, we should check all other operators as well and add `./example`s for the expected behavior of each operator. For example, `-`, `>` and `<=` also seem to evaluate the second argument first while `+`, `*`, `<` and `>=` evaluate the first argument first.

```haskell
GHCi> error "a" - error "b"
*** Exception: b
CallStack (from HasCallStack):
  error, called at <interactive>:12:13 in interactive:Ghci4
```

The Coq base library always evaluates the first argument first whereas the Agda base library evalutes the first argument of `+`, `-`, `*`, `<=` and `<` first and the second argument of `>` and `>=`. This should maybe be address this in a separate issue, though.

base/agda/Base/Free/Integer.agda
```diff
@@ -0,0 +1,69 @@
...
+
+_≥_ : ∀ {S P} → Free S P (ℤ S P) → Free S P (ℤ S P) → Free S P (𝔹 S P)
+mx ≥ my = my < mx
+
+_≟_ : ∀ {S P} → Free S P (ℤ S P) → Free S P (ℤ S P) → Free S P (𝔹 S P)
+mx ≟ my = mx >>= λ x → my >>= λ y → pure ⌊ x ≟ᵖ y ⌋
+
+_≠_ : ∀ {S P} → Free S P (ℤ S P) → Free S P (ℤ S P) → Free S P (𝔹 S P)
+mx ≠ my = not (mx ≟ my)
+instance
```
just95: I'd insert a blank line here since the `instance` definition does not belong to `_≠_`.

```suggestion

instance
```

base/agda/Base/Free/Integer.agda
```diff
@@ -0,0 +1,69 @@
...
+    { Constraint = λ _ → ⊤ᵖ
+    ; fromNeg    = λ n → - (+ n)
+    }
+
+-- If it encounters a negative exponent the Haskell implementation of @(^)@ raises an exception using @errorWithoutStackTrace@.
+_^_ : ∀ {S P} → ⦃ Partial S P ⦄ → Free S P (ℤ S P) → Free S P (ℤ S P) → Free S P (ℤ S P)
+mx ^ pure (+ 0)         = pure 1
+mx ^ pure (+ (suc n))   = mx * (mx ^ (pure (+ n)))
+mx ^ pure (ℤᵖ.negsuc n) = error "*** Exception: Negative exponent"
+mx ^ impure s pf        = impure s pf
```
just95: I think that we should use `>>=` for the definition of `_^_` such that we can understand the order of evaluation better.

base/agda/Base/Free/Bool.agda
```diff
@@ -0,0 +1,22 @@
...
+True = pure true
+
+False : ∀ {S P} → Free S P (𝔹 S P)
+False = pure false
+
+_∧_ : ∀ {S P} → Free S P (𝔹 S P) → Free S P (𝔹 S P) → Free S P (𝔹 S P)
+mx ∧ my = mx >>= λ x → my >>= λ y → pure (x ∧ᵖ y)
+
+_∨_ : ∀ {S P} → Free S P (𝔹 S P) → Free S P (𝔹 S P) → Free S P (𝔹 S P)
+mx ∨ my = mx >>= λ x → my >>= λ y → pure (x ∨ᵖ y)
```
just95: In Haskell, the `&&` and `||` operators are lazy. The second argument is only evaluated if their first argument does not determine the result already.

```suggestion
_∧_ : ∀ {S P} → Free S P (𝔹 S P) → Free S P (𝔹 S P) → Free S P (𝔹 S P)
mx ∧ my = mx >>= λ x → if x then my else False

_∨_ : ∀ {S P} → Free S P (𝔹 S P) → Free S P (𝔹 S P) → Free S P (𝔹 S P)
mx ∨ my = mx >>= λ x → if x then True else my
```

example/Agda/List.agda
```diff
@@ -0,0 +1,208 @@
...
+  concat :
+    ∀ {Shape} {Pos} {a} →
+    Free Shape Pos (List Shape Pos (List Shape Pos a)) →
+    Free Shape Pos (List Shape Pos a)
+  concat x = foldr (pure λ x₁ → pure λ x₂ → append x₁ x₂) Nil x
+  and :
+    ∀ {Shape} {Pos} →
+    Free Shape Pos (List Shape Pos (𝔹 Shape Pos)) →
+    Free Shape Pos (𝔹 Shape Pos)
+  and x = foldr (pure λ x₁ → pure λ x₂ → (x₁ ∧ x₂)) True x
```
just95: Generated code should not be committed. Use the `.gitignore`d `./example/generated` folder for code generated from examples. We may also want to configure the CI pipeline to convert the examples to Agda automatically. This is blocked by the Agda tasks of #98 and #97. We should add a separate issue for the change of the CI pipeline and `./tool/full-test.sh` script.

just95 (changes requested):

base/agda/Base/Free/Integer.agda
JonasHoefer: I split the definition to please the termination checker. Since `ℤ` is a type from the stdlib i cannot add size annotations.

just95: It should be possible to define an operation `_^ᵖ_ : ℤ → ℕ → ℤ` without size annotations and use it to define a non-recursive `_^_` using `>>=`.

JonasHoefer: Yes, this should be no problem. I just tried to keep the definition as simple as possible.

base/coq/Prelude/Integer.v
JonasHoefer: Concerning partial values this implementation of `powInteger` should behave similar to the [Haskell implementation](https://hackage.haskell.org/package/base-4.14.0.0/docs/src/GHC.Real.html#%5E) (i.e. handle exponents which are negative, 0 or 1 without evaluating `n1`). This implementation is also consistent with the current Agda version. I will implement the same checks in `_^_` before calling `_^ᵖ_ : ℤ → ℕ → ℤ`. 
The actual implementations of exponentiation differ between the 3 languages. I'm not sure if this a problem when proofing facts about functions using `^`.

nbun (approved):

base/coq/Prelude/Integer.v
just95: Regarding the order of evaluation of the other operators, I've now created a new issue (see #133).

base/agda/Base/Prelude.agda
```diff
@@ -1 +1,7 @@
 module Base.Prelude where
+
+open import Base.Free.Bool    public
+open import Base.Free.Integer public
+open import Base.Free.List    public
+open import Base.Free.Pair    public
+open import Base.Free.Unit    public
```
just95: I think the modules that are part of the `Prelude` should be moved from `./base/agda/Base/Free` to `./base/agda/Base/Prelude`.

just95 (changes requested):

just95 (approved):

PR #141, opened 2020-07-21
Translate Coq Expressions using the lifted IR

### LINKED ISSUE #123: Use lifted IR expression translation in the Coq back end
Integrate the translation from #106 in the Coq backend. Blocked by #114

Daniel-Teut:
### Issue
Closes #123.

### Description of the Change
Change the translation of Coq expressions such that the lifted IR translation for expressions is used.


src/lib/FreeC/Backend/Coq/Converter/Expr.hs
```diff
@@ -2,306 +2,85 @@
...
+  return $ Coq.Qualid qualid
+convertLiftedExpr (LIR.SmartCon srcSpan name) = do
+  qualid <- lookupSmartIdentOrFail srcSpan name
+  return $ Coq.Qualid qualid
+convertLiftedExpr (LIR.Var _ _ _ qualid) = return $ Coq.Qualid qualid
+convertLiftedExpr (LIR.App _ func typeArgs effects args freeArgs) = do
+  func' : args' <- mapM convertLiftedExpr $ func : args
+  typeArgs'     <- mapM convertLiftedType typeArgs
+  let effectArgs' =
+        [ Coq.Qualid (fst Coq.Base.partialArg) | Partiality `elem` effects ]
```
just95: I'd suggest to add a function `convertEffect` that performs pattern matching on `Effect` such that GHC produces a warning if a new effect is added that is not handled by the back end.

src/lib/FreeC/Backend/Coq/Converter/Free.hs
```diff
@@ -43,9 +43,20 @@ genericApply
...
+genericApply func = genericApply' $ Coq.Qualid func
+
+-- | Like 'genericApply' but takes a function or (type) constructor term instead
+--   of a qualified identifier as its first argument.
+genericApply'
+  :: Coq.Term -- ^ The function or (type) constructor term
+  -> [Coq.Term] -- ^ The type class instances to pass to the callee.
+  -> [Coq.Term] -- ^ Implicit arguments to pass explicitly to the callee.
+  -> [Coq.Term] -- ^ The actual arguments of the callee.
+  -> Coq.Term
```
just95:
```suggestion
  :: Coq.Term   -- ^ The function or (type) constructor term
  -> [Coq.Term] -- ^ The type class instances to pass to the callee.
  -> [Coq.Term] -- ^ Implicit arguments to pass explicitly to the callee.
  -> [Coq.Term] -- ^ The actual arguments of the callee.
  -> Coq.Term
```

src/lib/FreeC/LiftedIR/Syntax/Expr.hs
```diff
@@ -35,6 +37,7 @@ data Expr
         , exprAppTypeArgs :: [Type]   -- ^ Visible type applications.
         , exprEffects     :: [Effect] -- ^ Effect set.
         , exprAppArgs     :: [Expr]   -- ^ Applied arguments.
+        , exprFreeArgs    :: Bool     -- ^ If the arguments for the free monad are needed.
```
just95:
```suggestion
        , exprFreeArgs    :: Bool     -- ^ Whether the arguments for the free
                                      --   monad are needed.
```

src/lib/FreeC/LiftedIR/Syntax/Expr.hs
```diff
@@ -54,16 +57,20 @@ data Expr
...
               }
 
   | -- | An integer literal.
     IntLiteral { exprSrcSpan     :: SrcSpan
                , intLiteralValue :: Integer
                }
 
+  | -- | A string literal.
+    StringLiteral { exprSrcSpan :: SrcSpan
+                  , stringValue :: String
```
just95: Since the `intLiteralValue` field has `Literal` in the name, I'd do the same for string literals.

```suggestion
                  , stringLiteralValue :: String
```

src/lib/FreeC/LiftedIR/Converter/Expr.hs
```diff
@@ -205,35 +318,42 @@ generateApply mf (a : as) = mf `bind` \f -> generateApply (f `app` a) as
...
-bind arg              k = localEnv $ do
+bind
+  :: LIR.Expr
+  -> Maybe LIR.Type
+  -> (LIR.Expr -> Converter LIR.Expr)
+  -> Converter LIR.Expr
+bind (LIR.Pure _ arg) _       k = k arg -- We don't have to unwrap pure values.
+bind arg              argType k = localEnv $ do
   -- Generate a new name for lambda argument.
   argIdent <- freshIRQName $ fromMaybe "f" (guessName arg)
```
just95: I must have missed this in a previous review but the prefixes of fresh variables should are [defined in `FreeC.Environment.Fresh`](https://github.com/FreeProving/free-compiler/blob/8779013877645b9547bb36c93fa1df9856b2b1aa/src/lib/FreeC/Environment/Fresh.hs#L46) and shouldn't be hard coded. The prefix `f` for example should only be used in applications whereas `cond` should be used by default in `if` expressions. Can you add an additional argument to `bind` and for the right `fresh*Prefix`?

just95 (changes requested):

just95 (approved):

nbun (approved):
Good job!

PR #152, opened 2020-07-24
Fixes dev agda

JonasHoefer:
Merge `main` into `dev-agda` to resolve conflicts


just95 (approved):

MajaRet (changes requested):
This branch contains the classes `Partial` and `Injectable` twice - once in the `Class` folder and once in the `Free` folder. Only the ones in the `Class` folder should exist.

Removing them leads to one error: `Integer` in the Coq prelude imports `Free.Partial`, which no longer exists. Since it also imports `Free` (which exports all type classes) the import should simply be removed.

Other than that, I don't see any issues.

MajaRet (approved):

PR #156, opened 2020-07-27
Let the compiler generate induction schemes

### LINKED ISSUE #154: Generate induction schemes
## Description

The default induction scheme generated by Coq for each data type produced by our compiler does not include any induction hypotheses and are therefore worthless. Our compiler could generate a custom induction scheme for every generated data type.

## Motivation

The current approach is to define a custom induction scheme by hand for any data type when it is needed. The definitions and proofs of those schemes do always look the same and could be generated automatically. Therefore the compiler should take care of that.

MarvinLira:
### Issue

Closes #154

### Description of the Change

The translation process for data types has been modified such that an induction scheme and also a proof for that scheme is generated.
The proof relies on an additional tactic that has been added to the `base` library.

### Alternate Designs

Originally the induction schemes were defined as `Fixpoint`. Unfortunately our Coq representation does not allow to append a `Proof` to that `Fixpoint` definition and a missing proof would make the resulting file not compilable via Coq. Therefore I defined the induction scheme as `Definition` and used the low level `fix` tactic in the proof.

Instead of outsourcing the whole proof for an induction scheme to the new tactic `prove_ind`, one could think about creating a proof without that tactic to make the proof more visible. This would however complicate the creation of that proof in the compiler and the user is most likely only interested in using the induction scheme and not in the proof for the scheme.

### Verification Process

In the creation process I tested my implementation with the `List` and `Pair` data types as well as the `Expr` data type from Huttons Razor.

The existing tests for the data type translation are already quiet good in testing the induction scheme generation too. Nonetheless I created an additional test case where I think the induction scheme generation is hard.


src/lib/FreeC/Backend/Coq/Base.hs
```diff
@@ -52,6 +52,10 @@ freeArgs =
     )
   ]
 
+-- | The Coq identifier for the @ForFree@ property.
+forFree :: Coq.Qualid
+forFree = Coq.bare "ForFree"
+
```
just95, line 58: Since `ForFree` is now used in the generated code, user defined data types and constructors with the name `ForFree` must be renamed. Thus, `ForFree` should be added to the list of `reservedIdents`.

src/lib/FreeC/Backend/Coq/Converter/Free.hs
```diff
@@ -53,6 +53,12 @@ genericApply func effectArgs implicitArgs args
   allArgs :: [Coq.Term]
   allArgs = genericArgs ++ effectArgs ++ implicitArgs ++ args
 
+-- | Smart constructor for a @ForFree@ statement with a given type, property
+--   and @Free@ value.
+genericForFree :: Coq.Term -> Coq.Qualid -> Coq.Qualid -> Coq.Term
+genericForFree tipe prop fv =
+  genericApply Coq.Base.forFree [] [] [tipe, Coq.Qualid prop, Coq.Qualid fv]
```
just95: Avoid misspelling keywords in order to make them valid identifiers.

```suggestion
genericForFree :: Coq.Term -> Coq.Qualid -> Coq.Qualid -> Coq.Term
genericForFree typeTerm prop fv =
  genericApply Coq.Base.forFree [] [] [typeTerm, Coq.Qualid prop, Coq.Qualid fv]
```

src/lib/FreeC/Backend/Coq/Converter/Arg.hs
```diff
@@ -30,11 +30,30 @@ convertTypeVarDecls
...
+--   type arguments and returns also the bound variables.
+--
+--   E.g. the declaration of the type variable @a@ in @data D a = ...@ is
+--   translated to the binder @(a : Type)@. If there are multiple type variable
+--   declarations as in @data D a b = ...@ they are grouped into a single
+--   binder @(a b : Type)@ because we assume all Haskell type variables to be
+--   of kind @*@.
+--
+--   The first argument controls whether the generated binders are explicit
+--   (e.g. @(a : Type)@) or implicit (e.g. @{a : Type}@).
```
just95: Since "and returns also the bound variables" is the only difference to the comment of `convertTypeVarDecls`, I'd suggest to write the following instead.

```suggestion
-- | Like 'convertTypeVarDecls' but also returns the Coq identifiers of the
--   bound variables.
```

src/lib/FreeC/Backend/Coq/Converter/TypeDecl.hs
```diff
@@ -218,6 +224,71 @@ convertDataDecl (IR.DataDecl _ (IR.DeclIdent _ name) typeVarDecls conDecls) =
...
+    (propIdent , propBinder ) <- generateArg
+      (Coq.Arrow (genericApply tIdent [] [] (map Coq.Qualid tvarIdents))
+                 (Coq.Sort Coq.Prop)
+      )
+    (_hIdents, hBinders) <- mapAndUnzipM (generateInductionCase propIdent)
+                                         conDecls
+    (valIdent, valBinder) <- generateArg
+      (genericApply tIdent [] [] (map Coq.Qualid tvarIdents))
+    -- Stick everything together.
+    let schemeName = Coq.bare $ fromJust (Coq.unpackQualid tIdent) ++ "_Ind"
```
just95: Manually generating a new Coq identifier is not safe. Use `freshCoqQualid` instead.

src/lib/FreeC/Backend/Coq/Converter/TypeDecl.hs
```diff
@@ -218,6 +224,71 @@ convertDataDecl (IR.DataDecl _ (IR.DeclIdent _ name) typeVarDecls conDecls) =
...
+        binders =
+          genericArgDecls Coq.Explicit
+            ++ tvarBinders
+            ++ [propBinder]
+            ++ hBinders
+        term = Coq.Forall
+          (NonEmpty.fromList [valBinder])
+          (Coq.app (Coq.Qualid propIdent) [Coq.Qualid valIdent])
+        scheme = Coq.Assertion Coq.Definition schemeName binders term
+        proof  = Coq.ProofDefined (Text.pack "  fix H 1; intro; prove_ind.")
```
just95: If tactics like `prove_ind` can be shadowed by other functions, the identifier must be reserved as well. Any way there should be a constant for the name of the tactic in `FreeC.Backend.Coq.Base`.

src/lib/FreeC/Backend/Coq/Converter/TypeDecl.hs
MarvinLira: The function `freshCoqQualid` seems to be appending always an underscore and a value to that name. As the `schemeName` is used by the user in the proofs it would be nice if the name was untouched unless it is neccessary. Would `renameQualid` be an option here?

just95: Unfortunatly, `renameQualid` does not suffice since it doesn't insert an environment entry. However, `renameQualid` is used by `freshCoqQualid` internally. `freshCoqQualid` generates a Haskell name of the form `Prefix@Index`, uses `renameQualid` to convert it to a Coq identifier of the form `Prefix_Index` and finally inserts an environment entry such that the Coq identifier is not used again. The `@Index` postfix is only needed to ensure that the Haskell identifier (that identifies the environment entry) does not collide with a user defined identifier. You could try to strip the `@Index` suffix from identifiers passed to `renameQualid`. [`freshHaskellIdent`](https://github.com/FreeProving/free-compiler/blob/8779013877645b9547bb36c93fa1df9856b2b1aa/src/lib/FreeC/Environment/Fresh.hs#L86) does something similar. I think that this should be safe but you may have to update a lot of test cases.

MarvinLira: I changed `renameIdent` the way you suggested it for `renameQualid` because `renameQualid` calls this function and they should produce the same names.

src/lib/FreeC/Backend/Coq/Converter/TypeDecl.hs
MarvinLira: It seems like tactics have a different namespace than functions, lemmas, etc. Therefore it seems not to be necessary to reserve the identifier.

src/lib/FreeC/Environment/Renamer.hs
```diff
@@ -142,13 +143,19 @@ sanitizeIdent (firstChar : subsequentChars) =
 
 -- | Renames a Haskell identifier such that it can be safely used in Coq.
 --
+--
```
just95:
```suggestion
--
```

just95 (approved):

MajaRet:
I'm not sure the generated induction schemes work for data types with 'hidden' recursion. 
For example, consider the following tree data type:
```haskell
data TreeP = LeafP Integer | BranchP (TreeP, TreeP)
```

The generated induction scheme looks like this:

```coq
Definition TreeP_Ind (Shape : Type)
(Pos : Shape -> Type)
(a : Type)
(x : TreeP Shape Pos a -> Prop)
(x0 : x emptyP)
(x2 : forall (x1 : Free Shape Pos a), x (leafP x1))
(x4
  : forall (x3
      : Free Shape Pos (Pair Shape Pos (TreeP Shape Pos a) (TreeP Shape Pos a))),
    x (branchP x3)) : forall (x5 : TreeP Shape Pos a), x x5.
Proof.
  fix H 1; intro; prove_ind.
Defined.
```

So there is no induction hypothesis we can use on the `TreeP`s in the pair.

MarvinLira:
The problem with 'hidden' recursion, mentioned by @MajaRet, has been moved to a new issue #159.
This solution here will stick with induction schemes for direct recursion as they are already an improvement to the default induction schemes generated by Coq and they do not result in invalid code for 'hidden' recursion.

MajaRet (approved):

PR #166, opened 2020-08-16
Switch to Floskell

just95:
### Issue

See <https://github.com/FreeProving/guidelines/issues/5>.

### Description of the Change

 - [x] Add a Floskell configuration file based on the template in the [`guidelines`](https://github.com/FreeProving/guidelines) repository. Adapt the configuration file based on the repository's needs (e.g., language extensions and prefixes for grouping of imports). Do not change formatting options. Code style should be consistent across all repositories. 
 - [x] Update the `./tool/check-formatting.sh` and `./tool/format-code.sh` (and in case of the Free Compiler `./tool/full-test.sh`) scripts to use Floskell. Preferably copy an updated version from another repository.
 - [x] Update the configuration of the CI pipeline. Preferably copy updated CI jobs from another repository.
 - [x] Format code using Floskell.

### Verification Process

 
 - [x] Check the code for errors or odd formatting introduced by Floskell. Adapt the configuration file if necessary.

### Additional Notes

 - When a new infix operator is used, its fixity must be added to the `floskell.json` file.
 - Floskell currently removes blank lines after comments. Thus, comments of the form

   ```haskell
   -- Heading                                                                   --
   ```

   are directly followed by the Haddock comment of the next function declaration.

   ```haskell
   -- Heading                                                                   --
   -- | Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed rutrum facilisis nisi,
   --   eget dictum metus tempus vitae.
   ```


src/lib/FreeC/IR/SrcSpan.hs
```diff
@@ -70,23 +68,22 @@ lookupSrcFile = lookup
...
-    , srcSpanEndLine     :: Int      -- ^ The number of the last spanned line.
-    , srcSpanEndColumn   :: Int      -- ^ The offset within the last line.
-    , srcSpanCodeLines   :: [String] -- ^ The source code of the spanned lines.
-    }
+      { srcSpanFilename    :: String   -- ^ The name of the file.
+      , srcSpanStartLine   :: Int      -- ^ The number of the first line.
+      , srcSpanStartColumn :: Int      -- ^ The offset within the first line.
+      , srcSpanEndLine     :: Int      -- ^ The number of the last line.
+      , srcSpanEndColumn   :: Int      -- ^ The offset within the last line.
+      , srcSpanCodeLines   :: [String] -- ^ The spanned line's source code.
```
stu204767:
```suggestion
      , srcSpanCodeLines   :: [String] -- ^ The spanned lines' source code.
```

src/lib/FreeC/IR/DependencyGraph.hs
```diff
@@ -98,143 +93,146 @@ type DGEntry node = (node, DGKey, [DGKey])
...
 -- | Gets the entries of the given dependency graph.
-dependencyGraphEntries :: DependencyGraph node -> [DGEntry node]
-dependencyGraphEntries (DependencyGraph graph getEntry _) =
-  map getEntry (vertices graph)
+dgEntries :: DependencyGraph node -> [DGEntry node]
+dgEntries = liftM2 map dgGetEntry dgVertecies
+
+-- | Gets the vertices of the given dependency graph.
+dgVertecies :: DependencyGraph node -> [Vertex]
+dgVertecies = vertices . dgGraph
```
stu204767:
```suggestion
dgEntries = liftM2 map dgGetEntry dgVertices

-- | Gets the vertices of the given dependency graph.
dgVertices :: DependencyGraph node -> [Vertex]
dgVertices = vertices . dgGraph
```

src/lib/FreeC/IR/DependencyGraph.hs
```diff
@@ -303,54 +299,50 @@ groupModules = dependencyComponents . moduleDependencyGraph
...
-  -> DependencyComponent decl'
-mapComponent f (NonRecursive decl) =
-  let [decl'] = f [decl] in NonRecursive decl'
-mapComponent f (Recursive decls) = Recursive (f decls)
+mapComponent :: ([decl] -> [decl'])
+             -> DependencyComponent decl
+             -> DependencyComponent decl'
+mapComponent f (NonRecursive decl)
+  = let [decl'] = f [decl] in NonRecursive decl'
+mapComponent f (Recursive decls)   = Recursive (f decls)
```
stu204767, line 312: Here, the `=` is aligned with non-existing `=` two lines above. This is a bug, in my opinion, which could be reported to the Floskell developers.

src/test/FreeC/Environment/RenamerTests.hs
```diff
@@ -56,39 +53,47 @@ genVernacularCommand :: Gen String
...
+    it "defined identifiers must be renamed"
+      $ do
+        property
+          $ forAll genIdent
+          $ \ident ->
+          let env = addEntry TypeVarEntry
+                { entrySrcSpan   = NoSrcSpan
+                , entryName      = IR.UnQual (IR.Ident ident)
+                , entryIdent     = Coq.bare ident
+                , entryAgdaIdent = undefined -- ignore Agda identifiers for the moment - TODO: add Unit Tests for Agda renamer!
```
MajaRet: This line is 109 characters long with indentation (93 without), so Floskell doesn't appear to split up trailing comments that are too long.

src/lib/FreeC/Pretty.hs
```diff
@@ -80,23 +74,22 @@ instance Pretty a => Pretty (TrailingLine a) where
...
 --   which @80@ are allowed to be non-indentation characters.
 renderPretty' :: Pretty a => a -> SimpleDoc
 renderPretty' = renderPretty ribbonFrac maxLineWidth . pretty
  where
-  ribbonWidth, maxLineWidth :: Int
-  ribbonWidth  = 80
-  maxLineWidth = 120
+   ribbonWidth, maxLineWidth :: Int
+   ribbonWidth = 80
+
```
MajaRet, line 84: I think the empty line looks rather odd in this case, where the local definitions are just constants (and even share their type signature).

just95: There is an option to remove blank lines in `where`-blocks, but that would also remove the blank lines in `where`-blocks that contain more complex functions. I suggest to move the type signature instead.

```haskell
 where
   ribbonWidth :: Int
   ribbonWidth = 80

   maxLineWidth :: Int
   maxLineWidth = 120
```

src/test/FreeC/Environment/RenamerTests.hs
just95:
```suggestion
                , entryAgdaIdent = undefined -- TODO add Agda renamer tests.
```

src/lib/FreeC/IR/Subterm.hs
```diff
@@ -101,91 +94,80 @@ class Pretty a => Subterm a where
...
+  childTerms (IR.TypeAppExpr _ expr _ _) = [expr]
+  childTerms (IR.If _ e1 e2 e3 _) = [e1, e2, e3]
+  childTerms (IR.Case _ expr alts _) = expr : map IR.altRhs alts
+  childTerms (IR.Lambda _ _ expr _) = [expr]
+  childTerms (IR.Con _ _ _) = []
+  childTerms (IR.Var _ _ _) = []
+  childTerms (IR.Undefined _ _) = []
+  childTerms (IR.ErrorExpr _ _ _) = []
+  childTerms (IR.IntLiteral _ _ _) = []
+  childTerms (IR.Let _ binds e _) = e : map IR.bindExpr binds
```
stu204767: Why are the right-hand sides not aligned here?

src/lib/FreeC/IR/Unification.hs
```diff
@@ -44,38 +38,35 @@ data UnificationError
...
+    ++ "`."
+  RigidTypeVarError xSrcSpan x u -> reportFatal
+    $ Message srcSpan Error
+    $ "Could not match rigid type variable '"
+    ++ x
+    ++ "' (bound at '"
+    ++ showPretty xSrcSpan
+    ++ "') with type '"
+    ++ showPretty u
+    ++ "'."
```
stu204767: This has nothing to do with the switch to Floskell, but I think that ` should be used here instead of ', just as in the other cases.
```suggestion
    $ "Could not match rigid type variable `"
    ++ x
    ++ "` (bound at `"
    ++ showPretty xSrcSpan
    ++ "`) with type `"
    ++ showPretty u
    ++ "`."
```

src/lib/FreeC/Pass/DefineDeclPass.hs
```diff
@@ -24,19 +24,14 @@
 --   = Error cases
 --
 --   * The user is informed if a different name is assigned to an entry.
+module FreeC.Pass.DefineDeclPass ( defineTypeDeclsPass, defineFuncDeclsPass ) where
```
stu204767, line 27: This line is 84 characters long, this looks like a bug in Floskell to me.

src/lib/FreeC/Pass/CompletePatternPass.hs
```diff
@@ -94,59 +93,54 @@ completePatternPass ast = do
...
+     ++ showPretty (IR.funcDeclName funcDecl)
 
-  -- The type of the scrutinee shouldn't be function or a type var
-  getTypeConName IR.TypeVar{}                = Nothing
-  getTypeConName IR.FuncType{}               = Nothing
+   -- | Selects the name of the outermost type constructor from a type
+   getTypeConName :: IR.Type -> Maybe IR.TypeConName
+   getTypeConName (IR.TypeCon _ typeConName) = Just typeConName
+   getTypeConName (IR.TypeApp _ typeAppLhs _) = getTypeConName typeAppLhs
+   -- The type of the scrutinee shouldn't be function or a type var
```
stu204767:
```suggestion
   -- The type of the scrutinee shouldn't be a function or type var
```

stu204767 (changes requested):

src/lib/FreeC/Pass/DefineDeclPass.hs
just95: That's not really a bug. Floskell does not have a hard line limit but a penalty function. Seems like I have to increase the penalty for overfull lines or reduce the penalty for line breaks.

src/lib/FreeC/IR/Subterm.hs
just95: This is because Floskell would have to insert too many spaces in order to align the matches (it is currently configured to insert a padding of at most 10 spaces). This has now been mentioned multiple times (see <https://github.com/FreeProving/language-coq/pull/5#discussion_r472385599> and <https://github.com/FreeProving/haskell-src-transformations/pull/44#discussion_r472861300>). However for the reason I stated in <https://github.com/FreeProving/haskell-src-transformations/pull/44#discussion_r472946438>, I'm not sure whether we should actually increase the alignment limit. Maybe we should even disable alignment of function rules to prevent confusion... :thinking:

src/lib/FreeC/Pass/DefineDeclPass.hs
just95: After playing around with the penalties a bit more, it seems like they are not at fault in this case. It seems like Floskell does not consider a line overfull if the `where` keyword starts before the 80 character limit.

src/lib/FreeC/IR/Subterm.hs
stu204767: So, if I understand that correctly, the main problem is that Floskell doesn't try to group matches and align the right-hand sides differently for each group, like Brittany does, but either aligns all right-hand sides to the same position or doesn't align any right-hand sides.
I think this kind of alignment is very useful in places like this one, so we should increase this padding in the configuration. I would suggest a maximum of 20 to 30 spaces, unless that somehow causes other places to look weird.

src/lib/FreeC/Pass/DefineDeclPass.hs
stu204767: That's what I thought to have caused this long line. I would consider that to be a bug.

src/lib/FreeC/IR/DependencyGraph.hs
just95: Yes, this is definitely a bug and we should report it.

In this particular case, the formatting of this function has been fixed by another change (see 69810761aa38500fe54fab212fcd02a3d9c60892) that was needed for the `haskell-src-transformations` repository.

```haskell
mapComponent f (NonRecursive decl) = let [decl'] = f [decl]
                                     in NonRecursive decl'
mapComponent f (Recursive decls)   = Recursive (f decls)
```

src/lib/FreeC/Pass/TypeInferencePass.hs
```diff
@@ -249,16 +233,15 @@ data TypeEquation = TypeEquation
 --
 --    Rigid type variables in type equations cannot be substituted.
```
Daniel-Teut:
```suggestion
--   Rigid type variables in type equations cannot be substituted.
```

src/lib/FreeC/IR/Subterm.hs
just95: > So, if I understand that correctly, the main problem is that Floskell doesn't try to group matches and align the right-hand sides differently for each group, like Brittany does, but either aligns all right-hand sides to the same position or doesn't align any right-hand sides.

Exactly.

> I would suggest a maximum of 20 to 30 spaces, unless that somehow causes other places to look weird.

The example in <https://github.com/FreeProving/haskell-src-transformations/pull/44#discussion_r472946438> uses a limit of 20 and it kind of looks weird to me. Nonetheless, I've now increased the limit to 20 spaces (see 3eb9c0e90238806daddb048c0ce51891a0c0125e) such that we can see how that would impact the overall readability of the code.

src/lib/FreeC/Pass/TypeInferencePass.hs
```diff
@@ -452,7 +431,6 @@ inferFuncDeclTypes' funcDecls = withLocalState $ do
     -- This is required such that polymorphically recursive functions don't
     -- cause a type error.
```
Daniel-Teut:
```suggestion
  -- This is required such that polymorphically recursive functions don't
  -- cause a type error.
```
The comments are also misaligned on the two lines above, but i can't comment on them for some reason.

src/lib/FreeC/Frontend/Haskell/Simplifier.hs
```diff
@@ -737,132 +666,127 @@ simplifyExpr pragma@(HSE.SCCPragma _ _ expr) = do
...
+  = notSupported "Unboxed tuples" name
+simplifyConName name@(HSE.Special _ (HSE.UnboxedSingleCon _))
+  = notSupported "Unboxed tuples" name
+simplifyConName name@(HSE.Special _ (HSE.ExprHole _))
+  = notSupported "Expression holes" name
 
 -- | Simplifies a variable pattern (e.g. the parameters of a lambda abstraction
 --   or function declaration).
 --
 --  Parenthesis are ignored.
```
Daniel-Teut:
```suggestion
--   Parenthesis are ignored.
```

Daniel-Teut (commented):
While not directly related to Floskell, I found some misaligned comments.

src/lib/FreeC/Pass/TypeInferencePass.hs
just95: Fixed by 7208d091314bd736fa3fed8a2f2bc7aac095085e.

maclement (approved):

stu204767 (approved):

Daniel-Teut (approved):

MajaRet (approved):

PR #169, opened 2020-08-24
Implement a pass for expression-flattening and annotation of Sharing

### LINKED ISSUE #76: Add a pass that annotates the shared expressions in the IR
## Description

A new compiler pass should be added that inserts information about shared expressions into the IR.

## Motivation

A further step towards the goal of modelling Haskell's sharing is the implementation of a new compiler pass that checks the modules for shared expressions and annotates this in in the IR.  

## Additional Context
This pass uses annotations for shared expressions described in #75.

maclement:
The pass for the annotation counts all variables on right hand sides of function declarations and introduces `let` expressions that makes the sharing of variables explicit.

For example the function
```
twice :: Integer -> Integer
twice x = x + x
```
is transformed into

```
twice :: Integer -> Integer
twice x = let x@0 = x in x@0 + x@0
```
where `x@0` is a fresh variable that is also inserted into the environment. 

The Pass for expression-flattening transforms each function into a "flat" form.
An expression is "flattened" if each function `f` is only applied onto variables or constants.

The following function
```
fun f g x = f (g x)
```
is translated into
```
fun f g x = let z = g x in f z
```
where `z` is a fresh variable.

closes #76


src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,143 @@
...
+--
+--   = Specification
+--
+--   == Preconditions
+--
+--   There are no special requirements.
+--
+--   == Translation
+--
+--   All shared variables on right hand sides of function declarations are made
```
just95:
```suggestion
--   All shared variables on right-hand sides of function declarations are made
```

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,143 @@
+-- | This module contains a compiler pass that analyses the right hand sides of
+--   function declaration and introduces @let@ expression with new variables
+--   for each variable that occurs more than once on the right hand sides.
```
just95:
```suggestion
-- | This module contains a compiler pass that analyses the right-hand sides of
--   function declaration and introduces @let@ expression with new variables
--   for each variable that occurs more than once on the right-hand sides.
```

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,143 @@
...
+import           FreeC.Environment.Fresh   ( freshHaskellQName )
+import           FreeC.Environment.Renamer ( renameAndDefineVar )
+import qualified FreeC.IR.Syntax        as IR
+import           FreeC.IR.Subst
+import           FreeC.IR.SrcSpan
+import           FreeC.Monad.Converter
+import           FreeC.Pass
+
+-- | Checks all function declarations if they contain variables that occur
+--   multiple times on the same right hand side.
```
just95:
```suggestion
-- | Checks all function declarations if they contain variables that occur
--   multiple times on the same right-hand side.
```

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,143 @@
...
+--   introduced variable.
+sharingAnaylsisPass :: Pass IR.Module
+sharingAnaylsisPass
+    (IR.Module srcSpan name imports typeDecls typeSigs pragmas funcDecls) = do
+        funcDecls' <- mapM analyseSharingDecl (funcDecls)
+        return (IR.Module srcSpan name imports typeDecls typeSigs pragmas
+                funcDecls')
+
+-- | Checks a function declaration if it contains a variable that occurs
+--   multiple times on the right hand side.
```
just95:
```suggestion
-- | Checks a function declaration if it contains a variable that occurs
--   multiple times on the right-hand side.
```

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,143 @@
...
+            -- This type scheme has wrong srcSpanInformation
+            rhs = IR.Var srcSpan varName varTypeScheme
+            Just varPatIdent = IR.identFromQName varName'
+            varPatType = fmap IR.typeSchemeType varTypeScheme
+            varPat = IR.VarPat srcSpan varPatIdent varPatType False
+            bind = IR.Bind srcSpan varPat rhs
+        _ <- renameAndDefineVar srcSpan False varPatIdent varPatType
+        return ( bind : binds, subst : substs )
+
+-- | Counts all variable names on right hand sides of expression.
```
just95:
```suggestion
-- | Counts all variable names on right-hand sides of expression.
--
```

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,143 @@
...
+--   >   ; Just x -> Just (x + x)
+--   >   }
+--
+--   Should be transformed into
+--
+--   > twiceMaybe :: Maybe Integer -> Maybe Integer
+--   > twiceMaybe ( mx :: Maybe Integer ) = case ( mx :: Maybe Integer ) of
+--   >   { Nothing -> Nothing
+--   >   ; Just x -> let y = x in Just ( y + y )
+--   >   }
```
just95: The following formatting is more consistent with the documentation of the other passes. I don't think that the definition of the `Maybe` data type needs to be recapitulated.

```suggestion
--   = Examples
--
--   == Example 1
--
--   > twice :: Integer -> Integer
--   > twice (x :: Integer) = x + x
--
--   Should be transformed into
--
--   > twice :: Integer -> Maybe Integer
--   > twice (x :: Integer) = let (y :: Integer) = x in y + y
--
--   == Example 2
--
--   > twiceMaybe :: Maybe Integer -> Maybe Integer
--   > twiceMaybe (mx :: Maybe Integer) = case (mx :: Maybe Integer) of {
--   >     Nothing -> Nothing;
--   >     Just x  -> Just (x + x)
--   >   }
--
--   Should be transformed into
--
--   > twiceMaybe :: Maybe Integer -> Maybe Integer
--   > twiceMaybe (mx :: Maybe Integer) = case (mx :: Maybe Integer) of {
--   >     Nothing -> Nothing;
--   >     Just x  -> let y = x in Just (y + y)
--   >   }
```

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,143 @@
...
+import           FreeC.Environment.Renamer ( renameAndDefineVar )
+import qualified FreeC.IR.Syntax        as IR
+import           FreeC.IR.Subst
+import           FreeC.IR.SrcSpan
+import           FreeC.Monad.Converter
+import           FreeC.Pass
+
+-- | Checks all function declarations if they contain variables that occur
+--   multiple times on the same right hand side.
+--   If that is the case a @let@-expression is introduced that binds the
```
just95:
```suggestion
--   If that is the case, a @let@-expression is introduced that binds the
```

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,143 @@
...
+sharingAnaylsisPass :: Pass IR.Module
+sharingAnaylsisPass
+    (IR.Module srcSpan name imports typeDecls typeSigs pragmas funcDecls) = do
+        funcDecls' <- mapM analyseSharingDecl (funcDecls)
+        return (IR.Module srcSpan name imports typeDecls typeSigs pragmas
+                funcDecls')
+
+-- | Checks a function declaration if it contains a variable that occurs
+--   multiple times on the right hand side.
+--   If that is the case a @let@-expression is introduced.
```
just95:
```suggestion
--   If that is the case, a @let@-expression is introduced.
```

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,143 @@
...
+analyseSharingDecl :: IR.FuncDecl -> Converter IR.FuncDecl
+analyseSharingDecl (IR.FuncDecl srcSpan ident typeArgs args returnType rhs) = do
+    let varList = (map (\p -> ( fst p, snd $ snd p )) . filter
+                   ((> 1) . fst . snd) . Map.toList . countVarNames) rhs
+    rhs' <- buildLet rhs varList
+    return (IR.FuncDecl srcSpan ident typeArgs args returnType rhs')
+
+-- | Builds a @let@-expression from the given expression and variable names.
+--   Computes let-bindings from the given variables, composes the resulting
```
just95:
```suggestion
--
--   Computes @let@-bindings from the given variables, composes the resulting
```

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,143 @@
...
+buildLet expr []   = return expr
+buildLet expr vars = do
+    let srcSpan = IR.exprSrcSpan expr
+    ( binds, substs ) <- buildBinds srcSpan vars
+    return (IR.Let srcSpan binds (applySubst (composeSubsts substs) expr)
+            (IR.exprTypeScheme expr))
+
+-- | Converts the list containing variables into let-bindings where
+--   the variable pattern is a fresh variable and the right hand side is
+--   a variable that occured multiple times on right hand sides.
```
just95:
```suggestion
-- | Converts the list containing variables into @let@-bindings where
--   the variable pattern is a fresh variable and the right-hand side is
--   a variable that occurred multiple times on right hand sides.
--
```

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,144 @@
...
+--
+--   Should be transformed into
+--
+--   > twiceMaybe :: Maybe Integer -> Maybe Integer
+--   > twiceMaybe (mx :: Maybe Integer) = case (mx :: Maybe Integer) of {
+--   >     Nothing -> Nothing;
+--   >     Just x  -> let y = x in Just (y + y)
+--   >   }
+--
+--
```
just95: Maybe add something like

```suggestion
--
-- where @y@ is a fresh variable.
--
```

here and in the example above.

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,144 @@
...
+--   = Specification
+--
+--   == Preconditions
+--
+--   There are no special requirements.
+--
+--   == Translation
+--
+--   All shared variables on right-hand sides of function declarations are made
+--   explicit by introducing @let@-expressions.
```
just95: This sounds more like the postcondition to me.
The *Translation* section should provide details for how you decide whether a variable is considered shared or not and where exactly `let`-expressions are introduced.
For instance, if Example 2 is correct, don't forget to mention that `let`-expressions are added for shared variables from `case`-expression alternatives as well.
I was also wondering whether the arguments of lambda expressions don't have to be shared explicitly as well.

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,144 @@
+-- | This module contains a compiler pass that analyses the right-hand sides of
+--   function declaration and introduces @let@ expression with new variables
```
just95:
```suggestion
--   function declaration and introduces @let@-expression with new variables
```

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,144 @@
...
+import qualified FreeC.IR.Syntax        as IR
+import           FreeC.IR.Subst
+import           FreeC.IR.SrcSpan
+import           FreeC.Monad.Converter
+import           FreeC.Pass
+
+-- | Checks all function declarations if they contain variables that occur
+--   multiple times on the same right-hand side.
+--   If that is the case, a @let@-expression is introduced that binds the
+--   variables to a fresh ones and replaces the occurrences with the newly
```
just95:
```suggestion
--   variables to fresh ones and replaces the occurrences with the newly
```

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,144 @@
...
+--   introduced variable.
+sharingAnaylsisPass :: Pass IR.Module
+sharingAnaylsisPass
+    (IR.Module srcSpan name imports typeDecls typeSigs pragmas funcDecls) = do
+        funcDecls' <- mapM analyseSharingDecl (funcDecls)
+        return (IR.Module srcSpan name imports typeDecls typeSigs pragmas
+                funcDecls')
```
just95: This can be shortened using record syntax.

```suggestion
sharingAnaylsisPass ast = do
  funcDecls' <- mapM analyseSharingDecl (IR.modFuncDecls ast)
  return ast { IR.modFuncDecls = funcDecls' }
```

src/lib/FreeC/Pipeline.hs
```diff
@@ -38,6 +39,7 @@ pipeline =
     [typeInferencePass, defineFuncDeclsPass, partialityAnalysisPass]
   , completePatternPass
   , etaConversionPass
+  , sharingAnaylsisPass
```
just95: Why don't we move this pass above the `typeInferencePass` such that we don't have to manage type annotations manually?

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,144 @@
...
+                   ((> 1) . fst . snd) . Map.toList . countVarNames) rhs
+    rhs' <- buildLet rhs varList
+    return (IR.FuncDecl srcSpan ident typeArgs args returnType rhs')
+
+-- | Builds a @let@-expression from the given expression and variable names.
+--
+--   Computes @let@-bindings from the given variables, composes the resulting
+--   substitutions and applies the substitution on the expression.
+buildLet
+    :: IR.Expr -> [ ( IR.VarName, Maybe IR.TypeScheme ) ] -> Converter IR.Expr
```
just95: The formatting suggests that your branch may be based on an an early version of the `floskell` branch. Make sure to merge `main` into your branch to get the latest version of `floskell.json` and reformat the code such that the pipeline does not fail when this branch is merged.

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,144 @@
...
+countVarNames (IR.If _ e1 e2 e3 _) = (countVarNames e1)
+    `mergeMap` (countVarNames e2) `mergeMap` countVarNames e3
+countVarNames (IR.Case _ e alts _) = countVarNames e `mergeMap` foldr mergeMap
+    Map.empty (map (countVarNames . IR.altRhs) alts)
+countVarNames IR.Undefined {} = Map.empty
+countVarNames IR.ErrorExpr {} = Map.empty
+countVarNames IR.IntLiteral {} = Map.empty
+countVarNames (IR.Lambda _ _ rhs _) = countVarNames rhs
+countVarNames (IR.Let _ binds e _) = countVarNames e `mergeMap` foldr mergeMap
+    Map.empty (map (countVarNames . IR.bindExpr) binds)
```
just95: `case`-expressions alternatives, lambda expressions and `let`-bindings can shadow variables. Shadowed variables should probably not be counted.

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,144 @@
...
+--   into the map as well.
+countVarNames :: IR.Expr -> Map IR.VarName ( Integer, Maybe IR.TypeScheme )
+countVarNames IR.Con {} = Map.empty
+countVarNames (IR.Var _ varName maybeTypeScheme)
+    = Map.singleton varName ( 1, maybeTypeScheme )
+countVarNames (IR.App _ lhs rhs _)
+    = countVarNames lhs `mergeMap` countVarNames rhs
+countVarNames (IR.TypeAppExpr _ lhs _ _) = countVarNames lhs
+countVarNames (IR.If _ e1 e2 e3 _) = (countVarNames e1)
+    `mergeMap` (countVarNames e2) `mergeMap` countVarNames e3
```
just95: Looks like the CI pipeline has not been triggered for some reason. HLint should have complained about the redundant parentheses.

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,144 @@
...
+--   >     Just x  -> Just (x + x)
+--   >   }
+--
+--   Should be transformed into
+--
+--   > twiceMaybe :: Maybe Integer -> Maybe Integer
+--   > twiceMaybe (mx :: Maybe Integer) = case (mx :: Maybe Integer) of {
+--   >     Nothing -> Nothing;
+--   >     Just x  -> let y = x in Just (y + y)
+--   >   }
```
just95: If I understand the code correctly, this is not implemented yet.
However, I'm also not convinced that this is semantically correct. Consider the following program.

```haskell
let m42 = Just (trace "Test" 42) in (twiceMaybe m42, twiceMaybe m42)
```

This program should print "Test" only once but I think that it will be printed twice in the example because `y` is shared only locally.

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,144 @@
...
+    -> Converter ( [ IR.Bind ], [ Subst IR.Expr ] )
+buildBinds srcSpan = foldM buildBind ( [], [] )
+  where
+    buildBind :: ( [ IR.Bind ], [ Subst IR.Expr ] )
+        -> ( IR.VarName, Maybe IR.TypeScheme )
+        -> Converter ( [ IR.Bind ], [ Subst IR.Expr ] )
+    buildBind ( binds, substs ) ( varName, varTypeScheme ) = do
+        varName' <- freshHaskellQName varName
+        let subst = singleSubst' varName (\s -> IR.Var s varName' varTypeScheme)
+            -- This type scheme has wrong srcSpanInformation
```
just95:
```suggestion
            -- This type scheme has wrong source span information.
```

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,144 @@
...
+--
+--   Also computes substitutions mapping given variables to fresh variables.
+--   Returns the generated let-bindings and the substitutions.
+buildBinds :: SrcSpan -> [ ( IR.VarName, Maybe IR.TypeScheme ) ]
+    -> Converter ( [ IR.Bind ], [ Subst IR.Expr ] )
+buildBinds srcSpan = foldM buildBind ( [], [] )
+  where
+    buildBind :: ( [ IR.Bind ], [ Subst IR.Expr ] )
+        -> ( IR.VarName, Maybe IR.TypeScheme )
+        -> Converter ( [ IR.Bind ], [ Subst IR.Expr ] )
```
just95: I think this fold is just [`mapAndUnzipM`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Control-Monad.html#v:mapAndUnzipM) from `Control.Monad`.

```suggestion
buildBinds srcSpan = mapAndUnzipM buildBind
  where
    buildBind
        :: (IR.VarName, Maybe IR.TypeScheme)
        -> Converter (IR.Bind, Subst IR.Expr)
```

just95 (changes requested):

src/lib/FreeC/Pipeline.hs
maclement: That is a good idea. I wasn't sure where to put it.

src/lib/FreeC/Pass/SharingAnalysisPass.hs
maclement: Good catch. I didn't think of that.

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,144 @@
...
+    buildBind ( binds, substs ) ( varName, varTypeScheme ) = do
+        varName' <- freshHaskellQName varName
+        let subst = singleSubst' varName (\s -> IR.Var s varName' varTypeScheme)
+            -- This type scheme has wrong source span information.
+            rhs = IR.Var srcSpan varName varTypeScheme
+            Just varPatIdent = IR.identFromQName varName'
+            varPatType = fmap IR.typeSchemeType varTypeScheme
+            varPat = IR.VarPat srcSpan varPatIdent varPatType False
+            bind = IR.Bind srcSpan varPat rhs
+        _ <- renameAndDefineVar srcSpan False varPatIdent varPatType
```
Daniel-Teut: Currently only lets defined in this pass are added to the environment and lets already present in the source code are not.  In #179 all variables in let bindings are added during the translation to LIR. I think it won't be necessary to add them here.

Daniel-Teut (changes requested):
If the haskell program already contains let expressions for a shared variable, e.g.
```haskell
f x = let x' = x in x' + x' 
```
a new let is still introduced:
```
f x = let { x'@0 = x' } in (let { x'  = x } in ( x' +  x' )) 
```

src/lib/FreeC/Pass/SharingAnalysisPass.hs
maclement: Okay. I was unsure if the variables were added to environment somewhere else.

src/lib/FreeC/Pass/SharingAnalysisPass.hs
macmaccamcam: You are right. I think misunderstood the task here.

maclement: You are right. I overlooked to implement introduction of `let`-expressions on right hand sides of `case`-alternatives.

nbun: @just95 If I understand your example correctly, the `let m42` binding introduces sharing, which is not "overwritten" by the binding in `twiceMaybe`. So it does not make a difference whether there is a `let` clause in the right-hand side of the `Just` alternative.

just95: @nbun Ah, okay, I somehow thought `share` would have no effect on the computation inside `Just`. I've now played around with the Coq implementation and can confirm that "Test" is printed only once in my example.

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,254 @@
...
+--
+--   First all subexpressions of each function declaration are checked for variables
+--   occurring multiple times on the right hand sides of lambda abstractions or
+--   @case@-expression alternatives. If found the variables are replaced by a
+--   fresh variable and a @let@-binding is introduced binding the fresh variable
+--   to the old one.
+--
+--   After all subexpressions are checked the right hand side of the function
+--   declaration is checked as well.
+--   Variables already bound by @let@-bindings are not counted.
```
just95:
```suggestion
--   == Translation
--
--   First all subexpressions of each function declaration are checked for variables
--   occurring multiple times on the right hand sides of lambda abstractions or
--   @case@-expression alternatives. If found the variables are replaced by a
--   fresh variable and a @let@-binding is introduced binding the fresh variable
--   to the old one.
--
--   After all subexpressions are checked the right hand side of the function
--   declaration is checked as well.
--   Variables already bound by @let@-bindings are not counted.
--
--   == Postconditions
--
--   All shared variables on right-hand sides of function declarations are made
--   explicit by introducing @let@-expressions.
```

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,254 @@
...
+--   Returns the generated let-bindings and the substitutions.
+buildBinds :: SrcSpan -> [IR.VarName] -> Converter ([IR.Bind], [Subst IR.Expr])
+buildBinds srcSpan = mapAndUnzipM buildBind
+ where
+  buildBind :: IR.VarName -> Converter (IR.Bind, Subst IR.Expr)
+  buildBind varName = do
+    varName' <- freshHaskellQName varName
+    let subst            = singleSubst' varName
+          (\s -> IR.Var s varName' Nothing)
+        -- This type scheme has wrong source span information.
```
just95: There is no type annotation anymore.

```suggestion
```

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,254 @@
...
+        expr
+  buildLet expr varList
+
+-- | Builds a @let@-expression from the given expression and variable names.
+--
+--   Computes @let@-bindings from the given variables, composes the resulting
+--   substitutions and applies the substitution on the expression.
+buildLet :: IR.Expr -> [IR.VarName] -> Converter IR.Expr
+buildLet expr []   = return expr
+buildLet expr vars = do
```
just95: Due to the usage of `freshHaskellQName` in `buildBinds` the counter for fresh variables is incremented. Since the scope of the fresh variables is delimited by the `let` expression, I think that we can open a local environment such that the counter is reset afterwards.

```suggestion
buildLet expr vars = localEnv $ do
```

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,254 @@
...
+countVarNamesExcept :: [IR.VarName] -> IR.Expr -> Map IR.VarName Integer
+countVarNamesExcept = countVarNamesWith (++) elem
+
+-- | Counts all variable names on right-hand sides of expression that are also
+--   contained by the given list.
+--   Shadowed variables and variables from the list are not counted.
+--   Variables introduced on the left side of a @case@-alternative and @let@
+--   expressions are not counted.
+countVarNamesOnly :: [IR.VarName] -> IR.Expr -> Map IR.VarName Integer
+countVarNamesOnly = countVarNamesWith const notElem
```
just95: The documentation states that

> Shadowed variables and variables from the list are not counted

but actually only variables from the list are counted. Shadowed variables are counted 
 as well since `const` ignores bound variables.

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,254 @@
...
+--   The first arguments is the function that is used to combine
+--   the given list of variable names with variable names from left hand sides
+--   of @case@-alternatives or lambda abstractions.
+--   The second argument is the a predicate that determines if a variable
+--   is added to the map.
+countVarNamesWith :: ([IR.VarName] -> [IR.VarName] -> [IR.VarName])
+                  -> (IR.VarName -> [IR.VarName] -> Bool)
+                  -> [IR.VarName]
+                  -> IR.Expr
+                  -> Map IR.VarName Integer
```
just95: You've overengineered this function a little bit which makes it very hard to understand. The definition gets much easier without predicates and accumulator. In order to implement shadowing you just have to remove entries from the map, for example.

```haskell
countVarNames :: IR.Expr -> Map IR.VarName Integer
countVarNames (IR.Lambda _ args rhs _) 
  = countVarNames rhs `Map.withoutKeys` Set.fromList (map IR.varPatQName args)
```

The intended behavior of `countVarNamesOnly` can then be obtained by restricting the key set of the map.

```haskell
countVarNamesOnly :: [IR.VarName] -> IR.Expr -> Map IR.VarName Integer
countVarNamesOnly varNames expr
  = countVarNames expr `Map.restirctKeys` Set.fromList varNames 
```

This fixes the bug pointed out above automatically.

just95 (changes requested):

src/lib/FreeC/Pass/SharingAnalysisPass.hs
```diff
@@ -0,0 +1,253 @@
...
+  funcDecls' <- mapM analyseSharingDecl (IR.modFuncDecls ast)
+  return ast { IR.modFuncDecls = funcDecls' }
+
+-- | Checks a function declaration for @case@-expressions to introduce local
+--   @let@-expressions and applies the transformation on the right-hand side.
+analyseSharingDecl :: IR.FuncDecl -> Converter IR.FuncDecl
+analyseSharingDecl (IR.FuncDecl srcSpan ident typeArgs args returnType rhs) = do
+  rhs' <- analyseLocalSharing rhs
+  rhs'' <- analyseSharingExpr rhs'
+  return (IR.FuncDecl srcSpan ident typeArgs args returnType rhs'')
```
Daniel-Teut:
```suggestion
analyseSharingDecl funcDecl = do
  rhs' <- (analyseLocalSharing >=> analyseSharingExpr) (IR.funcDeclRhs funcDecl)
  return funcDecl { IR.funcDeclRhs = rhs' }
```
The function can be simplified a bit.

Daniel-Teut (changes requested):

src/lib/FreeC/Pass/SharingAnalysisPass.hs
maclement: That is a really good idea. I did not think of that. 
My intention was to avoid code duplication but that made the function way to complicated.

Daniel-Teut (approved):

Daniel-Teut (approved):

PR #170, opened 2020-08-24
Implement Normalform typeclass

### LINKED ISSUE #148: Implement Normalform typeclass
## Description

Implement a typeclass `Normalform` to pull effects into the outermost layer before handling a program.

## Motivation

TBA

## ToDo

- [ ] Implement a `Normalform` typeclass.
- [ ] Implement `Normalform` instances for pre-defined data types.
- [ ] Normalize programs before handling them in the test module.
- [ ] Add appropriate tests to the sharing handler test suite.

MajaRet:
### Issue

Closes #148 

### Description of the Change

This PR implements a typeclass `Normalform` that allows us to move effects from the components of a data structure to its root. 
For example, the list [0 ? 1, 2] with a non-deterministic element is transformed into the non-deterministic list [0, 2] ? [1, 2]. 
This allows us to handle such programs with effects in the components of data structures.

The class `Normalform` is parameterized over `Shape`, `Pos`, and two types `A` and `B` and features the function `nf'`, which, applied to an `A`, returns a `Free Shape Pos B`. 
The type parameter `B` describes the change of the original type `A` to an effect-free version of the same type.
Thus, for a type `T Shape Pos`, we define the `Normalform` instance `Normalform Shape Pos (T Shape Pos) (T Identity.Shape Identity.Pos)`.

In addition, for any types A and B for which a `Normalform` instance exists, the following functions are defined:
- `nf`, which, applied to a `Free Shape Pos A`, returns a `Free Shape Pos B`. (This is the function that will actually be used by the handlers.)
- `nf_impure`, which is a property that defines how `nf` is moved into the position function of an impure value.
- `nf_pure`, which is a property that defines how `nf` is unfolded on pure values.

### Verification Process

A testing module `NormalizationTests` is included that tests normalization of lists of boolean values. It considers the effects non-determinism, partiality and tracing.
In addition, programs in `SharingHandlerTests` are now normalized before they are evaluated. The tests continue to pass.


base/coq/Free/Verification/NormalizationTests.v
```diff
@@ -0,0 +1,392 @@
...
+From Base Require Import Prelude.
+From Base Require Import Free.Util.Search.
+
+Require Import Lists.List.
+Import List.ListNotations.
+
+(* Shortcuts to handle a program. *)
+
+(* Shortcut to evaluate a non-deterministic program to a result list.
+   list without normalization. *)
```
nbun:
```suggestion
(* Shortcut to evaluate a non-deterministic program to a result list
   without normalization. *)
```

base/coq/Prelude/Bool.v
```diff
@@ -25,3 +26,32 @@ Section SecBool.
...
+
+(* Normalform instance for Bool *)
+
+Section SecNFBool.
+
+Variable Shape : Type.
+Variable Pos : Shape -> Type.
+
+  Definition nfBool (n : Free Shape Pos (Bool Shape Pos)) 
+    := n >>= (fun n' => pure n').
```
MarvinLira: This can be written easier. This also applies to `Integer` and `Unit`.
```suggestion
    := n >>= pure.
```

base/coq/Prelude/Bool.v
```diff
@@ -25,3 +26,32 @@ Section SecBool.
     b1 >>= fun(b1' : Bool') => if b1' then True_ else b2.
 
 End SecBool.
+
+(* Normalform instance for Bool *)
+
+Section SecNFBool.
+
+Variable Shape : Type.
+Variable Pos : Shape -> Type.
```
MarvinLira: There should be 2 spaces of indentation inside of sections. You should check this also in the other files.
```suggestion
  Variable Shape : Type.
  Variable Pos : Shape -> Type.
```

base/coq/Prelude/List.v
```diff
@@ -27,6 +28,54 @@ End SecList.
...
+     | cons fx fxs => nf fx >>= fun nx =>
+                      fxs >>= fun xs =>
+                      nf'List xs >>= fun nxs =>
+                      pure (cons (pure nx) (pure nxs))
+     end.
+
+Definition nfList `{Normalform Shape Pos A B}
+                  (p : Free Shape Pos (List Shape Pos A))
+  : Free Shape Pos (List Identity.Shape Identity.Pos B)
+ := p >>= (fun p' => nf'List p').
```
MarvinLira: This can be written easier. This also applies similarly to `Pair`.
```suggestion
 := p >>= nf'List.
```

base/coq/Free/Class/Normalform.v
```diff
@@ -0,0 +1,21 @@
...
+        for recursive data types. *)
+    nf  : Free Shape Pos A -> Free Shape Pos B;
+    nf' : A -> Free Shape Pos B;
+    (** Property for moving nf into position functions *)
+    nf_impure: forall s (pf : _ -> Free Shape Pos A),
+        nf (impure s pf) = impure s (fun p => nf (pf p));
+    (** Property for unfolding nf on pure values *)
+    nf_pure : forall (x : A),
+        nf (pure x) = nf' x
+  }.
```
MarvinLira: The definition of `nf` seems to be always the same.
Couldn't you define this like the following?
That way, you would also not have to do the proofs in every instance definition.
```suggestion
Class Normalform {Shape : Type} {Pos : Shape -> Type}
                 (A B : Type) :=
  {
    (** The function is split into two parts due to termination check errors
        for recursive data types. *)
    nf' : A -> Free Shape Pos B
  }.

Definition nf {Shape : Type} {Pos : Shape -> Type}
    (A B : Type)
    {n : Normalform A B}
    (fx : Free Shape Pos A)
   : Free Shape Pos B :=
  fx >>= nf'.

(** Property for moving nf into position functions *)
Lemma nf_impure : forall {Shape : Type} {Pos : Shape -> Type}
    (A B : Type)
    (n : Normalform A B)
    s (pf : _ -> Free Shape Pos A),
    nf A B (impure s pf) = impure s (fun p => nf A B (pf p)).
Proof.
  trivial.
Qed.

(** Property for unfolding nf on pure values *)
Lemma nf_pure : forall {Shape : Type} {Pos : Shape -> Type}
    (A B : Type)
    (n : Normalform A B)
    (x : A),
  @nf Shape Pos A B n (pure x) = nf' x.
Proof.
  trivial.
Qed.

```

base/coq/Free/Verification/SharingHandlerTests.v
```diff
@@ -39,54 +38,114 @@ Definition evalTraceM {A : Type} p
...
+                             (fx : FreeA) (fy : FreeA)
+ : FreeA
+:= share (share fx >>= fun sx => f sx sx) >>= fun sy =>
+   share fy >>=  fun sz => f sy sz.
+
+(*
+let sx = val
+    sy = f sx fx
+    sz = f sy fy
+in f sx (f (sy (f sz val)))
```
MarvinLira:
```suggestion
in f sx (f sy (f sz val)) 
```

MarvinLira (changes requested):

base/coq/Free/Class/Normalform.v
MajaRet: That is a very good point and it will make things a lot simpler. Thank you!

MarvinLira (approved):

nbun (approved):

PR #171, opened 2020-08-24
Add a compiler flag for evaluation strategy selection

### LINKED ISSUE #168: Add compiler flag for evaluation strategy
## Description
Add a compiler flag `--strategy` to the compiler for choosing between call-by-need/call-by-name and call-by-value.

Daniel-Teut:
### Issue
Closes #168.

### Description of the Change
Adds a compiler option `--strategy STRAT` to allow the user to select a evaluation strategy. This strategy will change the behaviour of the `share` operator that is to be implemented. Implements a call-by-value and a call-by-need/name strategy.

### DIFF
--- a/src/exe/FreeC/Application/Options.hs
+++ b/src/exe/FreeC/Application/Options.hs
@@ -41,6 +41,8 @@ data Options = Options
     -- ^ The frontend to use.
   , optBackend                   :: String
     -- ^ The backend to use.
+  , optStrategy                  :: String
+    -- ^ The evalutation strategy to use.
   }
 
 -- | The default command line options.
@@ -68,4 +70,5 @@ makeDefaultOptions = do
     , optDumpTransformedModulesDir = Nothing
     , optFrontend                  = defaultFrontend
     , optBackend                   = defaultBackend
+    , optStrategy                  = "cbn"
     }
--- a/src/exe/FreeC/Application/Options.hs
+++ b/src/exe/FreeC/Application/Options.hs
@@ -2,9 +2,10 @@
 --   that stores the values of the command line options.
 module FreeC.Application.Options ( Options(..), makeDefaultOptions ) where
 
-import           Paths_free_compiler ( getDataFileName )
+import           Paths_free_compiler        ( getDataFileName )
 
 import {-# SOURCE #-} FreeC.Backend
+import           FreeC.Environment.Strategy
 import {-# SOURCE #-} FreeC.Frontend
 
 -- | Data type that stores the command line options passed to the compiler.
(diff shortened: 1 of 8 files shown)


src/exe/FreeC/Application/Options.hs
```diff
@@ -68,4 +70,5 @@ makeDefaultOptions = do
     , optDumpTransformedModulesDir = Nothing
     , optFrontend                  = defaultFrontend
     , optBackend                   = defaultBackend
+    , optStrategy                  = "cbn"
```
just95: Like with the front end and back end options, this value should not be hard-coded.

src/exe/FreeC/Application/Options/Descriptors.hs
```diff
@@ -80,4 +80,12 @@ optionDescriptors
                  , "Allowed values are: " ++ showBackends ++ "."
                  , "Defaults to: `" ++ defaultBackend ++ "`."
                  ])
+    , Option [] ["strategy"]
+        (ReqArg (\p opts -> opts { optStrategy = p }) "STRAT")
+        (unlines
+         [ "Optional. Specifies which evaluation strategy"
+         , "the resulting program will use."
+         , "Allowed values are: `cbn`(call-by-need/name) and `cbv`(call-by-value)."
+         , "Defaults to `cbn`."
```
just95: The allowed and default values in the help message should also be generated.

src/lib/FreeC/Environment/Strategy.hs
```diff
@@ -0,0 +1,15 @@
...
+--
+--   This strategy is decided by the user as an command line option and passed
+--   into the environment. It determines how the @share@ operator behaves.
+module FreeC.Environment.Strategy where
+
+import           Data.Map.Strict
+
+-- | The data type for an evaluation strategy.
+data Strategy = CallByValue | CallByNameOrNeed
+ deriving Show
```
just95: I'm currently not sure whether we should not have distinct values for *call-by-need* and *call-by-name* such that we have more flexibility when the strategies are actually implemented.

```suggestion
data Strategy = CallByValue | CallByName | CallByNeed
 deriving Show
```

Mixing *call-by-need* and *call-by-name* does especially make no sense to me right now since the compiler currently implements *call-by-name* and our goal is to add support for alternative evaluation strategies like *call-by-need* (by adding sharing) and *call-by-value*. Thus, it would be surprising to me if *call-by-need* can be treated the same as *call-by-name* whereas *call-by-value* has to be treated differently.

src/lib/FreeC/Environment.hs
```diff
@@ -38,6 +38,7 @@ module FreeC.Environment
   , lookupDecArg
   , lookupDecArgIndex
   , lookupDecArgIdent
+  , changeStrategy
```
just95: This does not really fit into the "Looking up Entries from the Environment" section and should either be moved to "Modifying Entries in the Environment" or a new section should be created for settings.

just95 (changes requested):
Documentation for the new command line option still has to be added to the README and CHANGELOG.

PR #173, opened 2020-08-26
Implement ShareableArgs typeclass

### LINKED ISSUE #149: Implement ShareableArgs typeclass
## Description

Implement deep sharing via a typeclass `ShareableArgs` to share arguments when using `cbneed`.

## Motivation

TBA

## Dependencies

This issue does not depend on any open issues.

## ToDo

- [ ] Implement a `ShareableArgs` typeclass.
- [ ] Implement `ShareableArgs` instances for pre-defined data types.
- [ ] Add sharing of arguments and a `ShareableArgs` constraint to `cbneed`.
- [ ] Add tests for deep sharing to the sharing handler tests.

MajaRet:
### Issue

Closes #149 
 
### Description of the Change

A typeclass `ShareableArgs` was implemented that allows us to share complex data structures not only wholly, but also share their individual components.

For example, previously it would not have been possible to share the choice in the expression (given in Haskell syntax for readability purposes): 
```haskell
share [0?1] >>= \sharedList -> headM sharedList + headM sharedList
```

Meaning the result of the computation would be `[0,1,1,2]` instead of the desired `[0,2]`. 
The `ShareableArgs` typeclass provides a function `shareArgs` that is used by `cbneed` to share components as well as the whole data structure, solving this problem.

`ShareableArgs` instances have been implemented for all data types in the Prelude.

### Possible Drawbacks

Because `cbneed` is now only defined for types `A` that have a `ShareableArgs` instance, the type of `share` provided by the `Shareable` type class had to be changed so that `share` would only apply to types with a `ShareableArgs` instance.
But as a consequence, because `ShareableArgs` instances use `cbneed` (which requires the `Share` effect), all instances of `Shareable` must now contain the `Share` effect - even the `Cbn` instance, which does not actually need it.

(Alternative solutions are welcome!)

### Verification Process

Examples testing deep sharing have been added to the sharing handler tests.

### Additional Notes

The sharing handlers were not correct before. They did not distinguish between the sharing scope and the state and could, for example, not handle deep sharing correctly. This problem has now been fixed.

Functions that contain/depend on sharing now need some additional constraints:
- An `'{Injectable Share.Shape Share.Pos Shape Pos}` constraint
- For every type variable `a` involved in sharing (or simply every type variable), a constraint `'{ShareableArgs Shape Pos a}`.

Also note that an explicit application of share now includes a `ShareableArgs` instance (and an `Injectable Share.Shape Share.Pos Shape Pos` instance, but that can be passed through), but it is not currently possible to determine this instance explicitly. However, it can always be inferred, so there should be no harm in using an underscore in its place.


base/coq/Free/Verification/SharingHandlerTests.v
```diff
@@ -32,7 +32,7 @@ Definition evalTracing {A : Type} p
 Definition evalNDM {A : Type} p
 := @collectVals (option A) (run (runChoice (runNDSharing (0,0) (runMaybe p)))).
 
-(* Shortcut to evaluate a traced partial program to a result and a list 
+(* Shortcut to evaluate a traced partial pro gram to a result and a list
```
nbun:
```suggestion
(* Shortcut to evaluate a traced partial program to a result and a list 
```

base/coq/Free/Verification/SharingHandlerTests.v
```diff
@@ -570,4 +628,57 @@ Example exOrRecFalseTracing
...
+   well. Since we take the first element twice, the second tracing message ("1") 
+   should not be logged and the first should be shared and thus logged once. 
+*)
+Example exAddDeepPairTrace
+ : evalTracing (nf (doubleDeepSharedPair Cbneed_ addInteger_ tracePair))
+  = (0%Z, ["0"%string]).
+Proof. constructor. Qed.
+
+(* 
+let sx = (trace "0" 0, trace "1" 1)
```
nbun:
```suggestion
let sx = [trace "0" 0, trace "1" 1]
```

base/coq/Free/Verification/SharingHandlerTests.v
```diff
@@ -570,4 +628,57 @@ Example exOrRecFalseTracing
...
+in head sx + head sx 
+=> The list is shared, so the effects inside the pair should be shared as
```
nbun:
```suggestion
=> The list is shared, so the effects inside the list should be shared as 
```

nbun (approved):

MarvinLira (approved):
> Because `cbneed` is now only defined for types `A` that have a `ShareableArgs` instance, the type of `share` provided by the `Shareable` type class had to be changed so that `share` would only apply to types with a `ShareableArgs` instance.
> But as a consequence, because `ShareableArgs` instances use `cbneed` (which requires the `Share` effect), all instances of `Shareable` must now contain the `Share` effect - even the `Cbn` instance, which does not actually need it.
> 
> (Alternative solutions are welcome!)

An alternative would be to remove those constraints `ShareableArgs` and `Injectable Share.Shape Share.Pos Shape Pos` from the class definition of `Shareable` and only introduce them in the instances of `Shareable` that need them like `Cbneed`. But to add the constraint `ShareableArgs` to the instance declarations the type argument `A` had to be moved from the function `share` to the class `Shareable`.

```Coq
Class Shareable (Shape : Type) (Pos : Shape -> Type) {A : Type} :=
  {
    share : 
      Free Shape Pos A -> Free Shape Pos (Free Shape Pos A)
  }.

Instance Cbneed (Shape : Type) (Pos : Shape -> Type) {A : Type}
                `{I : Injectable Share.Shape Share.Pos Shape Pos}
                `{S : ShareableArgs Shape Pos A}
 : @Shareable Shape Pos A | 1 := {
    share p := @cbneed Shape Pos A I (@shareArgs Shape Pos A S) p
}.

Instance Cbn (Shape : Type) (Pos : Shape -> Type) {A : Type}
 : @Shareable Shape Pos A | 2 := {
    share p := @cbn A Shape Pos p
}.
```

Another alternative would be to parametrize the class by some requirement `R` for supported types.

```Coq
Class Shareable (Shape : Type) (Pos : Shape -> Type) (R : Type -> Type):=
  {
    share {A : Type} `{S : R A} : 
      Free Shape Pos A -> Free Shape Pos (Free Shape Pos A)
  }.

Instance Cbneed (Shape : Type) (Pos : Shape -> Type)
                `{I : Injectable Share.Shape Share.Pos Shape Pos}
 : Shareable Shape Pos (ShareableArgs Shape Pos) | 1 := {
    share A S p := @cbneed Shape Pos A I (@shareArgs Shape Pos A S) p
}.

Instance Cbn (Shape : Type) (Pos : Shape -> Type) {A : Type}
 : Shareable Shape Pos (fun _ => unit) | 2 := {
    share A tt p := @cbn A Shape Pos p
}.
```

MajaRet:
Thanks for the input! Unfortunately, I don't think either of those ideas will work.

Adding the type as a parameter to `Shareable` is something we should not do. That's because there may be many type variables involved in sharing (for example if we share a pair, or if we have two consecutive `share` calls for values of different types). Therefore, we would need a `Shareable` instance parameter for each of them.
But since the `Shareable` instance determines the evaluation strategy, that could lead to different strategies being chosen for different types, which introduces inconsistencies.

As for the second idea, Coq cannot infer a value of type `R A` because it does not know what `R` is (because the `Shareable` instance is not fixed, but given as a parameter to each translated function). And we cannot give a value of type `R A` explicitly because we also do not know what `R` is at that point.

MajaRet:
Thanks for the input! I don't think the first idea is something we should do, but the second should work (at the price of possibly adding some more constraints to a function's type signature).

Adding the type as a parameter to `Shareable` is something we should not do because there may be many type variables involved in sharing (for example if we share a pair, or if we have two consecutive `share` calls for values of different types). Therefore, we would need a `Shareable` instance parameter for each of them.
But since the `Shareable` instance determines the evaluation strategy, that could lead to different strategies being chosen for different types (since `Shareable` instances are given manually by the user), which introduces inconsistencies.

I tried out the second approach, and from what I can see, it will work if we make the following changes to the parameters of a function: 
- for every call of `share` on an expression of type `Free Shape Pos A`, we add a constraint `R A` to our type signature (even if `A` is not a type variable, but for example a Bool or a list of `B`s) 
- If we call another function, we add all of its `R` constraints to our type signature as well (filtering out duplicates). 

That could lead to rather large type signatures in exchange for getting rid of the `Share` constraint. I'd say we wait to see how harmful the `Share` effect actually is in proofs before we decide whether to use this design. But it is definitely a promising idea!

PR #179, opened 2020-08-28
Add the `share` operator to the lifted IR

### LINKED ISSUE #176: Add a `share` operator to the lifted IR
## Description
Add a `Share` construct to the lifted IR and its translation to Coq. Use it to translate the `let` expressions in the IR to the lifted IR.

## Motivation
This allows us to generate the `share` operator with the compiler, so it can be used in the resulting Coq code.

Daniel-Teut:
### Issue
Closes #176.

### Description of the Change
Adds a `Share` construct to the lifted IR and a translation to Coq. Further use this new construct to translate IR `Let`.

### Possible Drawbacks
Since Agda does not implement a `share` operator, translations with `let`s to Agda could fail.
Also, currently the `Shareable` constraint is not generated in functions, which can lead to invalid Coq code.


src/lib/FreeC/Backend/Agda/Converter/Expr.hs
```diff
@@ -35,6 +36,9 @@ convertLiftedExpr (LIR.Bind _ arg k)
   = bind <$> convertLiftedExpr arg <*> convertLiftedExpr k
 convertLiftedExpr (LIR.Undefined _)           = return undefinedExpr
 convertLiftedExpr (LIR.ErrorExpr _)           = return errorExpr
+convertLiftedExpr (LIR.Share srcSpan)         = reportFatal
+  $ Message srcSpan Error
+  "The share operator is currently not supported in Agda."
```
just95: Once #169 is merged, `let` expressions (and therefore `share` operators) are added to almost all programs. Thus, this change would break the Agda back end. A safe fallback is to convert the `share` operator to `pure` in Agda. This corresponds to *call-by-name* semantics.

src/lib/FreeC/Backend/Coq/Base.hs
```diff
@@ -69,6 +73,25 @@ freeArgs = [ (Coq.bare "Shape", Coq.Sort Coq.Type)
                )
            ]
 
+-------------------------------------------------------------------------------
+-- Sharing                                                                   --
+-------------------------------------------------------------------------------
```
just95: In the export specifier list the *Partiality* section precedes the *Sharing* section. Thus, the code should be moved below the code for partiality as well.

src/lib/FreeC/Backend/Coq/Base.hs
```diff
@@ -69,6 +73,25 @@ freeArgs = [ (Coq.bare "Shape", Coq.Sort Coq.Type)
...
+-- | The Coq Identifier for the @Shareable@ type class.
+shareable :: Coq.Qualid
+shareable = Coq.bare "Shareable"
+
+-- | The name and type of the @Shareable@ instance that must be passed to
+--   shareable functions.
+shareableArg :: (Coq.Qualid, Coq.Term)
+shareableArg = ( Coq.bare "S"
+               , Coq.app (Coq.Qualid shareable)
+                   [Coq.Qualid (Coq.bare "Shape"), Coq.Qualid (Coq.bare "Pos")]
```
just95: Maybe we should add constants for `Coq.bare "Shape"` and `Coq.bare "Pos"`.

src/lib/FreeC/Backend/Coq/Base.hs
```diff
@@ -69,6 +73,25 @@ freeArgs = [ (Coq.bare "Shape", Coq.Sort Coq.Type)
...
+               )
+
+-- | The Coq Identifier for the @share@ operator.
+share :: Coq.Qualid
+share = Coq.bare "share"
```
just95: These names need to be added to the list of `reservedIdents` to avoid name conflicts with user defined identifiers.

src/lib/FreeC/LiftedIR/Effect.hs
```diff
@@ -5,5 +5,5 @@ module FreeC.LiftedIR.Effect where
 --
 --   These effects corresponds to type classes constraining the container used
 --   by the @Free@ monad. Currently just partiality is supported.
```
just95:
```suggestion
--   by the @Free@ monad.
```

src/lib/FreeC/LiftedIR/Converter/Expr.hs
```diff
@@ -326,3 +325,25 @@ rawBind srcSpan mx x varType expr = do
...
+--   The given expression is the right-hand side of the let.
+liftBinds :: [IR.Bind] -> IR.Expr -> Converter LIR.Expr
+liftBinds [] expr = liftExpr expr
+liftBinds
+  ((IR.Bind srcSpan (IR.VarPat patSrcSpan ident varPatType isStrict) bindExpr)
+   : bs) expr = do
+    _ <- renameAndDefineLIRVar srcSpan isStrict ident varPatType
+    expr' <- liftBinds bs expr
+    patType' <- mapM LIR.liftType varPatType
+    bindVarPat' <- varPat patSrcSpan (IR.UnQual $ IR.Ident ident) patType'
```
just95: We should rename the function `varPat` (for example to `makeVarPat`).

src/lib/FreeC/LiftedIR/Converter/Expr.hs
```diff
@@ -326,3 +325,25 @@ rawBind srcSpan mx x varType expr = do
...
+    shareType' <- mapM LIR.liftType' varPatType
+    bindExpr' <- liftExpr bindExpr
+    return
+      $ LIR.Bind srcSpan (LIR.App srcSpan (LIR.Share srcSpan)
+                          (maybeToList shareType') [Sharing] [bindExpr'] True)
+      (LIR.Lambda srcSpan [bindVarPat'] expr')
```
just95: This expression is pretty long. The following may be easier to comprehend.

```suggestion
  let shareExpr = LIR.App srcSpan (LIR.Share srcSpan) (maybeToList shareType')
        [Sharing] [bindExpr'] True
  return $ LIR.Bind srcSpan shareExpr (LIR.Lambda srcSpan [bindVarPat'] expr')
```

src/lib/FreeC/LiftedIR/Converter/Expr.hs
```diff
@@ -326,3 +325,25 @@ rawBind srcSpan mx x varType expr = do
...
+-- Let Expression Helper                                                     --
+-------------------------------------------------------------------------------
+-- | Lifts a list of let bindings.
+--
+--   The given expression is the right-hand side of the let.
+liftBinds :: [IR.Bind] -> IR.Expr -> Converter LIR.Expr
+liftBinds [] expr = liftExpr expr
+liftBinds
+  ((IR.Bind srcSpan (IR.VarPat patSrcSpan ident varPatType isStrict) bindExpr)
+   : bs) expr = do
```
just95: To avoid the variable to be visible outside of the `let` expression, a local environment must be used here.

```suggestion
   : bs) expr = localEnv $ do
```

just95 (changes requested):
The type class instance `S` for the `Sharing` effect is currently only passed to the `share` operator. There also has to be a type class instance argument for every function desclaration that uses the sharing effect (or another function that does) such that `S` is not undefined and `S` has to be passed to applications of such functions explicitly by adding the `Sharing` effect to their effect set. The latter requires the environment to be modified to store arbitrary effect sets (currently there is only a flag for partiality) and the existing `PartialityAnalysis` pass has to be extended to an `EffectAnalysis` pass. This may be done in a separate ticket but must be done before both this PR and #169 are merged.

src/lib/FreeC/Pass/EffectAnalysisPass.hs
```diff
@@ -0,0 +1,122 @@
...
+isPartialFuncName name | name == IR.Prelude.undefinedFuncName = return True
+                       | name == IR.Prelude.errorFuncName = return True
+                       | otherwise = inEnv $ hasEffect Partiality name
+
+-------------------------------------------------------------------------------
+-- Sharing                                                                   --
+-------------------------------------------------------------------------------
+-- | Tests whether the given functions contain any let expressions.
+areSharedFuncDecls :: [IR.FuncDecl] -> Converter Bool
+areSharedFuncDecls = return . any (isLet . IR.funcDeclRhs)
```
just95: The catch all case in the your definition of `isLet` can cause issues when the IR is extended. The following definition does not have to traverse the AST explicitly.

```suggestion
-- | Tests whether the given functions contain any let expressions.
areSharedFuncDecls :: [IR.FuncDecl] -> Converter Bool
areSharedFuncDecls = return . any (isJust . findFirstSubterm isLet . IR.funcDeclRhs)
 where
  isLet (IR.Let _ _ _ _) = True
  isLet _                = False
```

base/Prelude.toml
```diff
@@ -81,7 +81,7 @@ exported-values = [
   coq-name        = 'andBool'
   agda-name       = '_∧_'
   arity           = 2
-  partial         = false
+  effects         = []
```
just95, line 84: The `version` of the module interface configuration file format has to be increased in the encoder and decoder and the change to the format has to be documented in [`./doc/ModuleInterfaceFileFormat.md`](https://github.com/FreeProving/free-compiler/blob/main/doc/ModuleInterfaceFileFormat.md) as well as the comments in `FreeC.Environment.ModuleInterface.Decoder`.

src/lib/FreeC/Backend/Coq/Base.hs
```diff
@@ -92,6 +116,85 @@ partialUndefined = Coq.bare "undefined"
...
+-- | The Coq binder for the @ShareableArgs@ type class with the type variable
+--   with the given name.
+shareableArgsBinder :: Coq.Qualid -> Coq.Binder
+shareableArgsBinder typeArg = Coq.Generalized Coq.Implicit
+  $ Coq.app (Coq.Qualid shareableArgs)
+  $ map Coq.Qualid [shape, pos, typeArg]
+
+-- | The Coq identifier for an implicit argument.
+implicitArg :: Coq.Qualid
+implicitArg = Coq.bare "_"
```
just95: `language-coq` has a special constructor [`UnderscoreName`](https://github.com/FreeProving/language-coq/blob/c6ba1cfc806784f74a3247d2c7f8e9c6f9c9d00f/src/lib/Language/Coq/Gallina.hs#L230) for this purpose.

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Common.hs
```diff
@@ -5,6 +5,7 @@ module FreeC.Backend.Coq.Converter.FuncDecl.Common
     convertFuncHead
   ) where
 
+import           FreeC.Backend.Coq.Base
```
just95: This module [should be imported qualified](https://github.com/FreeProving/guidelines/blob/main/CONTRIBUTING.md#use-qualified-imports) as [`Coq.Base`](https://github.com/FreeProving/free-compiler/blob/617d43305da7aaf5ca72668ecbd47b982193c999/.hlint.yaml#L31).
Don't remove the import qualifier in `FreeC.Backend.Coq.Converter.FuncDecl.Rec.WithSections`.

src/lib/FreeC/Backend/Coq/Base.hs
```diff
@@ -92,6 +116,85 @@ partialUndefined = Coq.bare "undefined"
...
+
+-- | The Coq identifier for the argument of the @Injectable@ type class.
+injectableArg :: Coq.Qualid
+injectableArg = Coq.bare "I"
+
+-- | The Coq binder for the @Injectable@ type class.
+injectableBinder :: Coq.Binder
+injectableBinder = Coq.typedBinder' Coq.Generalizable Coq.Implicit injectableArg
+  $ Coq.app (Coq.Qualid injectable)
+  $ map Coq.Qualid [Coq.bare "Share.Shape", Coq.bare "Share.Pos", shape, pos]
```
just95: `Share.Share` and `Share.Pos` are not `bare` but [`Qualified`](https://github.com/FreeProving/language-coq/blob/c6ba1cfc806784f74a3247d2c7f8e9c6f9c9d00f/src/lib/Language/Coq/Gallina.hs#L236) identifiers.

src/lib/FreeC/Backend/Coq/Converter/FuncDecl/Rec/WithHelpers.hs
```diff
@@ -165,7 +165,7 @@ transformRecFuncDecl
     -- If the original function was partial, the helper function is partial as
     -- well.
```
just95:
```suggestion
    -- The helper function uses all effects that are used by the original 
    -- function.
```

src/lib/FreeC/Backend/Coq/Converter/Arg.hs
```diff
@@ -27,12 +27,15 @@ import           FreeC.Monad.Converter
 convertTypeVarDecls
   :: Coq.Explicitness -- ^ Whether to generate an explicit or implicit binder.
   -> [IR.TypeVarDecl] -- ^ The type variable declarations.
-  -> Converter [Coq.Binder]
+  -> Converter ([Coq.Binder], [Coq.Qualid])
```
just95: The meaning of the additional return value should be documented.
Alternatively, you could also just lookup the Coq identifiers in `convertFuncHead`.

just95 (changes requested):

maclement (approved):

src/lib/FreeC/Backend/Coq/Base.hs
```diff
@@ -92,6 +129,89 @@ partialUndefined = Coq.bare "undefined"
...
+-- | The Coq binder for the @ShareableArgs@ type class with the type variable
+--   with the given name.
+shareableArgsBinder :: Coq.Qualid -> Coq.Binder
+shareableArgsBinder typeArg = Coq.Generalized Coq.Implicit
+  $ Coq.app (Coq.Qualid shareableArgs)
+  $ map Coq.Qualid [shape, pos, typeArg]
+
+-- | The Coq identifier for an implicit argument.
+implicitArg :: Coq.Qualid
+implicitArg = Coq.bare $ showP Coq.Underscore
```
just95: The idea of my [original suggestion](https://github.com/FreeProving/free-compiler/pull/179#discussion_r484566419) was to use no bare identifier for `_` but the corresponding AST node. Since `implicitArg` is always wrapped with a `Coq.Qualid` constructor, you can just return a `Coq.Term` directly.

```suggestion
-- | The Coq identifier for an implicit argument.
implicitArg :: Coq.Term
implicitArg = Coq.Underscore
```

For example

```haskell
implicitEffectArgs' = map Coq.Qualid $ concatMap Coq.Base.selectImplicitArgs effects
```

would become

```haskell
implicitEffectArgs' = concatMap Coq.Base.selectImplicitArgs effects
```

src/lib/FreeC/Backend/Coq/Base.hs
```diff
@@ -47,6 +67,24 @@ generatedLibName = Coq.ident "Generated"
...
+
+-- | The Coq identifier for the @Pos@ argument of the @Free@ monad.
+pos :: Coq.Qualid
+pos = Coq.bare "Pos"
+
+-- | Like 'pos' but not wrapped in a 'Coq.Bare' constructor.
+posIdent :: Coq.Ident
+posIdent = let (Coq.Bare ident) = pos
+           in ident
+
```
just95: Defining `Qualid`s in terms of the `Ident`s would make more sense.

```suggestion
-- | The Coq identifier for the @Shape@ argument of the @Free@ monad.
shape :: Coq.Qualid
shape = Coq.Bare shapeIdent

-- | Like 'shape' but not wrapped in a 'Coq.Bare' constructor.
shapeIdent :: Coq.Ident
shapeIdent = Coq.ident "Shape"

-- | The Coq identifier for the @Pos@ argument of the @Free@ monad.
pos :: Coq.Qualid
pos = Coq.Bare posIdent

-- | Like 'pos' but not wrapped in a 'Coq.Bare' constructor.
posIdent :: Coq.Ident
posIdent = Coq.ident "Pos"

```

just95 (changes requested):

just95 (approved):

Issue #178, opened 2020-08-28
Check whether notation-scopes should be used


MarvinLira:
## Description

Gather information about [notation-scopes](https://coq.inria.fr/refman/user-extensions/syntax-extensions.html#notation-scopes) and check whether those should be used in #153 or whether they are not needed there.

## Motivation

We need to qualify notations like #153 to avoid name clashes. This could be done by naming those notations `ModuleName.NotationName` but the Coq-way to do this would be to use notation scopes like
```Coq
Open Scope ModuleName_notation_scope.
... NotationName ...
Close Scope ModuleName_notation_scope.
```
or `NotationName%ModuleName`.

## Alternatives

The current approach of #153 tries to stick close to the usage of an function. Therefore for every notation `NotationName` defined there an explicit application `@NotationName` is defined as well as qualified versions `ModuleName.NotationName` and `@ModuleName.NotationName`. Those qualified versions have to be in their own module to be imported seperatly.

MarvinLira:
When using notation scopes we would need to use scope keys with the `%` notation as opening multiple scopes could lead to name clashes. This would however make the notations look not so pretty as the scope key has to follow after the whole notation. For a list constructor it would be `(@MyCons Shape Pos a x xs)%MyData` instead of `@MyData.MyCons Shape Pos a x xs`.

Another more important note is, that scope names can not contain a `.`. So we could not use the module name as a scope name but had to e.g. replace every `.` with a `_`. This could however introduce name clashes as a haskell module `A_B` and a haskell module `A.B` would share the same notation scope.

The issue, that notations are not imported with a `From ... Require ... .` sentence and have to be in their own module, that can be imported with a `Import ... .`, applies to scoped notations as well.

For those reasons the usage of notation scopes to qualify notations are not an improvement to the current approach of #153, which defines for each notation `NotationName` in module `ModuleName` an additional qualified version `ModuleName.NotationName` of that notation. Instead it introduces the problem of name clashes due to naming restrictions for scopes.

PR #180, opened 2020-08-29
Define smart constructors for data types as notations

### LINKED ISSUE #153: Define smart constructors for data types as notations
## Description

Currently for each constructor of each data type a smart constructor is defined that lifts the return value of that constructor into a monadic value.
```Coq
Definition MyPair_ (Shape : Type) (Pos : Shape -> Type) {a b : Type} (x : Free Shape Pos a) (y : Free Shape Pos b)
   : Free Shape Pos (MyPair Shape Pos a b) :=
  pure (myPair_ x y).
```
It would be nice if those smart constructors would be implemented as `Notation` instead of `Definition`.

## Motivation

The approach with the `Definition` has two disadvantages:
- The smart constructor is not automatically used by Coq for output.
  Destructing a `Free Shape Pos (MyPair Shape Pos a b)` twice yields a `pure (myPair_ fx fy)` instead of a `MyPair_ fx fy`.
- Some simple tactics of coq fail because they do not unfold smart constructors.
  ```Coq
  Lemma lem1 : forall (p : MyPair Shape Pos a b),
    f (pure p) = C.
  Proof. Admitted.

  Lemma lem2 : forall (fx : Free Shape Pos a) (fy : Free Shape Pos b),
    f (MyPair_ Shape Pos fx fy) = C.
  Proof.
    intros fx fy.
    Fail rewrite lem1.
  Admitted.

  Lemma lem3 : forall (fx : Free Shape Pos a) (fy : Free Shape Pos b),
    f (pure (myPair_ fx fy)) = C.
  Proof.
    intros fx fy.
    Fail rewrite lem2.
  Admitted.
  ```
Both points do not apply if a `Notation` is used.

MarvinLira:
### Issue

Closes #153 

### Description of the Change

For each constructor of each data type declaration there is a smart constructor which calls the original constructor and packs the resulting value with a `pure`. This smart constructor was defined as a function before and is now defined as four notations to imitate the behavior of the function. Those notations represent the (fully applied) call or (fully applied) explicit call of the unqualified or (with the full module name) qualified smart constructor.
Consider the following example.
```haskell
module M.N where
data Foo a = FooCon a
```
The following notations are generated.
```Coq
(* Smart constructors for Foo *)
Notation "'FooCon' Shape Pos x_0" := (@pure Shape Pos (Foo Shape Pos _) (@fooCon
     Shape Pos _ x_0))
  ( at level 10, Shape, Pos, x_0 at level 9 ).
Notation "'@FooCon' Shape Pos a x_0" := (@pure Shape Pos (Foo Shape Pos a)
    (@fooCon Shape Pos a x_0))
  ( only parsing, at level 10, Shape, Pos, a, x_0 at level 9 ).

(* Qualified smart constructors *)
Module QualifiedSmartConstructorModule.
(* Qualified smart constructors for Foo *)
Notation "'M.N.FooCon' Shape Pos x_0" := (@pure Shape Pos (Foo Shape Pos _)
    (@fooCon Shape Pos _ x_0))
  ( at level 10, Shape, Pos, x_0 at level 9 ).
Notation "'@M.N.FooCon' Shape Pos a x_0" := (@pure Shape Pos (Foo Shape Pos a)
    (@fooCon Shape Pos a x_0))
  ( only parsing, at level 10, Shape, Pos, a, x_0 at level 9 ).
End QualifiedSmartConstructorModule.
```
Note that the local module `QualifiedSmartConstructorModule` contains all qualified smart constructors of that module. Importing the module `M.N` lets you use the unqualified smart constructors and importing `M.N.QualifiedSmartConstructorModule` lets you use the qualified smart constructors.
The notations are designed in such a way that the translation of smart constructor calls had not to to be changed.

### Alternate Designs

The smart constructors could have been qualified by using notation scopes instead of writing the `.` notation (see #178). This approach has been discarded due to restrictions to the names of notation scopes. The name of a notation scope must not contain a `.` and therefore the module name could not be used to give the qualified notations in each module a unique notation scope to prevent name clashes.

When importing a module only the unqualified notations are visible but not the qualified notations. There are two ways to change this:
1. The `QualifiedSmartConstructorModule` module could be exported. But as Coq uses the notation that was defined last for printing and the unqualified notations should be preferred in this case, the compiler would have to define the unqualified smart constructors after that export. This way they would no longer be directly under the definition of the corresponding data type as they used to be.
2. Copies of the qualified notations could be left outside of the local module, directly above the unqualified notation definitions.

In every module that imports `QualifiedSmartConstructorModule` modules, those modules could be exported in its own `QualifiedSmartConstructorModule` module. This way the qualified smart constructor notations of data structures would be available when importing a module with function definitions that uses function definitions of a second module that again uses datatypes of a third module. This was not done yet because we would have many (qualified) names in the environment of compiled Coq modules that are not in the environment of the original haskell modules. However, as those names are qualified it should be safe to do this.
```haskell
module A where
data MyBool = MB Bool

module B where
import A
myAnd :: MyBool -> MyBool -> MyBool
myAnd b1 b2 = case b1 of
  (MB b) -> if b then b2 else b1

module C where
import B
myId b = myAnd b b
```
```Coq
From Base Require Import Free.
From Generated Require Import C.

Lemma myId_id: forall S P mb,
  myId S P (pure mb) = pure mb.
Proof.
  intros; destruct mb as [b].
(* subgoal with current compilation:
   [myId S P (pure (A.mb b)) = pure (A.mb b)] *)
(* subgoal with descripted change:
   [myId S P (A.MB S P b) = A.MB S P b] *)
```

### Possible Drawbacks

Due to the importation behavior of notations in Coq the qualified notations are only visible if the submodule `QualifiedSmartConstructorModule` is imported.

The smart constructor notations take a `Shape` and a `Pos` argument as the smart constructor did it before and the compiler relies on this in the translation process. Hiding those arguments afterwards with a `Arguments` sentence is not possible for a notation. It would however be possible to define additional notations without those arguments.

Though those notations imitate the behavior of a function they do not offer all features of a function. E.g. they can not be partially applied. This special case is however no problem for our compiler as the compiler always fully applies functions (and constructors).

### Verification Process

The local module `QualifiedSmartConstructorModule` is only imported from non `Base` library modules that have a constructor entry in their interface whose name is qualified by the name of the module which the interface belongs to.

The current test cases for the data type translation have been adapted to test for the correct (unqualified) smart constructor notations.
The examples have been tested with the new smart constructors.
Additional tests for the compilation of the module have been made by @MajaRet .


src/lib/FreeC/Backend/Coq/Converter/TypeDecl.hs
```diff
@@ -15,6 +28,8 @@ import           FreeC.Backend.Coq.Converter.Free
 import           FreeC.Backend.Coq.Converter.Type
 import qualified FreeC.Backend.Coq.Syntax         as Coq
 import           FreeC.Environment
+import qualified FreeC.Environment.Fresh          as Fresh
```
MajaRet: `FreeC.Environment.Fresh` and `FreeC.Environment.Renamer` do not have entries in `.hlint.yaml`, so they should not be [imported qualified](https://github.com/FreeProving/guidelines/blob/main/CONTRIBUTING.md#use-qualified-imports).

src/lib/FreeC/Backend/Coq/Converter/TypeDecl.hs
```diff
@@ -177,23 +196,131 @@ convertDataDecl (IR.DataDecl _ (IR.DeclIdent _ name) typeVarDecls conDecls) = do
...
+        , Coq.sModLevel 10
+        , Coq.sModIdentLevel (NonEmpty.fromList expArgIdents) (Just 9)
+        ]
 -- Type synonyms are not allowed in this function.
 convertDataDecl (IR.TypeSynDecl _ _ _ _)
   = error "convertDataDecl: Type synonym not allowed."
+
+-- | Partitions a list of sentences into a list of sentences that belong to
+--   qualified smart constructors and a list of the remaining sentences.
+partitionIsQualifiedSmartConstructor
```
MajaRet: This seems like a somewhat roundabout way of separating qualified notations from other sentences. 
If we ever decide to add more (unrelated) notations to the translation of a type declaration, and happen to put them after the qualified notations, they would also be counted as qualified notations.

Since `generateSmartConDecls` already returns the qualified notations separately, couldn't they just get passed through to `convertTypeComponent`, i.e. couldn't `convertTypeComponent` return a pair of `Sentence` lists, with the first list containing the data type declarations, unqualified notations etc. and the second list containing the qualified notations?

MajaRet (changes requested):

src/lib/FreeC/Backend/Coq/Converter/TypeDecl.hs
MarvinLira: I wasn't sure which of those implementations was the best way. I decided to use the partition function to have few changes to the signatures of existing functions and keep the interface to the type conversion simple. But I guess your right that the explicit separation of those sentences through the code is more fail-safe regarding future changes and should therefore be used.

src/lib/FreeC/Backend/Coq/Converter/Module.hs
```diff
@@ -65,12 +80,41 @@ convertImportDecl (IR.ImportDecl _ modName) = do
...
+  Just interface <- inEnv $ lookupAvailableModule modName
+  return (mkImportSentences [(Coq.ident (showPretty modName), interface)])
  where
-  -- | Makes a @From … Require Import …@ or  @From … Require …@.
-  mkRequireSentence :: Coq.ModuleIdent -> [Coq.ModuleIdent] -> Coq.Sentence
-  mkRequireSentence | libName == Coq.Base.baseLibName = Coq.requireImportFrom
-                    | otherwise = Coq.requireFrom
+  -- | Makes a @From … Require Import …@ or @From … Require …@ and an
+  --   @Export … .QualifiedSmartConstructorModule@ if this local module exists.
+  mkImportSentences :: [(Coq.ModuleIdent, ModuleInterface)] -> [Coq.Sentence]
```
Daniel-Teut: Is it actually necessary to have a list as the argument? As far as I can tell, the function gets only called once with a singleton list. This would also allow to remove the pair in favor of two arguments.

Daniel-Teut (changes requested):

MajaRet (approved):

Daniel-Teut (approved):

PR #184, opened 2020-09-01
Adapt sharing infrastructure to flexible evaluation strategy approach

### LINKED ISSUE #183: Adapt sharing infrastructure to flexible strategy approach
## Description

Add another operation `call :: forall (A : Type), Free Shape Pos A -> Free Shape Pos (Free Shape Pos A)` to the `Shareable` class and add a `Cbv` instance for `Shareable`.
Since the `Shareable` class actually determines the evaluation strategy, it should also be renamed to `Strategy`.

## Motivation

We do not want to choose the modeled evaluation strategy at compile time, but want to enable the user to prove properties about the evaluation of their code for any of the provided strategies (call-by-value, call-by-name, call-by-need). 

To model call-by-value correctly, we need to introduce let bindings for every non-variable expression in a program. These let bindings are translated to the `share` operator.
However, in a call-by-need setting, the `share` operator introduces sharing syntax every time it is used - even when nothing is shared -, which will likely compliclate proofs.
Therefore, we introduce a second operator `call`, which does not introduce sharing syntax. 
The compiler will [analyze](https://github.com/FreeProving/free-compiler/pull/169) where values are actually shared and translate those `let` bindings into `share`s, the others into `call`s.

## ToDo

- [x] Rename the typeclass `Shareable` to `Strategy`.
- [x] Add a function `call` to the `Strategy` typeclass.
- [x] Add a `Strategy` instance `Cbv` for call-by-value evaluation.

MajaRet:
### Issue

Closes #183 

### Description of the Change

The `Shareable` class has been renamed to `Strategy`.
Another function `call : forall {A : Type}, Free Shape Pos A -> Free Shape Pos (Free Shape Pos A)` has also been added to `Strategy` to translate `let` bindings whose value does not occur more than once.

In addition, a `Cbv` instance of `Strategy` for call-by-value evaluation has been added.

This pull request also features several more small fixes and changes:
- The argument order in smart constructors and handlers has been fixed. Before this, `A` was given before `Shape` and `Pos`.
- Since it is needed to write a theorem, Malias is now exported by `Free`.
- The tests for the base library have been moved to the `example` folder (`example/Base`). They now use list utility functions defined in `example/Data/List.hs`. Another module `example/Data/Tuple.hs` has also been added that provides utility functions for tuples.
- The `Traceable` class was removed as it only had one instance anyway. Instead of being given an (explicit) `Traceable` instance, functions should now have an (implicit) `Injectable Trace.Shape Trace.Pos Shape Pos` constraint for the tracing effect.
- The handling order of non-deterministic partial programs in the tests has been changed. Now, partiality is handled after non-determinism, which means that an error in any branch results in an error for the whole program. Tracing is still handled the same way as before (with partiality being handled before tracing) because in Haskell, even if an error occurs, the message log until that point is still displayed.

### Alternate Designs

This design lets us avoid choosing an evaluation strategy at compile time. Alternate designs could, for example, let the user choose an evaluation strategy via a compiler flag. However, this could lead to inconsistencies if programs compiled with different evaluation options are combined. 

The additional function `call` is used to avoid unnecessary applications of `share` in a call-by-need setting, which would complicate the program and thus any proofs about it.
Alternatively, we could remove `call` and instead always use `share` to translate `let` bindings.  

### Verification Process

The sharing handler tests (which have been renamed to `SharingTests` and moved to the example folder) now use `call` and `share` the way those functions would be used in a translated module (i.e. either `call` or `share` is used on every non-variable expression).
They still succeed.
A recursive function `tails` was also added to the tests and tested with all three evaluation strategies on non-determinism and tracing. The decreasing argument cannot be shared in the helper function due to Coq's termination checker (see #192), but since it is shared in the main function, sharing still succeeds.


base/coq/Free/Class/Strategy.v
```diff
@@ -0,0 +1,14 @@
...
+From Base Require Import Free.Class.ShareableArgs.
+
+Class Strategy (Shape : Type) (Pos : Shape -> Type) 
+                `{Injectable Share.Shape Share.Pos Shape Pos} :=
+  {
+    share : forall {A : Type} `{ShareableArgs Shape Pos A}, 
+      Free Shape Pos A -> Free Shape Pos (Free Shape Pos A);
+    call : forall {A : Type}, 
+      Free Shape Pos A -> Free Shape Pos (Free Shape Pos A);
+  }.
```
just95: Some files are missing a [trailing newline](https://github.com/FreeProving/guidelines/blob/main/CONTRIBUTING.md#include-newline-at-end-of-file). There are also [trailing spaces](https://github.com/FreeProving/guidelines/blob/main/CONTRIBUTING.md#avoid-trailing-whitespace) that should be removed.
Most editors can be configured to make these changes automatically when a file is saved but I think the CoqIDE does not have such an option. Consider to open and save modified Coq files in another editor before committing your changes.

base/coq/Free/Verification/SharingHandlerTests.v
```diff
@@ -32,7 +32,7 @@ Definition evalTracing {A : Type} p
 Definition evalNDM {A : Type} p
 := @collectVals (option A) (run (runChoice (runNDSharing (0,0) (runMaybe p)))).
 
-(* Shortcut to evaluate a traced partial program to a result and a list 
+(* Shortcut to evaluate a traced partial pro gram to a result and a list
```
just95:
```suggestion
(* Shortcut to evaluate a traced partial program to a result and a list 
```

base/coq/Prelude/Pair.v
```diff
@@ -16,12 +17,25 @@ Section SecPair.
...
+                    | pair_ x _ => x
+                    end.
+
+  (* Second element *)
+  Definition sndPair {A B : Type} (fp : Free' (Pair A B)) 
+   : Free Shape Pos B
+  := fp >>= fun p => match p with
+                    | pair_ _ y => y
+                    end.
+
```
just95: If we make these functions part of the `Prelude`, we should also expose them to the user by adding them to the `Prelude.toml` file.

MajaRet: For now, I've removed them from the `Prelude` again and put them in a utility module that the tests use.

base/coq/Free/Instance/ND.v
```diff
@@ -23,35 +23,35 @@ Module ND.
...
+    Definition Fail (Shape' : Type) (Pos' : Shape' -> Type)
+      `{Injectable Shape Pos Shape' Pos'} {A : Type}
       : Free Shape' Pos' A :=
       impure (injS sfail) (fun p => (fun (x : Void) => match x with end) (injP p)).
 
-    Definition Choice_ {A : Type} (Shape' : Type) (Pos' : Shape' -> Type) 
-    `{Injectable Shape Pos Shape' Pos'} mid l r
-    : Free Shape' Pos' A := 
-       let s := injS (schoice mid) 
+    Definition Choice_(Shape' : Type) (Pos' : Shape' -> Type)
```
just95:
```suggestion
    Definition Choice_ (Shape' : Type) (Pos' : Shape' -> Type)
```

base/coq/Free/Class/Traceable.v
```diff
@@ -1,7 +0,0 @@
-From Base Require Import Free.Monad.
-Require Export Coq.Strings.String.
-
-Class Traceable (Shape : Type) (Pos : Shape -> Type) :=
-  {
-    trace : forall {A : Type}, string -> Free Shape Pos A -> Free Shape Pos A
-  }.
```
just95: The problem with removing the `Traceable` type class is that we force the user to use our implementation of the `Trace` effect. They cannot provide their own implementation. Furthermore, the `Injectable` constraint adds a strong coupling between compiler and base library. Thus, I think that we shouldn't remove the type class even if we have only one implementation right now.

just95 (changes requested):

base/coq/Free/Class/Traceable.v
MajaRet: That makes sense, but I'm a little worried about proofs. 
Any traced value would essentially become a black box when we use an unknown `Traceable` instance, which might make a proof about a program with tracing very difficult.
Using a specific `Traceable` instance in a theorem about a translated QuickCheck property means that the `Shape` and `Pos` arguments (plus any constraints on `Shape` and `Pos`) of a property need to be quantified in the theorem itself because we need to specify `Shape` and `Pos` for a `Traceable` instance. 
But if that's acceptable, I agree that it does make sense to keep the class in the interest of flexibility.

Even without tracing, we do also have another `Injectable` constraint for the `Share` effect. We _could_ potentially get rid of that, as well, but that could make type signatures large and somewhat cryptic, and would likely make translation more complicated. (See conversation on [this PR](https://github.com/FreeProving/free-compiler/pull/173).)

nbun (approved):

just95 (approved):

PR #189, opened 2020-09-04
Implement default handling functions

### LINKED ISSUE #177: Implement Handler typeclass
## Description

A `Handler` typeclass should be implemented in Coq that describes programs that can be handled.

## Motivation

We currently have a handler for each possible effect (or the combination of sharing with an effect). To handle a program that contains several effects, handlers are combined.
A `Handler` typeclass that contains a function `handle` would provide instances for each possible combination of effects. This has two advantages:

- To handle a program, one no longer has to combine handlers by hand. Instead, any program can be handled simply by using the function `handle`. 
In cases where the handling order is important (for example, in programs that contain sharing), a Handler instances has the advantage of fixing the order of the handlers so mistakes do not happen, although it is also possible to define instances for different handling orders if that is desired.

- Unlike the programs themselves, handlers cannot work with flexible effect stacks. When a combination of handlers is used, the effect stack becomes fixed to a certain extent. If the program is evaluated to a pure value (i.e. all effects are handled), the entire effect stack is fixed.
As a consequence, if we wanted to include the handling of a program in our derived properties without a `Handler` class, we would need to use a concrete combination of handlers - therefore fixing the effect stack - and thus would not be able to quantify `Shape` and `Pos` anymore.
A `Handler` typeclass will allow us to simply add the constraint that the program have a `Handler` instance, which would not impede the quantification of `Shape` and `Pos`. Concrete handlers could then be used e.g. for counterexamples or to prove a property in a certain setting (e. g. a total setting).

MajaRet:
### Issue

Closes #177 

### Description of the Change

A set of handling functions has been defined. A handling function evaluates all effects of a particular effect stack and produces a pure value of a certain type.

For given `Shape` and `Pos`, a handling function has the following type:
```coq
handle : Free Shape Pos A -> B
```

`B` is the type `A` is transformed into when the effects defined by `Shape` and `Pos` are handled.
For example, handling non-determinism turns a program of type `A` into a value of type `list A`. 
Handling tracing turns a program of type `A` into a value of type `A * list string`, where `A` ist the type of the result and `list string` is the message log.

A function of this type will be included as a parameter of a translated property so we can reason about the _semantics_ of a program (statements about its results when evaluated).

Handling functions are built for fixed effect stacks by combining individual effect handlers (e.g. `runChoice` or `runMaybe`). Therefore, there is one default handling function for almost any combination of effects, but the order of the effects in the stack is fixed. 
To use an alternative order, one can simply combine effect handlers themselves.

There are no default handling functions for effect stacks that contain all three effects sharing, tracing and non-determinism, as only one effect can currently be shared. 

A sharing handler for effects that do not interact with sharing has also been implemented. Without it, the sharing effect would have had to be handled with either the tracing or non-determinism handler, which unnecessarily introduces those effects into the effect stack and is reflected in the handling result (for example, the result is wrapped in a list if the non-determinism sharing handler is used). 

### Alternate Designs

As an alternative to a handling function, we could also implement a `Handler` typeclass (and that is what was done at first). However, as inferring `Handler` instances turned out to have unwanted results and the possibility to infer instances was the main advantage of a class, the function approach has been chosen. This way, a handling function must always be given explicitly when reasoning about a concrete effect stack and can no longer be inferred.  

### Possible Drawbacks

Because one default handling function usually contains more than one effect, they cannot be defined in the individual effects' modules like the simple effect handler functions (`runChoice`, `runMaybe` etc.).
Therefore, they are defined centrally in a module `Handlers`, which will make _all_ default handling functions available when imported.

Since there is a default handler for almost any combination of effects (disregarding order), the number of functions roughly doubles with every new effect.
But as long as effects are not added too often, this should still be manageable.

### Verification Process

Handling in the test modules is now done by calling the appropriate default handling function on a program (except for cases of non-standard handling without normalization in `NormalizationTests`). 
The tests still succeed.


base/coq/Free/Handlers.v
```diff
@@ -0,0 +1,208 @@
+(** Definitions for the (p : Free class. *)
```
nbun:
```suggestion
(** Handler definitions for the Free class. *)
```

base/coq/Free/Handlers.v
```diff
@@ -0,0 +1,208 @@
...
+From Base Require Import Free.Instance.Trace.
+
+From Base Require Import Prelude.
+
+From Base Require Import Free.Util.Search.
+
+Require Import Coq.Lists.List.
+Import List.ListNotations.
+
+(* No effect *)
```
nbun: Sections would be useful here, IMO.

base/coq/Free/Handlers.v
```diff
@@ -0,0 +1,208 @@
...
+Definition STrcId := Comb.Shape Trace.Shape Identity.Shape.
+Definition PTrcId := Comb.Pos Trace.Pos Identity.Pos.
+
+Definition handleTrace {A B : Type}
+                      `{Normalform STrcId PTrcId A B}
+                       (p : Free STrcId PTrcId A) 
+  : (B * list string) := 
+  collectMessages (run (runTracing (nf p))).
+
+(* Share :+: Identity handler *)
```
nbun: I guess this definition is technically valid, but is there a reason why one would have only sharing without a shared effect in the effect stack?

base/coq/Free/Handlers.v
```diff
@@ -0,0 +1,208 @@
...
+Definition SShrTrc := Comb.Shape Share.Shape (Comb.Shape Trace.Shape Identity.Shape).
+Definition PShrTrc := Comb.Pos Share.Pos (Comb.Pos Trace.Pos Identity.Pos).
+
+Definition handleShareTrace {A B : Type}
+                            `{Normalform SShrTrc PShrTrc A B}
+                            (p : Free SShrTrc PShrTrc A)
+  : (B * list string) := 
+  collectMessages (run (runTracing (runTraceSharing (0,0) (nf p)))).
+
+(* Maybe :+: Share :+: Identity handler *)
```
nbun: Why is `Share` not on top of the effect stack as for `handleShareND` and `handleShareTrace`?

nbun (changes requested):

base/coq/Free/Handlers.v
MajaRet: Every translated function that contains either the `call` or the `share` function (which is basically every function for a flattened program) has a constraint for the `Share` effect, so the `Share` effect will _always_ be in the effect stack, whether there are any other effects or not. 
`cbneed` works the same on any value, whether it contains a relevant effect or not, so actual sharing syntax is always introduced when call-by-need evaluation is used.
So we still need a handler that removes the sharing syntax (though it _could_ probably be defined simpler than here).

base/coq/Free/Handlers.v
MajaRet: `Share` is not always the topmost effect. `Maybe` needs to be on top if `Trace` is in the effect stack along with sharing, it needs to be at the bottom (just above `Identity`) if `ND` is in the effect stack along with sharing, but in this case I don't think it matters what the order is. 
I could change it, though.

base/coq/Free/Handlers.v
MajaRet: Oh, that was search-and-replace gone wrong. It should actually say something like "This module contains default handlers for individual effect stacks." or something.

base/coq/Free/Handlers.v
nbun: I see, good point.

base/coq/Free/Handlers.v
nbun: For consistency's sake I would prefer a version with `Share` on top.

nbun (approved):

MarvinLira (approved):

PR #194, opened 2020-09-11
Integrate Error effect

### LINKED ISSUE #119: Define new smart constructors and a handler for the Error effect
## Description

We should define smart constructors that produce values embedded in an effect stack for the `Error` effect. This means that instead of `Free Error.Shape Error.Pos A` (aka `Error A`), the smart constructors' return type should be `Free Shape Pos A` with the constraint that an instance `Injectable Error.Shape Error.Pos Shape Pos` exists.

Additionally, a handler for programs containing the `Error` effect should be implemented. 

## Motivation

Currently, we have extended the smart constructors and defined handlers for all effects except for `Error`. Therefore, we should do the same for `Error`.

## Dependencies

This issue does not depend on any other open issues.

MajaRet:
### Issue

Closes #119 

### Description of the Change

The `Error` effect has been integrated fully into the base library by doing the following.
- The smart constructors `NoError` and `ThrowError` and the effect's `Partial` instance have been modified so that they now embed the effect in an effect stack.
- An effect handler `runError` has been added that evaluates a program with the `Error` effect either to a value or an error message.
- Default handlers for effect stacks containing the `Error` effect have been added to `Free.Handlers`. 

### Verification Process

Two test cases using the `Partial` instance of `Error` were added to `NormalizationTests`.

The example proofs in `ConsUnconsProofs`, which use the `Error` effect's smart constructors and `Partial` instance, have been adjusted and still succeed.


example/Proofs/ConsUnconsProofs.v
```diff
@@ -39,16 +39,16 @@ Section ErrorMessages.
 
   (* To prove facts about the error messages we can write an abbreviation for
      an [error] with a specific message. *)
-  Definition EmptyListError {A : Type} := @error _ _ Error.Partial A "unconsE: empty list"%string.
+  Definition EmptyListError {A : Type} := @error _ _ (Error.Partial _ _) A "unconsE: empty list"%string.
 
   (* If we weren't looking for an actual [error] but for an [undefined] in haskell
```
nbun:
```suggestion
  (* If we weren't looking for an actual [error] but for an [undefined] in Haskell
```

example/Base/NormalizationTests.v
```diff
@@ -44,8 +45,9 @@ Definition evalNDMaybe {A : Type} p
 Definition IdS := Identity.Shape.
 Definition IdP := Identity.Pos.
 
-(* Infer Shape and Pos for the Maybe Partial instance for convenience- *)
+(* Infer Shape and Pos for Partial instances for convenience- *)
```
nbun:
```suggestion
(* Infer Shape and Pos for Partial instances for convenience. *)
```

nbun (approved):

MarvinLira (approved):

PR #198, opened 2020-09-14
Add support for case expressions with single variable pattern alternative 

### LINKED ISSUE #74: Add support for case expressions with single variable pattern alternative
## Description

Support for case expressions of the following form should be added.

```haskell
case e of
  x -> e'
```

According to the [Haskell 2010 Language Report (p. 29)](https://www.haskell.org/definition/haskell2010.pdf)

> 1. Matching the pattern `var` against a value `v` always succeeds and binds `var` to `v`.

Thus, the following translation should be performed in the [conversion from Haskell to IR](https://github.com/FreeProving/free-compiler/blob/master/src/lib/FreeC/Frontend/Haskell/Simplifier.hs).

```haskell
case e of { x -> e' } = (\x -> e') e
```

This translation is also safe in the presence of effects as demonstrated by the following example.

```haskell
GHCi> case undefined of { x -> () }
()
```

## Motivation

When the pattern matching compiler is enabled and a function that contains guards is translated, `case` expressions of the form mentioned above are generated. However, the Free Compiler supports only case expressions of the following form at the moment.

```haskell
case e of
  C₁ x₍₁,₁₎ … x₍₁,ₚ₁₎-> e₁
  ⋮
  Cₙ x₍ₙ,₁₎ … x₍ₙ,ₚₙ₎-> eₙ
```

## Alternatives

The generation of such case expressions could be removed from the pattern matching compiler. However, adding this feature to the Free Compiler may also be useful in other scenarios.

just95:
### Issue

Closes #74.

### Description of the Change

The transformation from the Haskell AST to the IR AST has been modified such that `case`-expressions of the form

```haskell
case e { x -> e' }
```

are transformed to immediately applied lambda abstractions.

```haskell
(\x -> e') e
```

### Verification Process

A test case for this transformation has been added. Since no tests for the `Simplifier` existed before, this is the first test for the transformation into IR. More test should be added in the future.


MajaRet (approved):

src/test/FreeC/Frontend/Haskell/SimplifierTests.hs
```diff
@@ -0,0 +1,45 @@
...
+-------------------------------------------------------------------------------
+-- Utility functions                                                         --
+-------------------------------------------------------------------------------
+-- | Parses a Haskell expression and converts it to IR.
+parseAndSimplifyExpr :: String -> Simplifier IR.Expr
+parseAndSimplifyExpr = (parseHaskell . mkSrcFile "<test-input>")
+  >=> simplifyExpr
+
+-- | Parses the given Haskell and IR expressions, converts the Haskell
+--   expression to IR .
```
stu204767: This seems incomplete:
```suggestion
--   expression to IR and sets the expectation that the given IR expression is
--   produced.
```
At the very least, the space after `IR` has to be removed.

src/test/FreeC/Frontend/Haskell/SimplifierTests.hs
```diff
@@ -0,0 +1,45 @@
...
+-------------------------------------------------------------------------------
+-- | Test group for "FreeC.Frontend.Haskell.Simplifier" tests.
+testSimplifier :: Spec
+testSimplifier = describe "FreeC.Frontend.Haskell.Simplifier" $ do
+  testSimplifyExpr
+
+-- | Test group for 'simplifyExpr' tests.
+testSimplifyExpr :: Spec
+testSimplifyExpr = context "simplifyExpr" $ do
+  it "simplifies single variable pattern case expression to lambda abstractions"
```
stu204767: I think the hyphens make this easier to understand, but you don't have to add them. `expression` should be in the plural, however.
```suggestion
  it "simplifies single-variable-pattern case expressions to lambda abstractions"
```
With this change, the line becomes 82 characters long, though.

src/test/FreeC/Frontend/Haskell/Tests.hs
```diff
@@ -0,0 +1,11 @@
+-- | Test group for tests of modules below @FreeC.Frontend.Haskell@.
```
stu204767: This formulation is probably used for many test modules, but I don't think `below` fits here, as I couldn't find a dictionary mentioning affiliation as one of the meanings of this word. You could use `of`, `belonging to` or something else instead.

stu204767 (changes requested):

src/test/FreeC/Frontend/Haskell/Tests.hs
just95: See 76f018e97820f0fffe9b197b6a67be1eeb34f2ee.

stu204767 (approved):

PR #203, opened 2020-09-15
Implement translation of tracing effect

### LINKED ISSUE #134: Detect effects in functions
## Description

We should annotate in the environment whether a function contains an effect or depends on a function that contains that effect. Additionally, a compiler pass should be added that determines which functions contain which effects directly or indirectly.

For details about how that could be done, see @just95's comment below.

## Motivation

In Coq, a function that contains an effect must be given an instance of a corresponding typeclass as an argument. Similarly, functions that depend on such a function also need to be given that instance.
We currently analyze functions for the partiality effect and denote it in the environment via a marker `isPartial`. We should do the same for other supported effects (currently sharing and tracing), and ideally group these analyses together into a single effect analysis pass.

## ToDo (courtesy of @just95)
 - [x] Add `IR.Trace` constructor to [`IR.Expr`](https://github.com/FreeProving/free-compiler/blob/bb31c765bad641d45f28d73d89fe54f225c3baeb/src/lib/FreeC/IR/Syntax/Expr.hs#L18).
 - [x] Add `trace` to the [IR parser](https://github.com/FreeProving/free-compiler/blob/bb31c765bad641d45f28d73d89fe54f225c3baeb/src/lib/FreeC/Frontend/IR/Parser.hs#L577).
 - [x] Add `trace` to the Haskell to IR [`Simplifier`](https://github.com/FreeProving/free-compiler/blob/bb31c765bad641d45f28d73d89fe54f225c3baeb/src/lib/FreeC/Frontend/Haskell/Simplifier.hs#L616).
 - [x] Add `Tracing` constructor to [`LIR.Effect`](https://github.com/FreeProving/free-compiler/blob/bb31c765bad641d45f28d73d89fe54f225c3baeb/src/lib/FreeC/LiftedIR/Effect.hs#L9).
 - [x] Replace [`isPartial`](https://github.com/FreeProving/free-compiler/blob/bb31c765bad641d45f28d73d89fe54f225c3baeb/src/lib/FreeC/Environment/Entry.hs#L94) by a field of type `[LIR.Effect]`.
    + [x] Update the [encoder](https://github.com/FreeProving/free-compiler/blob/main/src/lib/FreeC/Environment/ModuleInterface/Encoder.hs) and [decoder](https://github.com/FreeProving/free-compiler/blob/main/src/lib/FreeC/Environment/ModuleInterface/Decoder.hs) for module interface files.
    + [x] Document the changes in `./doc/ModuleInterfaceFileFormat.md`.
    + [x] Bump the module interface file format version.
    + [x] Update `Prelude.toml` and `QuickCheck.toml`.
 - [x] Replace [`PartialityAnalysisPass`](https://github.com/FreeProving/free-compiler/blob/bb31c765bad641d45f28d73d89fe54f225c3baeb/src/lib/FreeC/Pass/PartialityAnalysisPass.hs) by an `EffectAnalysisPass`.

just95:
### Issue

Closes #134.

### Description of the Change

 - AST nodes for the tracing effect have been added to the IR and lifted IR.
 - Translation of `trace` to Coq has been implemented. The Agda back end reports a fatal error when `trace` is used.
 - Support for `trace` has been added to the Haskell and IR front ends.
 - The `EffectAnalysisPass` and module interface file decoder have been extended such that they work with arbitrary effects.

### Alternate Designs

In theory it would also be possible to remove `trace`, `error` and `undefined` from the IR and add them to the base library's `.toml` file instead. However, since `trace` and `error` have a `String` argument but strings are not yet supported, this is not possible at the moment. The `Debug.Trace` module has already been added to the base library but is empty at the moment. Without this module, imports of `Debug.Trace` would not be allowed which means that modules that use `trace` could not be compiled with the GHC.

### Possible Drawbacks

Because `trace` is implemented as special syntax, it is available in every module, i.e., no import of `Debug.Trace` is necessary.

### Verification Process

Test cases for `trace` have been added to the tests of all modified modules that had tests for `error` or `undefined` already.


src/lib/FreeC/IR/Syntax/Expr.hs
```diff
@@ -250,7 +256,10 @@ prettyExprPred' n expr@(TypeAppExpr _ e t _)
   | n <= 1 = prettyExprPred 1 e <+> char '@' <> prettyTypePred 2 t
   | otherwise = parens (prettyExprPred' 1 expr)
 prettyExprPred' n expr@(ErrorExpr _ msg _)
-  | n <= 1 = prettyString "error" <+> prettyString (show msg)
+  | n <= 1 = prettyString "error" <+> prettyString msg
```
Daniel-Teut: `show` must be called on `msg`, otherwise the message string is missing `"` in the output (same with the `Trace` case).
The `Trace` is also missing a specific case for visible type application resulting in the wrong order of arguments, i.e. currently `trace "Hello" x @a` is generated and should be `trace @a "Hello" x`.

src/exe/Main.hs
```diff
@@ -198,53 +196,48 @@ loadImport decl = do
...
   findIfaceFile :: [FilePath] -> Application FilePath
   findIfaceFile []       = reportFatal
     $ Message srcSpan Error
-    $ "Could not find imported module " ++ showPretty modName
+    $ "Could not find imported module " ++ showPretty modName ++ "."
   findIfaceFile (d : ds) = do
-    let ifaceFile = d </> filename
-    exists <- liftIO $ doesFileExist ifaceFile
-    if exists then return ifaceFile else findIfaceFile ds
+    let ifaceFiles = [d </> filename <.> ext | ext <- extensions]
```
Daniel-Teut: I think I would be more readable to also move the `ds` argument into the list comprehension and use a `case` on the result of `findM`.

Daniel-Teut (changes requested):

Daniel-Teut (approved):

maclement (approved):

PR #175, opened 2020-08-26
Generate typeclass instances

MajaRet:
### Issue

Closes #150
Closes #151 

### Description of the Change

The compiler now generates `Normalform` and `ShareableArgs` instances for user-defined data types.

In order to properly handle a program, it needs to be normalized first so that handlers do not have to consider effects in the _components_ of data structures. 
To be able to perform this normalization, a data type (and its argument types) must have a `Normalform` instance.

Furthermore, the `share` operator requires that the type of the computation that is shares has a `ShareableArgs` instance so its components can be shared.

`Normalform` and `ShareableArgs` instances exist for all Prelude types, but users will also define their own Haskell data types. Therefore, the compiler needs to generate these instances for user-defined data types.

### Verification Process

Tests verifying the functionality of generated `Normalform` instances for various types have been added to the `example` folder by @marvin2706. Additional functions and tests have been added (in the same modules) to test the generated `ShareableArgs` instances.

### Additional Notes

Because the generation of induction schemes has been moved into its own module, there will likely be merge conflicts with #206.

PR #205, opened 2020-09-17
Incorporate handling into theorems

### LINKED ISSUE #202: Add handlers and evaluation strategy to theorems about translated properties
## Description
[WIP]
We should modify how `Prop`s are produced from `Property`s in two ways:
- A handler is introduced as an additional argument. Handlers should be universally quantified in effect-generic proofs or can be given explicitly to conduct proofs about the semantics of a program in a specific setting (effect stack).
- An evaluation strategy is chosen explicitly. Because strategy-agnostic proofs do not make much sense, an evaluation strategy should always be chosen explicitly when one writes a theorem about a translated property.

## Motivation
[WIP]
A QuickCheck property is translated into a function whose return value is a `Free Property`, where `Property` is a synonym for Coq's `Prop` type. 
These properties can be transformed into `Prop`s by using the `quickCheck` function on them, which allows us to actually write proofs for them.

Currently, programs are not handled, meaning we prove properties directly about `Free` values. However, this becomes very limiting when sharing is introduced, as, for example, often two `Free` values that are not _syntactically_ the same are still _semantically_ equivalent (for some evaluation strategies).

So we should make it possible to reason about handled programs instead of `Free` values.

## Alternatives
-
## Additional Context
-

MajaRet:
### Issue

Closes #202 

### Dependencies

#175 

### Description of the Change

It is now possible to use handlers in theorems. To achieve that, several changes were made.

The default handling functions have been moved back into a `Handler` class that is parameterized over a single type variable (and `Shape` and `Pos`). The return type of the handling function is accessible via the constant `handledType` that is also provided by the `Handler` class.

The evaluation strategy's type has been changed from `Strategy Shape Pos` to `forall (Shape : Type) (Pos : Shape -> Type) {Injectable Share.Shape Share.Pos Shape Pos}`, mainly for convenience reasons. It is now possible to pass simply `Cbv`, `Cbv` or `Cbneed` as an argument to a property or function instead of `Cbv Shape Pos` etc. 
To improve readability, a notation "EvaluationStrategy" has been defined for the type.

The strategy remains tied to the sharing effect and is given as the fourth argument (the third explicit argument). 
So in order to instantiate the evaluation strategy in an effect-generic theorem, it is necessary to write it as follows.

```Coq
Theorem myTheorem : quickCheck (fun Shape Pos I => @myProp Shape Pos I Cbneed).
```

This is not ideal, but moving the evaluation strategy to the front caused too many unwelcome changes.

For now, `quickCheck` still translates properties without handling anything (though the `Normalform` constraints are added, as the `Free` values are actually 'handled', but with the identity function). There is another function `quickCheckHandle` that allows a handler to be passed as an argument and then handles the program(s) according to that handler.
To use this function, it is necessary to explicitly infer the `Shape` and `Pos` arguments of the property (or write the exact effect stack of the handler). 
So it is used like this:
```Coq
Theorem handledTheorem : quickCheckHandle (myProp _ _ Cbneed) HandlerShareMaybe.
```

There is also a third function `quickCheckQuantifyHandler` that universally quantifies the handler. However, proving equalities with a universally quantified handler is equivalent to proving the equality of its arguments, and proving inequalities should be impossible (as the handler could theoretically be instantiated with a constant function, so everything is equal). So unless we add some properties to the `Handler` class and prove them for the instances, the function's usefulness is questionable.

### Possible Drawbacks
Because the current handlers normalize (strictly evaluate) the result of a computation before handling it, they induce strictness in properties/proofs (see #208).
This leads to some counterexamples becoming invalid.
However, it remains possible to reason about unhandled programs.

### Verification Process
The proofs in the example folder have been adjusted and still work. They all use `quickCheck`, meaning the programs we are reasoning about are not handled.

`quickCheckHandle` has been tested "by hand", but useful test cases can only really be added once other functionality (tracing, sharing analysis, expression flattening) is merged.


base/coq/Test/QuickCheck.v
```diff
@@ -77,37 +147,44 @@ Section SecQuickCheck.
...
+   := pure (fun handler => toProperty fb handler -> toProperty p handler).
 
   (* [(===) :: a -> a -> Property] *)
-  Definition eqProp (A : Type) (fx : Free' A) (fy : Free' A)
+  Definition eqProp (A : Type) `(NF : Normalform Shape Pos A)
+    (fx : Free' A) (fy : Free' A)
     : Free' Property'
-   := pure (fx = fy).
+   := pure (fun handler => 
+             @handle Shape Pos A (handler A NF) fx = handle fy).
```
MarvinLira: It seems kind of odd to use the explicit call for `handle` here but not in the other definitions like `neqProp`.
```suggestion
             handle fx = handle fy).
```

src/lib/FreeC/Backend/Coq/Base.hs
```diff
@@ -191,10 +191,16 @@ strategy = Coq.bare "Strategy"
 strategyArg :: Coq.Qualid
 strategyArg = Coq.bare "S"
 
--- | The Coq binder for the @Strategy@ type class.
+-- | A notation for the term
+--   (forall (Shape : Type) (Pos : Shape -> Type)
+--    `{Injectable Share.Shape Share.Pos Shape Pos},
+--    Strategy Shape Pos).
+strategyNotation :: Coq.Qualid
+strategyNotation = Coq.bare "EvaluationStrategy"
```
MarvinLira, line 223: This identifier should be added to the list of reserved identifiers.

MarvinLira (changes requested):

MarvinLira (approved):

PR #207, opened 2020-09-18
Add `call` operator. 

### LINKED ISSUE #196: Add `call` operator to the lifted IR
## Description
Add a `Call` construct to the lifted IR and an appropriate `call` operator to the Coq translation.
Further alter the lifting of IR `Let` such that `let`s with only one variable are translated to `Call` instead of `Share`.

## Motivation
#184 implemented a new approach to the evaluation strategy selection.
This approach needs the aforementioned `call` operator.

Daniel-Teut:
### Issue
Closes #196.

### Description of the Change
Adds the `call` operator to the compiler and changes the `let` translation in such a way, that the correct `share` or `call` operator is selected.

### Verification Process
Several test case were written to verify the correct selection of operators.


src/lib/FreeC/LiftedIR/Converter/Expr.hs
```diff
@@ -332,14 +332,51 @@ rawBind srcSpan mx x varType expr = do
...
+  expr' <- liftBinds bs expr
+  patType' <- mapM LIR.liftType varPatType
+  varPat' <- makeVarPat patSrcSpan (IR.varPatQName varPat) patType'
+  shareType' <- mapM LIR.liftType' varPatType
+  bindExpr' <- liftExpr bindExpr
+  let countExprs = expr : map IR.bindExpr bs
+      shareOp
+        = if sum (map (countVarInExpr $ IR.varPatQName varPat) countExprs) > 1
+          then LIR.Share
+          else LIR.Call
```
just95: Maybe break this up to make the condition easier to understand.

```suggestion
  let varName    = IR.varPatQName varPat
      countExprs = expr : map IR.bindExpr bs
      count      = sum (map (countVarInExpr varName) countExprs)
      shareOp    | count > 1 = LIR.Share
                 | otherwise = LIR.Call
```

src/lib/FreeC/LiftedIR/Converter/Expr.hs
```diff
@@ -332,14 +332,51 @@ rawBind srcSpan mx x varType expr = do
...
+
+-- | Counts the number of times the variable with the given qualified name
+--   occurs in the given expression.
+countVarInExpr :: IR.QName -> IR.Expr -> Int
+countVarInExpr varPat = countVarInExpr'
+ where
+  countVarInExpr' :: IR.Expr -> Int
+  countVarInExpr' IR.Con {}                    = 0
+  countVarInExpr' (IR.Var _ varName _)
+    = if varPat == varName then 1 else 0
```
just95: Consider to rename `varPat` since it is not an `IR.VarPat` but the `IR.VarName` of an `IR.VarPat`.

```suggestion
countVarInExpr :: IR.QName -> IR.Expr -> Int
countVarInExpr varName = countVarInExpr'
 where
  countVarInExpr' :: IR.Expr -> Int
  countVarInExpr' IR.Con {}             = 0
  countVarInExpr' (IR.Var _ varName' _) | varName == varName' = 1
                                        | otherwise = 0
```

src/lib/FreeC/LiftedIR/Converter/Expr.hs
```diff
@@ -332,14 +332,51 @@ rawBind srcSpan mx x varType expr = do
...
+    = countVarInExpr' cond + countVarInExpr' true `max` countVarInExpr' false
+  countVarInExpr' (IR.Case _ expr alts _)      = countVarInExpr' expr
+    + maximum
+    (map (\(IR.Alt _ _ varPats rhs) -> countVarInBinds varPats rhs) alts)
+  countVarInExpr' IR.Undefined {}              = 0
+  countVarInExpr' IR.ErrorExpr {}              = 0
+  countVarInExpr' IR.IntLiteral {}             = 0
+  countVarInExpr' (IR.Lambda _ varPats expr _) = countVarInBinds varPats expr
+  countVarInExpr' (IR.Let _ binds expr _)      = countVarInExpr' expr
+    + sum (map (countVarInExpr' . IR.bindExpr) binds)
```
just95: Since the `let`-expression could shadow the counted variable, `countVarInBinds` should be used here as well.

src/lib/FreeC/LiftedIR/Converter/Expr.hs
```diff
@@ -332,14 +332,51 @@ rawBind srcSpan mx x varType expr = do
...
+
+  -- | Returns the number of all occurrences of the variable with the given name
+  --   in the given expression.
+  --
+  --   Returns @0@ if the variable occurs in the given variable patterns.
+  countVarInBinds :: [IR.VarPat] -> IR.Expr -> Int
+  countVarInBinds varPats exprs
+    | any (\varPat' -> IR.varPatQName varPat' == varPat) varPats = 0
```
just95: This should be equivalent to the following.

```suggestion
    | varPat `elem` map IR.varPatQName varPats = 0
```

src/lib/FreeC/LiftedIR/Converter/Expr.hs
```diff
@@ -332,14 +332,51 @@ rawBind srcSpan mx x varType expr = do
...
+    | otherwise = countVarInExpr varPat exprs
```
just95: Since `varPat` is not changed, you can call `countVarInExpr` directly. The variable `exprs` should be renamed to `expr` since it is a single expression.

```suggestion
    | otherwise = countVarInExpr' expr
```

just95 (changes requested):

maclement (approved):

just95 (approved):

PR #212, opened 2020-09-24
Update Pattern Matching Compiler

### LINKED ISSUE #71: Parameterize IR modules over their contents
## Description

The [`IR.Module`](https://freeproving.github.io/free-compiler/docs/master/FreeC-IR-Syntax-Module.html#t:Module) data type should be expressed in terms of a polymorphic data type that is parameterized over a data type that determines the contents of the module.

```haskell
data ModuleOf contents = ModuleOf { …, modContents :: contents }
data IR.TopLevelDecl = …
type Module = ModuleOf [TopLevelDecl]
```

## Motivation

Dependency analysis currently has to be performed in multiple places of the compiler (for example, [in the compiler passes](https://github.com/FreeProving/free-compiler/blob/master/src/lib/FreeC/Pass/DependencyAnalysisPass.hs) and [in the back end](https://github.com/FreeProving/free-compiler/blob/1b666387fe35fca42652471b6e46c8d3f1305c03/src/lib/FreeC/Backend/Coq/Converter/FuncDecl.hs#L19)). This is mainly because the information about strongly connected components of the dependency graph cannot be stored within `IR.Module`s. In order to perform dependency analysis only once, lists of `DependencyComponent`s have to be stored in `IR.Module`s instead of lists of declarations after dependency analysis has been performed. In this first step, the AST should be extended to allow for such a generalization of dependency analysis in the future. 

In theory it would work to parameterize `IR.Module` over a component data type (i.e., `Identity` for the current behavior and `DependencyComponent` for the analysed modules). However, it would be more flexible if we move the entire contents of the module (i.e., everything except for it's name and `SrcSpan`) into a parameter.
I'd recommend to add a new polymorphic data type `IR.ModuleOf contents` and express `IR.Module` in terms of this new data type. `IR.ModuleOf IR.FuncDecl` should represent a module that contains a single function declaration, for example.

Most of the existing code should not be affected by this change. Apart of `FreeC.IR.Syntax.Module`, I expect that only code that uses record updates to overwrite module contents has to be changed.

## Tasks

 - [x] Add a polymorphic data type `IR.ModuleOf contents`.
 - [x] Add a data type `IR.TopLevelDecl` for all declarations that can occur on top-level of IR modules with record fields and predicates for each constructor.
 - [x] Express `IR.Module` in terms of `IR.ModuleOf [IR.TopLevelDecl]`.
 - [x] Express getters for content fields (i.e., `IR.modImports` through `IR.modFuncDecls`) of `IR.Module` using `map` and `filter`.
 - [x] Replace uses of record updates for `IR.Module` with the more general record updates for `IR.ModuleOf`.

### LINKED ISSUE #210: Update Pattern Matching Compiler
## Description

Update the Haskell frontend to use the latest version of the pattern matching compiler.
Known bugs of the pattern matching compiler that have been fixed in this version should be removed from the documentation.

## Motivation

During the master's project bugs in the pattern matching compiler have been fixed and new features have been implemented. However, the Free Compiler still uses version 0.1.1.0 of the pattern matching compiler from before the master's project. In order to profit from the improvements of the pattern matching compiler, we have to update the corresponding parts of the Haskell frontend.

just95:
### Issue

Closes #210.
Closes #71.

### Description of the Change

 - [x] The `haskell-src-transformations` dependency has been updated to the unreleased development version from the [v0.2.0.0-rc branch](https://github.com/FreeProving/haskell-src-transformations/tree/v0.2.0.0-rc).
 - [x] The new effects of the pattern matching compiler are initialized properly.
   + [x] The `Report` effect is handled in terms of `MonadReporter`.
   + [x] The `Env` effect is initialized with module interfaces that are generated from the module interfaces of the `Converter`'s environment.
   + [x] The `Fresh` effect is initialized with the identifiers that occur in the module. Alternatively, we could use our own generator for fresh variables.
   + [x] The `InputFile` effect is initialized with the file name and contents from the transformed modules source span.
 - [x] The [documentation](https://github.com/FreeProving/free-compiler/blob/main/doc/ExperimentalFeatures/PatternMatchingCompilation.md#limitations-and-known-bugs) of limitations and known bugs of the pattern matching compiler has been updated.
 - [x] Since source spans are not discarded by the pattern matching compiler anymore, the help message for `--transform-pattern-matching` has been updated.
 - [x] The module dependency analysis and loading of module interfaces is now performed before the pattern matching compilation. 

### Verification Process

 - [x] An example which uses pattern matching and imports a data type from another module was added.

### Additional Notes

 - [x] Once `v0.2.0.0` of the pattern matching compiler has been released, the version and tag have to be updated in the Cabal configuration accordingly.


example/PatternMatching.hs
```diff
@@ -15,6 +17,15 @@ plus :: Peano -> Peano -> Peano
 plus Zero q     = q
 plus (Succ p) q = plus p (Succ q)
 
+-------------------------------------------------------------------------------
+-- Impoted Data Types                                                        --
```
stu204767:
```suggestion
-- Imported Data Types                                                       --
```

example/PatternMatching.hs
```diff
@@ -53,23 +64,20 @@ unzip :: [(a, b)] -> ([a], [b])
 unzip []             = ([], [])
 unzip ((x, y) : xys) = case unzip xys of
   (xs, ys) -> (x : xs, y : ys)
+
 -------------------------------------------------------------------------------
 -- Guards                                                                    --
 -------------------------------------------------------------------------------
 -- The following two functions cannot be translated at the moment, because the
 -- pattern matching compiler generates @let@ expressions when eliminating guards
 -- but our compiler does not support local declarations.
```
stu204767: The following examples seem to work now, so this comment should be adjusted.

src/exe/FreeC/Frontend.hs
```diff
@@ -54,45 +76,60 @@ defaultFrontend = frontendName haskellFrontend
...
     case maybeDumpDir of
-      Nothing      -> return haskellAst'
+      Nothing      -> return inputModule { IR.modContents = outputModule }
       Just dumpDir -> do
         -- Generate name of dump file.
-        modName <- liftConverter $ extractModName haskellAst'
-        let modPath  = map (\c -> if c == '.' then '/' else c) modName
-            dumpFile = dumpDir </> modPath <.> "hs"
+        let modName = IR.modName inputModule
+        let modPath      = map (\c -> if c == '.' then '/' else c) modName
```
stu204767: The second `let` here is superfluous.
```suggestion
        let modName      = IR.modName inputModule
            modPath      = map (\c -> if c == '.' then '/' else c) modName
```

src/exe/Main.hs
```diff
@@ -156,27 +163,34 @@ convertInputModule converter ast = do
...
---   with the given name.
+-- | Output a module that has been generated by the given backend from an IR
+--   module with the given name.
 --
---  The generated file has the given file extension.
-outputModule :: String -> IR.ModName -> String -> Application ()
-outputModule ext modName outputStr = do
+--   If the @--output@ command line option was specified, the output module is
+--   written to a file whose name is based on the module name and whose
+--   extension depends on the given back end. Otherwise, the output module is
```
stu204767, line 179: If there was more time, I would suggest spelling "front end" and "back end" with a space in all comments (most are spelled without the space currently). But I think this can be dismissed for today.

src/lib/FreeC/Frontend/Haskell/PatternMatching.hs
```diff
@@ -1,117 +1,273 @@
...
+import           Polysemy
+  ( Member, Members, Sem, interpret, runM )
+import           Polysemy.Embed                    ( Embed, embed )
 
-import           FreeC.Environment
 import           FreeC.Environment.Entry
+import           FreeC.Environment.LookupOrFail
 import           FreeC.Environment.ModuleInterface
-import qualified FreeC.IR.Base.Prelude             as IR.Prelude
+import           FreeC.IR.Base.Prelude             as IR.Prelude
```
stu204767: Was this qualification removed intentionally?

src/lib/FreeC/Frontend/Haskell/PatternMatching.hs
```diff
@@ -1,117 +1,273 @@
...
--- | Converts an entry of the 'Environment' to an entry of the constructor map
---   for the 'initialState'.
-makeConsMapEntry :: EnvEntry -> Maybe (String, [(HSE.QName (), Int, Bool)])
-makeConsMapEntry entry
-  | not (isConEntry entry) = Nothing
-  | otherwise = do
-    typeConIdent <- extractTypeConIdent (entryReturnType entry)
-    return (typeConIdent, [(conQName, arity, isInfix)])
+--   Since the pattern matching compiler library does not support source
+--   spans, location information is removed during the transformation.
```
stu204767: The pattern matching compiler does support source spans, so what do you mean here?

src/lib/FreeC/Frontend/Haskell/PatternMatching.hs
```diff
@@ -1,117 +1,273 @@
...
+    return (ToHSE.transformModule outputModule)
+
+-------------------------------------------------------------------------------
+-- Source Spans                                                              --
+-------------------------------------------------------------------------------
+-- | Front end configuration type for the @haskell-src-exts@ front end with
+--   source spans of type 'SrcSpan'.
+type Frontend = HSE SrcSpan
+
+-- | Converts a source span to a HST source span.
```
stu204767:
```suggestion
-- | Converts a source span to an HST source span.
```

src/lib/FreeC/Frontend/Haskell/PatternMatching.hs
```diff
@@ -1,117 +1,273 @@
...
+initEnv inputModule@(HST.Module _ _ _ imports _) = do
+  let importSrcSpans = map (ToHSE.transformSrcSpan . HST.importSrcSpan) imports
+      importNames    = map (HST.prettyName . HST.importModule) imports
+  ifaces
+    <- embed $ zipWithM lookupAvailableModuleOrFail importSrcSpans importNames
+  preludeIface
+    <- embed $ lookupAvailableModuleOrFail NoSrcSpan IR.Prelude.modName
+  return HST.Environment
+    { HST.envCurrentModule   = HST.createModuleInterface inputModule
+    , HST.envImportedModules = zip imports (map convertModuleInterface ifaces)
```
stu204767: To adapt to the new `Environment` data type of the PMC, this could be changed to something like this:
```suggestion
    , HST.envImportedModules = zipWith (\import iface -> ([import], convertModuleInterface iface) imports ifaces
```
This needs formatting, of course, and it assumes that no module interface is imported more than once.

src/lib/FreeC/Frontend/Haskell/PatternMatching.hs
```diff
@@ -1,117 +1,273 @@
...
+-- | Converts the given module interface to a module interface of the pattern
+--   matching compiler.
+convertModuleInterface :: ModuleInterface -> HST.ModuleInterface Frontend
+convertModuleInterface iface = HST.ModuleInterface
+  { HST.interfaceModName   = Just
+      (HST.ModuleName HST.NoSrcSpan (interfaceModName iface))
+  , HST.interfaceDataCons  = Map.fromList
+      (map (convertQName *** convertDataEntryCons) (Map.assocs dataEntries))
+  , HST.interfaceTypeNames = Map.fromList
+      (map (convertQName *** convertConEntryType) (Map.assocs conEntries))
```
stu204767: Since the names and types of these maps have been changed, this part has to be reworked as well, together with the `convertDataEntryCons` function (which should be renamed to `convertDataEntry`). For the second map, `convertConEntry` should be used instead of `convertConEntryType`).

src/lib/FreeC/Frontend/Haskell/PatternMatching.hs
```diff
@@ -1,117 +1,273 @@
...
+    Just specialName -> HST.Special HST.NoSrcSpan specialName
     Nothing          -> case qName of
       (IR.UnQual name) -> convertName name
       (IR.Qual _ name) -> convertName name
 
-  -- | Converts an unqualified IR name to a Haskell name.
-  convertName :: IR.Name -> HSE.QName ()
-  convertName (IR.Ident ident) = HSE.UnQual () (HSE.Ident () ident)
-  convertName (IR.Symbol sym)  = HSE.UnQual () (HSE.Symbol () sym)
+  -- | Converts an unqualified IR name to a HST name.
```
stu204767:
```suggestion
  -- | Converts an unqualified IR name to an HST name.
```

src/lib/FreeC/Frontend/Haskell/PatternMatching.hs
```diff
@@ -1,117 +1,273 @@
...
+  convertConEntry :: EnvEntry -> HST.ConEntry Frontend
+  convertConEntry entry = HST.ConEntry
+    { HST.conEntryName    = convertQName (entryName entry)
+    , HST.conEntryArity   = entryArity entry
+    , HST.conEntryIsInfix = isInfixConQName (entryName entry)
+    , HST.conEntryType    = convertConEntryType entry
+    }
+
+  -- | Converts the entry of an exported data constructor entry to a list of
+  --   its constructor entries.
```
stu204767: This is not what this function does. But maybe it gets removed anyway when adapting to the newest PMC version.

src/lib/FreeC/Frontend/Haskell/PatternMatching.hs
```diff
@@ -1,117 +1,273 @@
...
+
+  -- | Gets the name of the data type from the return type of the constructor.
+  extractTypeConQName :: IR.Type -> IR.QName
+  extractTypeConQName (IR.TypeCon _ conName) = conName
+  extractTypeConQName (IR.TypeApp _ t1 _) = extractTypeConQName t1
+  extractTypeConQName _
+    = error "extractTypeConQName: Expected type constructor."
+
+-------------------------------------------------------------------------------
+-- Interpretation for `Report` Effect                                        --
```
stu204767: I would add an article here.
```suggestion
-- Interpretation for the `Report` Effect                                    --
```

src/lib/FreeC/IR/Syntax/Module.hs
```diff
@@ -47,3 +50,60 @@ data ImportDecl
...
+
+-------------------------------------------------------------------------------
+-- Function Declarations                                                     --
+-------------------------------------------------------------------------------
+modFuncDecls :: Module -> [FuncDecl]
+modFuncDecls = map topLevelFuncDecl . filter isTopLevelFuncDecl . modContents
+
+modWithFuncDecls :: [FuncDecl] -> Module -> Module
+modWithFuncDecls decls = modWithContents $ \contents ->
+  map TopLevelFuncDecl decls ++ filter (not . isTopLevelFuncDecl) contents
```
stu204767: These types and functions are all missing a Haddock comment, but I think this is not a big problem here.

src/test/FreeC/Frontend/IR/ParserTests.hs
```diff
@@ -570,63 +570,98 @@ testImportDeclParser = context "import declarations" $ do
 -- |Test group for 'Parseable' instance of 'IR.Module'.
```
stu204767:
```suggestion
-- | Test group for 'Parseable' instance of 'IR.Module'.
```

stu204767 (changes requested):

stu204146-kiel (approved):

src/lib/FreeC/Frontend/Haskell/PatternMatching.hs
just95: This is an old comment that just looks like an addition due to GitHub's diffing algorithm. I'll remove it : )

src/lib/FreeC/IR/Syntax/Module.hs
```diff
@@ -81,29 +89,41 @@ instance Pretty TopLevelDecl where
 -------------------------------------------------------------------------------
 -- Type Declarations                                                         --
 -------------------------------------------------------------------------------
+-- | Gets the type-level declarations of the given module.
 modTypeDecls :: Module -> [TypeDecl]
 modTypeDecls = map topLevelTypeDecl . filter isTopLevelTypeDecl . modContents
 
+-- | Sets the type-level declarations of the given module.
+--
+--   All other type-level declarations are discarded.
```
stu204767: `type-level` : )
```suggestion
-- | Gets the type declarations of the given module.
modTypeDecls :: Module -> [TypeDecl]
modTypeDecls = map topLevelTypeDecl . filter isTopLevelTypeDecl . modContents

-- | Sets the type declarations of the given module.
--
--   All other type declarations are discarded.
```

stu204767 (changes requested):

stu204767 (approved):

PR #224, opened 2022-05-09
Fix #223

### LINKED ISSUE #223: cabal.project references use a deprecated git format
### Description

Currently, trying to install the project produces an `The unauthenticated git protocol on port 9418 is no longer supported.` error.
The error seems to lie in the `cabal.project` file where some references use a [deprecated git format](https://github.blog/2021-09-01-improving-git-protocol-security-github/). 

### Steps to Reproduce

1. Clone the repo
2. Install dependencies
3. Run `cabal new-install freec`

#### Expected Behavior

Free-of-errors installation.

#### Actual Behavior

```
~/dev/free-compiler main
❯  cabal new-install freec
fatal: remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```

### Versions

 - Git commit: 6931b9ca652a185a92dd824373f092823aea4ea9
 - `ghc --version`: 8.6.5
 - `cabal --version`: 3.4.0.0
 - `coqc --version`: 8.12.2
 - OS: macOS Monterey 12.3.1

javiergarea:
### Issue

#223 

### Description of the Change

Updated the git references of the `cabal.project` file so they use a supported git protocol.

### Verification Process

Running `cabal new-install freec` does not produce the error mentioned in #223.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -2,7 +2,7 @@
 -- is downloaded directly from the FreeProving GitHub organization instead.
 source-repository-package
   type: git
-  location: git://github.com/FreeProving/language-coq.git
+  location: git@github.com:FreeProving/language-coq.git
   tag: v0.4.0.0
 
 -- The `haskell-src-transformations` package is still in an early development
@@ -10,7 +10,7 @@ source-repository-package
 -- directly from the FreeProving GitHub organization instead.
 source-repository-package
   type: git
-  location: git://github.com/FreeProving/haskell-src-transformations.git
+  location: git@github.com:FreeProving/haskell-src-transformations.git
   tag: v0.2.0.0
 
 -- Cabal configuration file for this package.


javiergarea:
As the check's log reports, SSH access is not configured in the pipeline/repo. Maybe an SSH key can be included as a secret in the repo so something like [ssh-agent](https://github.com/webfactory/ssh-agent) can be used to set it in the workflow.

just95:
Hi, thank you very much for your two reports and pull requests. I've applied your patches locally and can confirm that they fix the reported issues. Besides the SSH problem that you've mentioned in your comment above, the checks also seem to fail due to file permission issues in the Docker images for Coq.

The easiest solution to the SSH problem seems to be to use HTTPS instead. I've tried this in 4871d0f4d0430844148bacd2d316d30508b026ec and it works. This way we don't have to change the workflow or add another secret. The file permission problem should be fixed with 92e3dd13672e72c1c45002ca523ae30c32b687db.

I am going to merge your PRs and reapply my fixes to the workflow. Thanks again and sorry for your inconvenience. This project has not received much attention lately. I'll configure the workflow to run every once in a while to make sure that at least the checks keep passing.

